PR #3, opened 2017-05-15
Demo for how TensorFlow could be used for this [Don't Merge]

oxinabox:
In the attached [notebook:](https://github.com/JuliaDiffEq/NeuralNetDiffEq.jl/blob/tensorflow_demo/proto/TensorFlow%20DiffEq%20Solving.ipynb)
is some demo code showing how problem 2 can be solved using Tensorflow.

Note the complete lack of analytic solution to the loss function w.r.t the derivative and back-propagation. 
instead using `du_dt = gradients(u, tt)`

There is no reason to merge this.
I just made it a PR, so that it could be found and discussed.


oxinabox:
See: https://github.com/JuliaDiffEq/TensorFlowDiffEq.jl
for fuller demo

PR #5, opened 2019-06-07
Сompatible solution with version 0.7.10

KirillZubov:
It is a rewritten code that is compatible with the version of Julia 0.7.10. All grammatical structures that are no longer supported by the new version of the compiler have been rewritten  by those that are supported.

### DIFF
--- a/src/NeuralNetDiffEq.jl
+++ b/src/NeuralNetDiffEq.jl
@@ -1,72 +1,58 @@
 __precompile__()
 
 module NeuralNetDiffEq
-#dependencies
 
+#dependencies
 using Reexport
 @reexport using DiffEqBase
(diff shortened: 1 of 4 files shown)


src/NeuralNetDiffEq.jl
```diff
@@ -1,72 +1,58 @@
 __precompile__()
 
 module NeuralNetDiffEq
-#dependencies
 
+#dependencies
 using Reexport
 @reexport using DiffEqBase
 using Knet, Compat, ForwardDiff
```
oxinabox: Can we drop the Compat requirment yet?

ChrisRackauckas: yes

PR #6, opened 2019-06-10
compatible with version 1.1.1

KirillZubov:
It is a rewritten code that is compatible with the version of Julia 1.1.1. All grammatical structures that are no longer supported by the new version of the compiler have been rewritten by those that are supported.


ChrisRackauckas:
Great!

PR #9, opened 2019-06-24
high-dimensional  semilinear pde solver using deep bsde algorithm

KirillZubov:
prototype of one-dimensional pde solver using deep bsde algorithm


src/PdeSolve.jl
```diff
@@ -0,0 +1,82 @@
...
+    chainU = U0(hide_layer_size, d)
+    ps = Flux.params(chainU, chains...)
+
+    # brownian motion
+    dw(dt) = sqrt(dt) * randn()
+    # the Euler-Maruyama scheme
+    x_sde(x_dim,t,dwa) = [x_dim[i] + μ(t,x_dim[i])*dt + σ(t,x_dim[i])*dwa[i] for i = 1: d]
+
+    get_x_sde(x_cur,l,dwA) = [x_sde(x_cur[i], ts[l] , dwA[i]) for i = 1: m]
+    reduceN(x_dim, l, dwA) = sum([gradU*dwA[i] for (i, gradU) in enumerate(chains[l](x_dim))])
```
ChrisRackauckas: gradU is a function?

src/PdeSolve.jl
```diff
@@ -0,0 +1,82 @@
...
+
+    data = Iterators.repeated((), maxiters)
+    ts = t0:dt:tn
+
+    #hidden layer
+    hide_layer_size = neuralNetworkParams[1]
+    opt = neuralNetworkParams[2]
+
+    U0(hide_layer_size, d) = neuralNetworkParams[3](hide_layer_size, d)
+    gradU(hide_layer_size, d) = neuralNetworkParams[4](hide_layer_size, d)
```
ChrisRackauckas: I think this is a good spot for unicode on the grad.

src/PdeSolve.jl
```diff
@@ -0,0 +1,82 @@
...
+    get_x_sde(x_cur,l,dwA) = [x_sde(x_cur[i], ts[l] , dwA[i]) for i = 1: m]
+    reduceN(x_dim, l, dwA) = sum([gradU*dwA[i] for (i, gradU) in enumerate(chains[l](x_dim))])
+    getN(x_cur, l, dwA) = [reduceN(x_cur[i], l, dwA[i]) for i = 1: m]
+
+    x_0 = [x0 for i = 1: m]
+
+    function sol()
+        x_cur = x_0
+        U = [chainU(x)[1] for x in x_0]
+        global x_prev
```
ChrisRackauckas: why global this? that shouldn't be needed. `local x_prev` should be fine.

src/PdeSolve.jl
```diff
@@ -0,0 +1,82 @@
...
+    ts = t0:dt:tn
+
+    #hidden layer
+    hide_layer_size = neuralNetworkParams[1]
+    opt = neuralNetworkParams[2]
+
+    U0(hide_layer_size, d) = neuralNetworkParams[3](hide_layer_size, d)
+    gradU(hide_layer_size, d) = neuralNetworkParams[4](hide_layer_size, d)
+
+    chains = [gradU(hide_layer_size, d) for i=1:length(ts)]
```
ChrisRackauckas: I think it be much easier to read if it was just 

```julia
u0 = neuralNetworkParams[3](hide_layer_size, d)
∇u = [neuralNetworkParams[4](hide_layer_size, d) for i=1:length(ts)]
```

since then it would be notationally a lot closer.

src/PdeSolve.jl
```diff
@@ -0,0 +1,82 @@
...
+    chainU = U0(hide_layer_size, d)
+    ps = Flux.params(chainU, chains...)
+
+    # brownian motion
+    dw(dt) = sqrt(dt) * randn()
```
ChrisRackauckas: delete this so you can just use `dw` for the Brownian motions. It took a bit to see that `dwa` was just a renaming of `dw array` since `dw` was taken.

src/PdeSolve.jl
```diff
@@ -0,0 +1,82 @@
...
+    # the Euler-Maruyama scheme
+    x_sde(x_dim,t,dwa) = [x_dim[i] + μ(t,x_dim[i])*dt + σ(t,x_dim[i])*dwa[i] for i = 1: d]
+
+    get_x_sde(x_cur,l,dwA) = [x_sde(x_cur[i], ts[l] , dwA[i]) for i = 1: m]
```
ChrisRackauckas: this is going to need to be iterating through time. I am not sure how x_cur[i] is supposed to act as the previous value here?

src/PdeSolve.jl
```diff
@@ -0,0 +1,82 @@
...
+    reduceN(x_dim, l, dwA) = sum([gradU*dwA[i] for (i, gradU) in enumerate(chains[l](x_dim))])
+    getN(x_cur, l, dwA) = [reduceN(x_cur[i], l, dwA[i]) for i = 1: m]
```
ChrisRackauckas: I'm not sure what this is doing. What equation does this correspond to?

src/PdeSolve.jl
```diff
@@ -0,0 +1,82 @@
...
+    gradU(hide_layer_size, d) = neuralNetworkParams[4](hide_layer_size, d)
+
+    chains = [gradU(hide_layer_size, d) for i=1:length(ts)]
+    chainU = U0(hide_layer_size, d)
+    ps = Flux.params(chainU, chains...)
+
+    # brownian motion
+    dw(dt) = sqrt(dt) * randn()
+    # the Euler-Maruyama scheme
+    x_sde(x_dim,t,dwa) = [x_dim[i] + μ(t,x_dim[i])*dt + σ(t,x_dim[i])*dwa[i] for i = 1: d]
```
ChrisRackauckas: sigma is not necessarily diagonal, so you cannot essentially broadcast along the dimensions here.

PR #15, opened 2019-07-29
pde_solver using neural_sde

KirillZubov:


src/neural_sde.jl
```diff
@@ -0,0 +1,88 @@
...
+σᵀ∇u = Flux.Chain(Dense(d,hls,relu),
+                  Dense(hls,hls,relu),
+                  Dense(hls,d))
+g(X) = sum(X.^2)
+
+function trueODEfunc(dh,h,p,t)
+    u =  h[end]
+    X =  h[1:end-1]
+    _σᵀ∇u = σᵀ∇u(X) #TODO tracked σᵀ∇u
+    dh[end] = Flux.data(f(X, u, _σᵀ∇u, p, t))
```
ChrisRackauckas: it's definitely going to fail if you `.data` this

src/neural_sde.jl
```diff
@@ -0,0 +1,88 @@
...
+    dh[1:end-1] .= μ(X,p,t)
+    dh
+end
+
+function true_noise_func(dh,h,p,t)
+    X = h[1:end-1]
+    _σᵀ∇u = Flux.data(σᵀ∇u(X))
+    dh[end] = sum(_σᵀ∇u) # TODO _σᵀ∇u*dW nondioganal
```
ChrisRackauckas: yeah this isn't correct, but I think that works just for this case of sigma = 1?

src/neural_sde.jl
```diff
@@ -0,0 +1,88 @@
...
+    solve(prob, args...; kwargs...) |> Tracker.collect
+end
+
+n_sde = init_cond->neural_sde(init_cond,trueODEfunc,true_noise_func,tspan,SOSRI(),saveat=t,reltol=1e-1,abstol=1e-1)
+
+x0 = fill(1.0 , d)
+init_cond = fill(1.0 , d+1)
+
+function predict_n_sde()
+    _u0 = Flux.data(u0(x0)[1])
```
ChrisRackauckas: The initial data has to be tracked. Take a look at the DiffEqFlux examples.

ChrisRackauckas:
Yeah your issue is you're dropping all of the Tracker information, so there's no way for it to get a valid gradient.

src/neural_sde.jl
KirillZubov: yeah, I know. I fix this today

KirillZubov:
> what does that stand for?

src/pde_solve_ns.jl
```diff
@@ -0,0 +1,84 @@
...
+
+    data = Iterators.repeated((), maxiters)
+
+
+    #hidden layer
+    opt = alg.opt
+    u0 = alg.u0
+    σᵀ∇u = alg.σᵀ∇u
+    sde_algorithm = alg.sde_algorithm
+    ps = Flux.params(u0, σᵀ∇u...)
```
ChrisRackauckas:
```suggestion
    ps = Flux.params(u0, σᵀ∇u)
```

src/pde_solve_ns.jl
```diff
@@ -0,0 +1,84 @@
...
+    alg::NNPDENS;
+    timeseries_errors = true,
+    save_everystep=true,
+    adaptive=false,
+    abstol = 1f-6,
+    reltol = 1f-5,
+    verbose = false,
+    maxiters = 300,
+    dt,
+    trajectories)
```
ChrisRackauckas: splat kwargs to the sde solver

src/pde_solve_ns.jl
```diff
@@ -0,0 +1,84 @@
...
+    #hidden layer
+    opt = alg.opt
+    u0 = alg.u0
+    σᵀ∇u = alg.σᵀ∇u
+    sde_algorithm = alg.sde_algorithm
+    ps = Flux.params(u0, σᵀ∇u...)
+
+    function F(h, p, t)
+        u =  h[end]
+        X =  h[1:end-1].data
```
ChrisRackauckas: why are you taking the data part here?

Project.toml
```diff
@@ -8,6 +8,9 @@ DiffEqBase = "2b5f629d-d688-5b77-993f-72d75c75574e"
 Flux = "587475ba-b771-5e3f-ad9e-33799f191a9c"
 Reexport = "189a3867-3050-52da-a836-e630ba90ab69"
 Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
+StochasticDiffEq = "789caeaf-c7a9-5a7d-9973-96adeb23e2a0"
```
ChrisRackauckas: You don't need to directly depend on StochasticDiffEq

src/pde_solve_ns.jl
KirillZubov: it is ```ERROR: LoadError: MethodError: no method matching Float32(::Tracker.TrackedReal{Float32})``` with case where σ or μ dependent from Tracked X

ChrisRackauckas: Interesting that it still works with this... but since it works it seems to be fine.

Project.toml
KirillZubov: all the same, dependency on StochasticDiffEq  needed

KirillZubov: https://ci.appveyor.com/project/ChrisRackauckas/neuralnetdiffeq-jl/builds/26537921

PR #16, opened 2019-08-11
added option to save steps of iteration

KirillZubov:


src/pde_solve_ns.jl
```diff
@@ -69,13 +70,16 @@ function DiffEqBase.solve(
         mean(sum(abs2, g(X.data) - u) for (X,u) in predict_n_sde())
     end
 
+    iters = Float32[]
```
ChrisRackauckas: eltype(X0)

ChrisRackauckas:
add to the other method as well?

KirillZubov:
> add to the other method as well?

ok

src/pde_solve_ns.jl
KirillZubov: In pde_solve tests x0 is Array of Float64. I suppose, it should be eltype(u0(X0)[1].data) but it is always Array of Float32.

ChrisRackauckas: You should make the inputs in the test be Float32 then for type stability

KirillZubov: ok

PR #27, opened 2019-09-17
Update README.md

matthieugomez:
Tests for NNPDENS are not passing for me. I'm doing this pull request to check whether they still pass on Travis.


ChrisRackauckas:
Looks like it's passing. Peculiar.

PR #30, opened 2019-09-17
Rename ans -> sol

matthieugomez:
ans returns the value of the last value in Julia, so I am not sure the tests were testing for anything.

Tests for NNPDENS fail for me after this correction

### DIFF
--- a/test/NNPDEHan_tests.jl
+++ b/test/NNPDEHan_tests.jl
@@ -28,16 +28,16 @@ u0 = Flux.Chain(Dense(d,hls,relu),
 
 alg = NNPDEHan(u0, σᵀ∇u, opt = opt)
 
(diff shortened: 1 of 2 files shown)


matthieugomez:
How are you interpreting Travis vs AppVeyor?
Maybe each test could set a random seed?

ChrisRackauckas:
Yeah I think it could just be some kind of tolerance thing. We should probably run the HJB one longer.

ChrisRackauckas:
Should be fixed now, but it also doesn't change the results from what I could see: It doesn't seem to overwrite ans anymore if you explicitly write to it?

ChrisRackauckas:
If it is an issue, this should get rebased to a newer version, it's just a bit too far back now!

PR #32, opened 2019-09-28
move the call of X.data inside the method

KirillZubov:
issue https://github.com/JuliaDiffEq/NeuralNetDiffEq.jl/issues/29

### DIFF
--- a/src/pde_solve_ns.jl
+++ b/src/pde_solve_ns.jl
@@ -33,14 +33,14 @@ function DiffEqBase.solve(
 
     function F(h, p, t)
         u =  h[end]
-        X =  h[1:end-1]
+        X =  h[1:end-1].data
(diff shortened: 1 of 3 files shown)


ChrisRackauckas:
Why should these not be tracked?

KirillZubov:
> Why should these not be tracked?
X doesn't need tracked градиент but if pass tracked X in μ(X, p, t), we get an error. Really, just a call of x.data was moved inside the method. I didn't change the method behavior.

ChrisRackauckas:
Why would that give the right gradient?

KirillZubov:
> Why would that give the right gradient?

X is points. We don't need the gradient of points to calculate backpropagation.

ChrisRackauckas:
oh yes, `X` can be considered a random constant.

PR #33, opened 2019-09-29
update to the new Flux based on Zygote

KirillZubov:


ChrisRackauckas:
this doesn't do it quite yet because it needs to test Flux master right now to see. But it will fail because the `.data` stuff only should happen on Tracker variables.

ChrisRackauckas:
This is now just waiting on https://github.com/FluxML/Flux.jl/pull/986 and then we are on Zygote.

ChrisRackauckas:
Note we can be forward-mode AD compatible in the neural net ODE solver, but it's probably still not a great idea to use AD there.

PR #42, opened 2019-12-12
Second Order ODE Solver

rayhanrazzak:
Still needs to support partitioned dynamical ODEs. Open to comments.


ChrisRackauckas:
The binaries need to be removed and squashed, otherwise they will fill the git history.

ChrisRackauckas:
Looks mostly good, it just needs to handle the common `SecondOrderODEProblem` form, i.e. `DynamicalODEProblem`.

PR #47, opened 2019-12-25
initialized Neural RODE

alhajri:
This is just to get the ball rolling on adding the RODE functionality to this package. 

Currently, this algorithm has only been validated on one RODE (i.e. not a system). The algorithm trains on the discrete time steps, however changing it to train on randomly selected time points is a possibility. 

Also, this assumes that the randomness can be directly sampled from a normal distribution. I would like to change that so that it takes in the random noise process from the problem specification and uses that for training.

Finally, what should the solve function output? currently it only outputs one run of the RODE solution, maybe a better thing would be to output an ensemble problem.

### DIFF
--- a/src/NeuralNetDiffEq.jl
+++ b/src/NeuralNetDiffEq.jl
@@ -32,8 +32,8 @@ end
 include("ode_solve.jl")
 include("pde_solve.jl")
 include("pde_solve_ns.jl")
+include("rode_solve.jl")
 
-
-export NNODE, TerminalPDEProblem, NNPDEHan, NNPDENS
+export NNODE, TerminalPDEProblem, NNPDEHan, NNPDENS, NNRODE
 
 end # module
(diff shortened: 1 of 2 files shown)


ChrisRackauckas:
Good stuff! This is definitely a good starting point for someone to continue from.

>Also, this assumes that the randomness can be directly sampled from a normal distribution. I would like to change that so that it takes in the random noise process from the problem specification and uses that for training.

https://docs.juliadiffeq.org/latest/features/noise_process/

One way to do it may be to solve for a trajectory of the noise process and then sample from it via interpolation.

>Finally, what should the solve function output? currently it only outputs one run of the RODE solution, maybe a better thing would be to output an ensemble problem.

Possibly the right thing to do is to take in an Ensemble{RODEProblem}. Though an interesting thing about this is that after the neural network is trained, many solutions can quickly be generated using that trained neural network, so documentation of some interface over that is probably a good idea. Returning an ensemble solution from a single solve would be a weird interface violation though IMO

ChrisRackauckas:
Thanks! This ended up becoming https://github.com/SciML/NeuralNetDiffEq.jl/pull/75 which will close the issue. Thanks for starting this work!

PR #48, opened 2019-12-27
fix ode_solver

KirillZubov:
Fix some issues in ode_solver.  
The method didn't work with some type of equations(examples:f = (u,p,t) -> 1.01*u and fitzhugh nagumo model)


ChrisRackauckas:
Interesting that doesn't drop gradients. The only issue here is that we'll need to drop the `.data` pieces once we switch to Zygote, which should be fairly soon.

PR #54, opened 2020-01-23
Adding 2nd order ODE solver & test

chinglamchoi:


src/NeuralNetDifEq.jl
```diff
@@ -15,11 +15,8 @@ struct TerminalPDEProblem{G,F,Mu,Sigma,X,T,P} <: DiffEqBase.DEProblem
     X0::X
     tspan::Tuple{T,T}
     p::P
-    TerminalPDEProblem(g,f,μ,σ,X0,tspan,p=nothing) = new{typeof(g),typeof(f),
-                                                         typeof(μ),typeof(σ),
-                                                         typeof(X0),eltype(tspan),
-                                                         typeof(p)}(
-                                                         g,f,μ,σ,X0,tspan,p)
+    TerminalPDEProblem(g,f,μ,σ,X0,tspan,p=nothing) = new{typeof(g),typeof(f),typeof(μ),typeof(σ),typeof(X0),eltype(tspan),
```
ChrisRackauckas: why was this changed?

src/second_ode_solve.jl
```diff
@@ -0,0 +1,63 @@
+using Flux
```
ChrisRackauckas: shouldn't need to be in this file.

src/NeuralNetDifEq.jl
chinglamchoi: There seems to be some formatting issues (I was using 2 both vim and sublime text), I should probably create a new pull request.

test/NNODE2_tests.jl
```diff
@@ -0,0 +1,28 @@
...
+tspan = (0.0f0, 2.0f0)
+u0 = [1.0f0]
+du0 = [0.0f0]
+dt = [1/20f0]
+prob = SecondOrderODEProblem(f, u0, du0, tspan)
+chain = Chain(Dense(1,5,σ),Dense(5,1))
+opt = ADAM(0.1, (0.9, 0.95))
+sol = solve(prob, NeuralNetDiffEq.NNODE2(chain,opt), dt=dt, verbose = true, abstol=1e-10, maxiters = 200)
+
+function an_sol(tspan, dt)
```
ChrisRackauckas:
```julia
t = tspan[1]:dt:tspan[2]
an_sol = @. 1/5 * exp(-3*t) *(3*exp(5*t)+2.0)
```

much simpler.

src/second_ode_solve.jl
chinglamchoi: corrected

src/second_ode_solve.jl
```diff
@@ -0,0 +1,63 @@
...
+        push!(ts, cnt)
+        cnt += dt[1]
+    end
+    
+    #Initial Value Problem
+    phi(t) = (u0 .+ du0 .* (t .- tspan[1]) .+ (t .- tspan[1]) .^ 2 .* chain([t]))[1]
+
+
+    dfdx = t -> (phi(t+sqrt(eps(typeof(dt[1])))) - phi(t)) / sqrt(eps(typeof(dt[1])))
+    g(tp) = sum(dfdx(tp))
```
ChrisRackauckas: I don't think this is right because of the `sum`, nor a necessarily good way to do it. I would directly calculate the second derivative, i.e. 

```julia
epsilon = cbrt(eps(real(T)))
(phi(t+epsilon) - 2phi(t) + phi(t-epsilon))/epsilon^2
```

https://en.wikipedia.org/wiki/Finite_difference where `epsilon = cbrt(eps(real(T)))` is taken from https://github.com/JuliaDiff/FiniteDiff.jl/blob/master/src/epsilons.jl#L29 as an optimal choice.

test/NNODE2_tests.jl
chinglamchoi: Forgot to amend this after reading the ode solver code, will do!

src/second_ode_solve.jl
chinglamchoi: I have briefly read through the links (will study them more later) and will amend the code!

test/NNODE2_tests.jl
```diff
@@ -0,0 +1,21 @@
+using Test, Flux
```
ChrisRackauckas: this test isn't actually ran. It should be added to runtests.jl

src/second_ode_solve.jl
```diff
@@ -0,0 +1,63 @@
...
+        push!(ts, cnt)
+        cnt += dt[1]
+    end
+    
+    #Initial Value Problem
+    phi(t) = (u0 .+ du0 .* (t .- tspan[1]) .+ (t .- tspan[1]) .^ 2 .* chain([t]))[1]
+
+
+    dfdx = t -> (phi(t+sqrt(eps(typeof(dt[1])))) - phi(t)) / sqrt(eps(typeof(dt[1])))
+    epsilon(t) = cbrt(eps(real(t)))
```
ChrisRackauckas: this is going to cause problems at zero. don't make it a function of `t`.

test/NNODE2_tests.jl
```diff
@@ -0,0 +1,21 @@
...
+using Plots
+using DiffEqDevTools
+
+# Homogeneous
+f(du, u, p, t) = -du+6*u
+tspan = (0.0f0, 2.0f0)
+u0 = [1.0f0]
+du0 = [0.0f0]
+dt = [1/20f0]
```
ChrisRackauckas: 1/20f0

src/second_ode_solve.jl
chinglamchoi: Declared t0 and changed to real(t-t0)

chinglamchoi: t0 is a float though, so even if t = 0, wouldn't this still be a float?

src/second_ode_solve.jl
```diff
@@ -0,0 +1,64 @@
...
+    adaptive=false,
+    abstol = 1f-6,
+    verbose = false,
+    maxiters = 100)
+
+    DiffEqBase.isinplace(prob) && error("Only out-of-place methods are allowed!")
+
+    u0, du0 = prob.u0
+    tspan = prob.tspan
+    f = prob.f
```
ChrisRackauckas:
```suggestion
    f = prob.f.f1
```

ChrisRackauckas: it's from how it's handled internally, so 1st order ODE solvers can work with it: https://github.com/JuliaDiffEq/DiffEqBase.jl/blob/master/src/problems/ode_problems.jl#L143-L176

src/second_ode_solve.jl
ChrisRackauckas:
```julia
julia> eps(0.0)
5.0e-324
```

is a really bad idea for numerical error.

src/second_ode_solve.jl
chinglamchoi: I see, sure!

src/second_ode_solve.jl
chinglamchoi: Oic! Perhaps a function of dt?

ChrisRackauckas: This is why I suggested it be based on `eps(Float32)` which is based on 1 since what it needs to be based on is the relative scale of phi(t) which is unknown until it's computed. The theory for `cbrt` is on the arbitrary size of that

chinglamchoi: Ah! At first i didn't really understand eps(Float32)

src/second_ode_solve.jl
```diff
@@ -0,0 +1,64 @@
...
+    
+    #Initial Value Problem
+    phi(t) = (u0 .+ du0 .* (t .- tspan[1]) .+ (t .- tspan[1]) .^ 2 .* chain([t]))[1]
+
+
+    dfdx = t -> (phi(t+sqrt(eps(Float32))) - phi(t)) / sqrt(eps(Float32))
+    epsilon(t) = cbrt(eps(Float32))
+    #second order central
+    d2fdx2(t) = (phi(t+epsilon(t)) - 2phi(t) + phi(t-epsilon(t)))/epsilon(t)^2
+    loss = () -> sum(abs2,sum(abs2,d2fdx2(t) .- f(phi(t), phi'(t),p,t)[1]) for t in ts)
```
KirillZubov: probably here should be `f(phi'(t), phi(t), p,t)[1])` because the interface is`f(du, u, p, t) `

KirillZubov: moreover, I think the problem in that method `phi'(t)`

ChrisRackauckas: It shouldn't use Zygote for it. phi' should instead be defined as a the first difference directly.

chinglamchoi: Thanks! Forgot to change the pr version

src/second_ode_solve.jl
```diff
@@ -0,0 +1,60 @@
...
+    opt    = alg.opt
+    ps     = Flux.params(chain)
+    data   = Iterators.repeated((), maxiters)
+
+    #train points generation
+    cnt = 0.0f0
+    ts  = tspan[1]:dt:tspan[2]
+    
+    #Initial Value Problem
+    phi(t) = (u0 .+ du0 .* (t .- tspan[1]) .+ (t .- tspan[1]) .^ 2 .* chain([t])[1])[1]
```
KirillZubov: these should be just `chain([t])` instead 'chain([t])[1])'

test/NNODE2_tests.jl
```diff
@@ -0,0 +1,58 @@
...
+  Dense(512, 128, relu), 
+  Dense(128, 64, tanh), 
+  Dense(64, 1))
+=#
+opt = ADAM(1e-04, (0.9, 0.95))
+sol = solve(prob, NeuralNetDiffEq.NNODE2(chain,opt), dt=dt, verbose = true, abstol=1e-10, maxiters = 200)
+
+t = tspan[1]:dt:tspan[2]
+an_sol = @. 1/5 * exp(-3*t) *(3*exp(5*t)+2.0)
+plot((t, an_sol), title="Second Order Linear ODE", label="Analytical", lw=3)
```
ChrisRackauckas: now remove the plotting and make these into tests.

test/NNODE2_tests.jl
```diff
@@ -0,0 +1,53 @@
...
+du0 = [0.0f0]
+dt = 1/20f0
+function an_sol(du0, u0, p, t)
+  [1/5 * exp(-3*t) *(3*exp(5*t)+2.0)]
+end
+prob = SecondOrderODEProblem(ODEFunction{false}(f, analytic=an_sol), u0, du0, tspan)
+chain = Chain(Dense(1,5,σ),Dense(5,1))
+opt = ADAM(1e-04, (0.9, 0.95))
+sol = solve(prob, NeuralNetDiffEq.NNODE2(chain,opt), dt=dt, verbose = true, abstol=1e-10, maxiters = 200)
+println(sol)
```
ChrisRackauckas: should add a test here.

test/NNODE2_tests.jl
```diff
@@ -0,0 +1,53 @@
...
+du0 = [0.0f0]
+dt = 1/20f0
+function an_sol(du0, u0, p, t)
+  [1/50 * exp(-t) * (2*exp(t) * (25*t^2 - 20*t + 58) - 13*sin(2*t) - 66*cos(2*t))]
+end
+prob = SecondOrderODEProblem(ODEFunction{false}(f, analytic=an_sol), u0, du0, tspan)
+chain = Chain(Dense(1,5,σ),Dense(5,1))
+opt = ADAM(1e-04, (0.9, 0.95))
+sol = solve(prob, NeuralNetDiffEq.NNODE2(chain,opt), dt=dt, verbose = true, abstol=1e-10, maxiters = 200)
+println(sol)
```
ChrisRackauckas: should add a test here

test/NNODE2_tests.jl
ChrisRackauckas: for reference, see the NNODE tests. It checks a solution value against an analytical solution and tests that it's close enough.

src/second_ode_solve.jl
```diff
@@ -0,0 +1,60 @@
...
+    t0 = tspan[1]
+
+    #hidden layer
+    chain  = alg.chain
+    opt    = alg.opt
+    ps     = Flux.params(chain)
+    data   = Iterators.repeated((), maxiters)
+
+    #train points generation
+    cnt = 0.0f0
```
KirillZubov: delete this line, it doesn't need any more

src/second_ode_solve.jl
```diff
@@ -0,0 +1,60 @@
...
+    ts  = tspan[1]:dt:tspan[2]
+    
+    #Initial Value Problem
+    phi(t) = (u0 .+ du0 .* (t .- tspan[1]) .+ (t .- tspan[1]) .^ 2 .* chain([t]))[1]
+
+
+    dfdx = t -> (phi(t+sqrt(eps(Float32))) - phi(t)) / sqrt(eps(Float32))
```
KirillZubov: Float32 -> typeof(dt)

src/second_ode_solve.jl
```diff
@@ -0,0 +1,58 @@
+struct NNODE2{C,O} <: NeuralNetDiffEqAlgorithm
+    chain::C
+    opt::O
+end
+NNODE2(chain;opt=Flux.ADAM(0.1)) = NNODE2(chain,opt)
+
+function DiffEqBase.solve(
+    prob::DiffEqBase.AbstractODEProblem,
+    alg::NeuralNetDiffEqAlgorithm,
```
ChrisRackauckas: this should dispatch on NNODE2

test/NNODE2_tests.jl
```diff
@@ -0,0 +1,74 @@
...
+dt = 1/5f0
+function an_sol(du0, u0, p, t)
+  [1/5 * exp(-3*t) *(3*exp(5*t)+2.0)]
+end
+prob = SecondOrderODEProblem(ODEFunction{false}(f, analytic=an_sol), u0, du0, tspan)
+chain = Chain(Dense(1,5,σ),Dense(5,1))
+opt = ADAM(1e-04, (0.9, 0.95))
+sol = solve(prob, NeuralNetDiffEq.NNODE2(chain,opt), dt=dt, verbose = true, maxiters = 200)
+#println(sol)
+err = sol.errors[:12]
```
KirillZubov: `12` -> `l2` but now `sol.errors` is empty

src/second_ode_solve.jl
```diff
@@ -0,0 +1,58 @@
...
+    
+    #Initial Value Problem
+    phi(t) = (u0 .+ du0 .* (t .- tspan[1]) .+ (t .- tspan[1]) .^ 2 .* chain([t]))[1]
+
+
+    dfdx = t -> (phi(t+sqrt(eps(typeof(dt)))) - phi(t)) / sqrt(eps(typeof(dt)))
+    epsilon(t) = cbrt(eps(typeof(dt)))
+    #second order central
+    d2fdx2(t) = (phi(t+epsilon(t)) - 2phi(t) + phi(t-epsilon(t)))/epsilon(t)^2
+    loss = () -> sum(abs2,sum(abs2,d2fdx2(t) .- f(dfdx(t), phi(t),p,t)[1]) for t in ts)
```
KirillZubov: add support for systems of equations here. You can see how this is done in ode_solver `if u0 isa Number .. else` and etc

src/second_ode_solve.jl
```diff
@@ -0,0 +1,58 @@
...
+
+    cb = function ()
+        l = loss()
+        verbose && println("Current loss is: $l")
+        l < abstol && Flux.stop()
+    end
+    Flux.train!(loss, ps, data, opt; cb = cb)
+
+    #solutions at timepoints
+    u = [phi(t) for t in ts]
```
KirillZubov: and here too

src/second_ode_solve.jl
KirillZubov: probably, we can avoid using `if...else`. Anyway, it is needed to be fixed. We get `boundary Error` for the example from `NNODE2_tests2.jl`

test/NNODE2_tests.jl
chinglamchoi: Fixed typo, have not been able to get it working for 2nd order yet

chinglamchoi: Fixed typo, haven't been able to get it to work with 2nd order yet

test/NNODE2_tests2.jl
```diff
@@ -0,0 +1,40 @@
...
+function f(du, u, p, t)
+	[(-p[3]*u[1]-p[4]*(u[1]-u[2]))/p[1], (-p[4]*(u[2]-u[1]))/p[2]]
+end
+p = [1.0f0, 1.0f0, 6.0f0, 4.0f0]
+u0 = [1.0f0, 0.0f0]
+du0 = [2.0f0, 0.0f0]
+prob = SecondOrderODEProblem{false}(f, [u0], [du0], (0.0f0, 3.0f0), p)
+dt = 1/20f0
+opt = ADAM(1e-03)
+chain = Chain(Dense(1,5,σ),Dense(5,1))
```
KirillZubov: `chain = Chain(Dense(1,5,σ),Dense(5,2))`

test/NNODE2_tests2.jl
```diff
@@ -0,0 +1,40 @@
...
+	ddu[1] = (-p[3]*u[1]-p[4]*(u[1]-u[2]))/p[1]
+	ddu[2] = (-p[4]*(u[2]-u[1]))/p[2]
+end
+function f(du, u, p, t)
+	[(-p[3]*u[1]-p[4]*(u[1]-u[2]))/p[1], (-p[4]*(u[2]-u[1]))/p[2]]
+end
+p = [1.0f0, 1.0f0, 6.0f0, 4.0f0]
+u0 = [1.0f0, 0.0f0]
+du0 = [2.0f0, 0.0f0]
+prob = SecondOrderODEProblem{false}(f, [u0], [du0], (0.0f0, 3.0f0), p)
```
KirillZubov: why do you use `[u0], [du0]`? Probably, it will be easy just to use `u0, du0` and change accordingly `phi`

chinglamchoi: It's actually because if I don't, it just gets passed as a (4,) and i can't easily separate u0 and du0 when calling `u0, du0 = prob.u0` in solver

test/NNODE2_tests2.jl
```diff
@@ -0,0 +1,40 @@
...
+    Conv((1,), 1028=>1028), 
+    x -> reshape(x, :, size(x, 4)), 
+    Dense(1028, 512, tanh), 
+    Dense(512, 128, relu), 
+    Dense(128, 64, tanh), 
+    Dense(64, 2), 
+    softmax)
+=#
+sol = solve(prob, NeuralNetDiffEq.NNODE2(chain, opt), dt=dt, verbose=true, abstol=1e-10, maxiters=50)
+#plot(sol)
```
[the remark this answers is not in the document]
KirillZubov: @chinglamchoi

test/NNODE2_tests.jl
```diff
@@ -0,0 +1,71 @@
...
+f_analytic = t -> [1/5 * exp(-3*t) *(3*exp(5*t)+2.0)]
+prob = SecondOrderODEProblem(ODEFunction{false}(f, analytic=f_analytic), u0, du0, tspan)
+chain = Chain(Dense(1,5,σ),Dense(5,1))
+opt = ADAM(1, (0.9, 0.95))
+sol = solve(prob, NeuralNetDiffEq.NNODE2(chain,opt), dt=dt, verbose = true, maxiters = 100)
+err = sqrt(sum((sol.u - [f_analytic(t)[1] for t in tspan[1]:dt:tspan[2]]).^2)[1])
+println("L2 Norm: ", err, "\n")
+sol = solve(prob, NeuralNetDiffEq.NNODE2(chain,opt), dt=1/20f0, verbose = true, maxiters = 100)
+err = sqrt(sum((sol.u - [f_analytic(t)[1] for t in tspan[1]:1/20f0:tspan[2]]).^2)[1])
+println("L2 Norm: ", err, "\n")
```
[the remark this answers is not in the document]
KirillZubov: @chinglamchoi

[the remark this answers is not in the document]
chinglamchoi: yes I can see the comments now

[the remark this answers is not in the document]
chinglamchoi: Will update!

ChrisRackauckas:
Looks like tests fail and the example doesn't converge.

ChrisRackauckas:
@KirillZubov did you ever finish this one up?

KirillZubov:
> did you ever finish this one up?
It is implemented in PINNs

ChrisRackauckas:
Good point

PR #56, opened 2020-02-02
Wrote tests for NNRODE solver.

ashutosh-b-b:
1. Slight change in the solver.
2. Wrote two examples to test the RODE solver.


ChrisRackauckas:
The .DS_Store is still in the git history, so this needs to be squashed before merging.

src/rode_solve.jl
```diff
@@ -0,0 +1,51 @@
...
+    ps     = Flux.params(chain)
+    data   = Iterators.repeated((), maxiters)
+
+    #train points generation
+    ts = tspan[1]:dt:tspan[2]
+
+    #The phi trial solution
+    ϕ(t,W) = u0 .- (t.-tspan[1]).*chain(Tracker.collect([t,W]))
+    
+    dfdx = (t,W) -> Tracker.gradient((t,W) -> sum(ϕ(t,W)), t,W; nest = true)[1]
```
ChrisRackauckas: This needs to be updated to the non-Tracker (Zygote) version

src/NeuralNetDiffEq.jl
```diff
@@ -2,9 +2,11 @@ module NeuralNetDiffEq
 
 using Reexport, Statistics
 @reexport using DiffEqBase
+
 using Flux, Zygote, DiffEqSensitivity
 import Tracker
```
ChrisRackauckas: this probably can be removed?

test/runtests.jl
```diff
@@ -8,8 +8,10 @@ const is_TRAVIS = haskey(ENV,"TRAVIS")
 if GROUP == "All" || GROUP == "Test1"
     @time @safetestset "NNODE" begin include("NNODE_tests.jl") end
     @time @safetestset "NNPDEHan" begin include("NNPDEHan_tests.jl") end
+    @time @safetestset "NNRODE" begin include("NNRODE_tests.jl") end
```
ChrisRackauckas: only add it to this group

test/NNRODE_tests.jl
```diff
@@ -0,0 +1,31 @@
...
+  2u*sin(W)
+end
+u0 = 0.00
+tspan = (0.0,5.0)
+prob = RODEProblem(f,u0,tspan)
+
+chain = Flux.Chain(Dense(2,5,σ),Dense(5,16 ,σ ) , Dense(16,2))
+opt = Flux.ADAM(0.1, (0.9, 0.95))
+sol = solve(prob, NeuralNetDiffEq.NNRODE(chain,opt), dt=1/20f0, verbose = true,
+            abstol=1e-10, maxiters = 1000)
```
ChrisRackauckas: is there any good test we can add on the value, say doing it against Euler-Maruyama on the same noise process?

ashutosh-b-b: Yes I can. But how do we compare them?

ChrisRackauckas: Take a look at https://docs.juliadiffeq.org/latest/features/noise_process/ for how to directly work with a noise process. We'll want to reuse the same noise process between the two solves.

ashutosh-b-b: Agreed. We can use alternate noise process. But how do we compare the results obtained from NNRODE and RandomEM?

ChrisRackauckas: Instead of directly generating and using random numbers, generate a `NoiseGrid` https://docs.juliadiffeq.org/latest/features/noise_process/#NoiseGrid-2 and perform the fit against that, i.e. write the loss function so that it evaluates the loss using a full NoiseProcess. Then after the network is trained, we will have one solution, but also the trained neural network is stored in the alg. We can then solve RODEs with `RandomEM` and then use sol.W to grab the noise process and throw that in to the NN. If trained well, the difference between the two solutions will be small.

ashutosh-b-b: Yes. This has been done in the latest commit. Can you please review this?

ChrisRackauckas:
The PR is all messed up: looks like the merge failed and the diff shows every change over the last 5 months.

test/NNRODE_tests.jl
```diff
@@ -0,0 +1,40 @@
...
+prob = RODEProblem(f,u0,tspan)
+dt = 0.1/20f0
+t = tspan[0]:dt:tspan[1]
+brownian_values = cumsum([0;[sqrt(dt1)*randn() for i in 1:length(t1)-2]])
+W = NoiseGrid(t1,brownian_values)
+chain = Flux.Chain(Dense(2,5,σ),Dense(5,16 ,σ ) , Dense(16,2))
+opt = Flux.ADAM(0.1, (0.9, 0.95))
+sol = solve(prob, NeuralNetDiffEq.NNRODE(chain,W,opt), dt=1/20f0, verbose = true,
+            abstol=1e-10, maxiters = 100)
+sol_real = sol_real = solve(prob,RandomEM(),dt=1/20f0)
```
ChrisRackauckas: now turn this into a test: how close it is? It should be close?

ChrisRackauckas:
Closing this off for https://github.com/SciML/NeuralNetDiffEq.jl/pull/75

PR #58, opened 2020-02-04
rename μ -> μ_f σ -> σ_f

KirillZubov:


azev77:
@KirillZubov @ChrisRackauckas 
In typical HJB problems, μ(u,x,t) is the drift of the state vector X and f() is the flow objective function. 

Maybe it would be more appropriate (closer to the math) to name these functions: 
μ_X( ) and  σ_X( ) ?

ChrisRackauckas:
Sure. These are just the names in the examples though, so it's not the biggest deal.

Issue #22, opened 2019-08-20
Kolmogorov equations through neural nets


ChrisRackauckas:
https://arxiv.org/pdf/1806.00421.pdf

ChrisRackauckas:
https://github.com/juliusberner/Deep-Multilevel-Kolmogorov-PDE-Solver

ashutosh-b-b:
I have started working on this.

ChrisRackauckas:
https://github.com/JuliaDiffEq/NeuralNetDiffEq.jl/pull/61 is it. Remaining details:

- Instead of a `KolmogorovPDEProblem`, it should just use `SDEProblem` with distribution input.
- Needs documentation
- Needs the ability to do high-dimensional SDEs
- Needs a high dimensional SDE test
- Use the ensemble interface https://docs.juliadiffeq.org/latest/features/ensemble/ for the ensemble of solutions to parallelize it.

I think if you take this all the way, a nice paper can be written on it.

ChrisRackauckas:
Completed by @ashutosh-b-b

PR #61, opened 2020-02-23
Intitialised Kolmogorov Equations

ashutosh-b-b:
Made a new problem for testing KolmogorovPDEs. I tested it with the Heat Equation given in the paper and it works.


ChrisRackauckas:
it looks like this branch is continuing your other branch?

ashutosh-b-b:
I have cleaned up the branch, sorry about the mess.

ashutosh-b-b:
This PR is still WIP.

src/kolmogorov_solve.jl
```diff
@@ -0,0 +1,47 @@
...
+    d = prob.d
+    T = tspan[2]
+    #hidden layer
+    chain  = alg.chain
+    opt    = alg.opt
+
+    ps     = Flux.params(chain)
+    xi     = rand(Uniform(xspan[1] , xspan[2]), d , maxiters )
+    N = Normal(0 , sqrt(2. *T ))
+    x_sde = xi + μ(xi)*T + sigma(xi)*rand(N , d , maxiters)
```
ChrisRackauckas: this should instead be defining SDEProblem and using a higher order SDE solver: this is a very inaccurate method.

ChrisRackauckas:
Any test case?

src/kolmogorov_solve.jl
ashutosh-b-b: I have changed this to the sde solver.

ashutosh-b-b:
I tested it against the Toy problem given in paper. It looks allright to me.

Project.toml
```diff
@@ -6,6 +6,8 @@ version = "1.1.0"
 [deps]
 DiffEqBase = "2b5f629d-d688-5b77-993f-72d75c75574e"
 DiffEqSensitivity = "41bf760c-e81c-5289-8e54-58b1f1f8abe2"
+DifferentialEquations = "0c46a032-eb83-5123-abaf-570d42b7fbaa"
```
ChrisRackauckas: just use StochasticDiffEq. It's already imported.

Project.toml
```diff
@@ -6,6 +6,8 @@ version = "1.1.0"
...
+Distributions = "31c24e10-a181-5473-b8eb-7969acd0382f"
```
ChrisRackauckas: not necessary

ChrisRackauckas:
There isn't a test.

src/kolmogorov_solve.jl
```diff
@@ -0,0 +1,51 @@
...
+    #hidden layer
+    chain  = alg.chain
+    opt    = alg.opt
+
+    ps     = Flux.params(chain)
+    xi     = rand(Uniform(xspan[1] , xspan[2]), d , N[1] )
+    #Finding Solution to the SDE having initial condition xi. Y = Phi(S(X , T))
+    sdeproblem = SDEProblem(μ,sigma,xi,tspan)
+    sol = solve(sdeproblem,EM(),dt=dt , saveat = T)
+    x_sde = sol[2]
```
ChrisRackauckas: why the second?

ashutosh-b-b: The solution is saved at 0 and T, thereby taking the solution evaluated at T

Project.toml
ashutosh-b-b: I removed the DifferentialEquations from dependencies.

test/NNKolmogorov_tests.jl
```diff
@@ -0,0 +1,42 @@
...
+    y = reshape(y , size(xi)[1] , size(xi)[2] )
+    return y
+end               
+real_sol = analytical(sol[1])
+err = 0
+for (r , s) in zip(real_sol , sol[2] )
+    err  = ((r - s)^2) + err
+end
+errorl2 = err/size(sol[2])[2]
+println("error_l2 = ", errorl2, "\n")
```
ChrisRackauckas: turn this into a test?

ashutosh-b-b: Added a test under the name kolmogorov_tests.jl

ChrisRackauckas: but it doesn't test anything

ashutosh-b-b: Added a test to test the error.

src/kolmogorov_solve.jl
ChrisRackauckas: Then do save_everystep=false and sol[end]. Just a more robust style

ChrisRackauckas: Seems like it would be nice to extend this to be time dependent though. A follow up making the algorithm time dependent and higher dimensional would be good.

src/kolmogorov_solve.jl
```diff
@@ -0,0 +1,51 @@
...
+    T = tspan[2]
+    #hidden layer
+    chain  = alg.chain
+    opt    = alg.opt
+
+    ps     = Flux.params(chain)
+    xi     = rand(Uniform(xspan[1] , xspan[2]), d , N[1] )
+    #Finding Solution to the SDE having initial condition xi. Y = Phi(S(X , T))
+    sdeproblem = SDEProblem(μ,sigma,xi,tspan)
+    sol = solve(sdeproblem,EM(),dt=dt , saveat = T)
```
ChrisRackauckas: Allow this to use other solvers. Add an sdealg to the NN alg.

src/kolmogorov_solve.jl
```diff
@@ -0,0 +1,51 @@
...
+ 
+function DiffEqBase.solve(
+    prob::KolmogorovPDEProblem,
+    alg::NeuralNetDiffEqAlgorithm;
+    abstol = 1f-6,
+    verbose = false,
+    maxiters = 300,
+    save_everystep = false,
+    dt,
+    )
```
ChrisRackauckas: Splat kwargs to the solver

test/NNKolmogorov_tests.jl
ChrisRackauckas: Awesome

ChrisRackauckas:
This is looking quite good. Added a few comments. Work out the conflicts and we can merge. In a follow-up we can work through making this higher dimensional and time dependent

ashutosh-b-b:
Added kwargs and sdealg in to the NNalg. Resolved conflicts.

src/kolmogorov_solve.jl
```diff
@@ -0,0 +1,53 @@
...
+    d = prob.d
+    ts = tspan[1]:dt:tspan[2]
+    N = size(ts)
+    T = tspan[2]
+    #hidden layer
+    chain  = alg.chain
+    opt    = alg.opt
+    sdealg = alg.sdealg
+    ps     = Flux.params(chain)
+    xi     = rand(Uniform(xspan[1] , xspan[2]), d , N[1] )
```
ChrisRackauckas: Do this directly instead of using `Uniform` and drop the dependency: it'll just make things more complex.

Project.toml
```diff
@@ -6,6 +6,7 @@ version = "1.2.0"
 [deps]
 DiffEqBase = "2b5f629d-d688-5b77-993f-72d75c75574e"
 DiffEqSensitivity = "41bf760c-e81c-5289-8e54-58b1f1f8abe2"
+Distributions = "31c24e10-a181-5473-b8eb-7969acd0382f"
 Flux = "587475ba-b771-5e3f-ad9e-33799f191a9c"
```
ChrisRackauckas: The test error is that you need to add Random here.

src/NeuralNetDiffEq.jl
```diff
@@ -2,7 +2,7 @@ module NeuralNetDiffEq
 
 using Reexport, Statistics
 @reexport using DiffEqBase
-using Flux, Zygote, DiffEqSensitivity
+using Flux, Zygote, DiffEqSensitivity, Distributions, Random,StochasticDiffEq
```
ChrisRackauckas: StochasticDiffEq isn't needed.

src/kolmogorov_solve.jl
ashutosh-b-b: How do I do this directly? Take an input in the problem?

src/kolmogorov_solve.jl
```diff
@@ -0,0 +1,53 @@
...
+    #Finding Solution to the SDE having initial condition xi. Y = Phi(S(X , T))
+    sdeproblem = SDEProblem(μ,sigma,xi,tspan)
+    sol = solve(sdeproblem, sdealg ,dt=dt , save_everystep=false , kwargs...)
+    x_sde = sol[end]
+    y = phi(x_sde)
+    data   = Iterators.repeated((xi , y), maxiters)
```
ChrisRackauckas: you're only testing on one data point: there's no way that is going to be the algorithm. It's likely looping over the solve?

ashutosh-b-b: xi is of shape [dimensions , number of data points] , and it is passed as an initial condition for sde. so same dimensions for x_sde. y is of the shape [dim_y , number of data points] .
Instead of looping over again for sde , I added all data points in initial condition.

ashutosh-b-b: Screenshot 2020-02-27 at 12 36 55 PM
 Although this can be done too.

ChrisRackauckas: but you're matching the network to only one run, instead of always resampling. I guess if the number of samples is high enough that's okay.

We'll want to move this to https://docs.juliadiffeq.org/latest/features/ensemble/ to get to higher dimensions and parallelism.

ChrisRackauckas: but you're matching the network to only one run, instead of always resampling. I guess if the number of samples is high enough that's okay.

We'll want to move this to https://docs.juliadiffeq.org/latest/features/ensemble/ to get to higher dimensions and parallelism.

ChrisRackauckas:
Can you show convergence like Figure 1?

src/kolmogorov_solve.jl
```diff
@@ -0,0 +1,54 @@
...
+    N = size(ts)
+    T = tspan[2]
+    #hidden layer
+    chain  = alg.chain
+    opt    = alg.opt
+    sdealg = alg.sdealg
+    ps     = Flux.params(chain)
+    xi     = rand(xs , d , N[1])
+    #Finding Solution to the SDE having initial condition xi. Y = Phi(S(X , T))
+    sdeproblem = SDEProblem(μ,sigma,xi,tspan)
```
ChrisRackauckas: the initial conditions are supposed to have phi as its probability distribution

ChrisRackauckas: the initial conditions are supposed to have phi as its probability distribution

src/kolmogorov_solve.jl
```diff
@@ -0,0 +1,54 @@
...
+    sol = solve(sdeproblem, sdealg ,dt=dt , save_everystep=false , kwargs...)
+    x_sde = sol[end]
+    y = phi(x_sde)
+    data   = Iterators.repeated((xi , y), maxiters)
+
+    #MSE Loss Function
+    loss(x , y) =Flux.mse(chain(x), y)
```
ChrisRackauckas: you want to randomize the input's x into chain.

src/kolmogorov_solve.jl
```diff
@@ -0,0 +1,54 @@
...
+    )
+
+    tspan = prob.tspan
+    phi = prob.phi
+    xspan = prob.xspan
+    sigma = prob.sigma
+    μ = prob.μ
+    d = prob.d
+    ts = tspan[1]:dt:tspan[2]
+    xs = xspan[1]:0.001:xspan[2]
```
ChrisRackauckas: this should just be the support of the input distribution?

ChrisRackauckas: this should just be the support of the input distribution?

src/kolmogorov_solve.jl
ashutosh-b-b: Sorry to ping you again. But do you mean to take some random values from existing data and put it in chain? or something else?

ashutosh-b-b: @ChrisRackauckas

ChrisRackauckas: I mean you should be taking new random values as you are taking the loss. Using one set of random numbers will bias the training.

ChrisRackauckas:
The test will need to be a lot stricter, like 5 digits of accuracy or so. If you see Figure 1, they get it down to like zero, so it should just happen. I don't think the implementation is correct because of how the randomness is fixed, so I don't think it should merge until it can show a lot more accuracy.

ChrisRackauckas:
Let me know if you still had questions on the comments.

ashutosh-b-b:
Screenshot 2020-03-11 at 9 36 47 PM
This is how convergence plot looks like

ashutosh-b-b:
Screenshot 2020-03-11 at 4 31 00 AM
This was plot for analytical vs algorithm solution

ChrisRackauckas:
Let's follow up with https://github.com/JuliaDiffEq/NeuralNetDiffEq.jl/issues/22#issuecomment-592056069

ashutosh-b-b:
> Let's follow up with [#22 (comment)](https://github.com/JuliaDiffEq/NeuralNetDiffEq.jl/issues/22#issuecomment-592056069)

Yes. I already posted a PR for documentation. I ll look into other points as well. Is this a follow on after the merge ?

ChrisRackauckas:
The docs should probably come after a change to the input. PM me and we can discuss

PR #69, opened 2020-03-05
General PDE solving with NNs

KirillZubov:
first prototype with ODEs


ChrisRackauckas:
I looks like this just drops the structural imposition and just does it all in the loss function like a PINN?

KirillZubov:
> I looks like this just drops the structural imposition and just does it all in the loss function like a PINN?
@ChrisRackauckas  Yes, it was easy to do for ODE. I'm going to implement the same for the general PDE solver like the next step.

src/NeuralNetDiffEq.jl
```diff
@@ -58,13 +58,95 @@ function Base.show(io::IO, A::KolmogorovPDEProblem)
...
+end
+Base.summary(prob::GeneranNN3DimPDEProblem) = string(nameof(typeof(prob)))
+
+function Base.show(io::IO, A::GeneranNN3DimPDEProblem)
+  println(io,summary(A))
+  print(io,"timespan: ")
+  show(io,A.tspan)
+end
+
+struct GeneranNNPDEProblem{PF,BF,S,D,P} <:DiffEqBase.DEProblem
```
ChrisRackauckas, line 112: sp

src/NeuralNetDiffEq.jl
```diff
@@ -58,13 +58,95 @@ function Base.show(io::IO, A::KolmogorovPDEProblem)
   show(io , A.sigma)
 end
 
+struct GeneranNN2DimPDEProblem{PF,BC,IC,T,X,DT,DX,P} <:DiffEqBase.DEProblem
```
ChrisRackauckas, line 61: why does 2D and 3D need to be treated differently?

src/general_ode_solve.jl
```diff
@@ -0,0 +1,112 @@
+struct NNGenODE{C,IC,O,K} <: NeuralNetDiffEqAlgorithm
```
ChrisRackauckas, line 1: why does this need to be different from the other one?

src/NeuralNetDiffEq.jl
KirillZubov: it is old code. I going to delete it soon.

src/general_ode_solve.jl
KirillZubov: yeah, it is the hard code. It also will be deleted.

src/NeuralNetDiffEq.jl
```diff
@@ -58,10 +58,10 @@ function Base.show(io::IO, A::KolmogorovPDEProblem)
   show(io , A.sigma)
 end
 
-struct GeneranNNPDEProblem{PF,BF,S,D,P} <:DiffEqBase.DEProblem
+struct GeneranNNPDEProblem{PF,BF,SP,D,P} <:DiffEqBase.DEProblem
   pde_func::PF
   bound_funcs::BF
-  space ::S
+  space ::SP
```
ChrisRackauckas, line 64: that's not what I meant, I meant there's a spelling error in all of these.

KirillZubov: oh yeah, thanks

src/general_nn_pde_solve.jl
```diff
@@ -0,0 +1,170 @@
+struct NNGeneralPDE{C,O,K} <: NeuralNetDiffEqAlgorithm
```
Vaibhavdixit02, line 1: I think the `General` can be dropped from these to maintain consistency https://github.com/SciML/NeuralNetDiffEq.jl/blob/d6831a0cb5a566b244e4e8de8b378ed7aef68830/src/ode_solve.jl#L1 and it always makes sense to add on terms for additional methods and use the generic version for the default?

Vaibhavdixit02: Same for the problem struct?

KirillZubov: thanks, yeah, it is better here use the generic version "NNDE{C,O,P,K}" for both solvers.

ChrisRackauckas:
The parsing from https://github.com/SciML/DiffEqOperators.jl/pull/250 might be a bit more robust: you might want to do something similar to @emmanuellujan to grab what to differentiate.

KirillZubov:
thanks, it is actually what I need

test/NNPDE_tests.jl
```diff
@@ -0,0 +1,123 @@
...
+eq = Dt(u(t,θ)) ~ t^3 + 2*t + (t^2)*((1+3*(t^2))/(1+t+(t^3))) - u(t,θ)*(t + ((1+3*(t^2))/(1+t+t^3)))
+bcs = [u(0.) ~ 1.0 , u(1.) ~ 1.202]
+
+# Space and time domains
+domains = [t ∈ IntervalDomain(0.0,1.0)]
+
+# Method of lines discretization
+dx = 0.1
+order = 1
+discretization = MOLFiniteDifference(dx,order)
```
ChrisRackauckas, line 23:
```suggestion
discretization = PhysicsInformedNN(dx)
```

test/NNPDE_tests.jl
```diff
@@ -0,0 +1,123 @@
...
+
+# neural network and optimizer
+opt = Flux.ADAM(0.1)
+chain = FastChain(FastDense(1,12,Flux.σ),FastDense(12,1))
+
+pde_system = PDESystem(eq,bcs,domains,[t],[u])
+prob = NeuralNetDiffEq.NNPDEProblem(pde_system,discretization)
```
ChrisRackauckas, line 30:
```suggestion
prob = discretize(pde_system,discretization)
```

ChrisRackauckas: It would actually be nice for this to kick out an OptimizationProblem, but that could come later.

src/pinns_pde_solve.jl
```diff
@@ -0,0 +1,224 @@
...
+    indvars = Expr.(_indvars)
+    vars = :($(depvars...), $(indvars...), θ, derivative,second_order_derivative)
+
+    _left_expr = (ModelingToolkit.simplified_expr(eq.lhs))
+    left_expr = simplified_derivative(_left_expr,indvars,depvars,dict_indvars)
+
+    _right_expr = (ModelingToolkit.simplified_expr(eq.rhs))
+    right_expr = simplified_derivative(_right_expr,indvars,depvars,dict_indvars)
+
+    _f = eval(:(($vars) -> $left_expr - $right_expr))
```
ChrisRackauckas, line 50: we can fix this up later.

src/NeuralNetDiffEq.jl
```diff
@@ -60,19 +60,19 @@ function Base.show(io::IO, A::KolmogorovPDEProblem)
   show(io , A.g)
 end
 
-struct NNPDEProblem{PDESystem,MOLFiniteDifference,P} <:DiffEqBase.DEProblem
+struct discretize{PDESystem,PhysicsInformedNN,P} <:DiffEqBase.DEProblem
```
ChrisRackauckas, line 63: discretize is a function in DiffEqBase that you should be extending.

KirillZubov: ok

src/pinns_pde_solve.jl
```diff
@@ -0,0 +1,250 @@
...
+    end
+
+    #loss function
+    loss(θ) = 1.0f0/τf * loss_domain(θ) + 1.0f0/τb * loss_boundary(θ) #+ 1.0f0/τc * custom_loss(θ)
+
+    cb = function (p,l)
+        verbose && println("Current loss is: $l")
+        l < abstol
+    end
+    res = DiffEqFlux.sciml_train(loss, initθ, opt; cb = cb, maxiters=maxiters, alg.kwargs...)
```
Vaibhavdixit02, line 247: I am registering GalacticOptim https://github.com/JuliaRegistries/General/pull/17124 which is supposed to replace the `sciml_train` interface, you would want to make a note to switch to that

Vaibhavdixit02:
I guess the Readme should be updated as well this looks like a lot of updates, maybe in a follow up PR?

KirillZubov:
> I guess the Readme should be updated as well this looks like a lot of updates, maybe in a follow up PR?

yeah, I think, in a follow PR will better.

src/pinns_pde_solve.jl
KirillZubov: ok, I will keep it in mind

Issue #71, opened 2020-03-05
Investigate physics-informed neural network (PINN) training strategies


ChrisRackauckas:
https://github.com/JuliaDiffEq/NeuralNetDiffEq.jl/issues/46 shows that the basic strategy of training on a fixed training grid doesn't work so well even on a 1-dimensional PINN (an ODE) if the nonlinearity is strong enough. To improve the training process, we should look at changing the strategies a bit. Here's some things that come to mind:

- [x] Random sampling in the domain, making a stochastic loss function to prevent local minima
- [x] Random minibatching
- [x] Starting the training with a small time interval, and successively growing it (user side)
- [x] Growing quasi-Monte Carlo sampling (https://github.com/JuliaDiffEq/QuasiMonteCarlo.jl) (done https://github.com/SciML/NeuralPDE.jl/pull/189)

A new type should be added to the `alg` to allow dispatching on `strategy`.

PR #74, opened 2020-03-15
Stochastic Loss Function

ChrisDeGrendele:
For Issue #71 
The general strategy is:
- Calculate the loss along with a random number between [1,2]
- rand_kick = the random number times a multiplier (x4 to start)
- If our loss (without the kick) remains steady (within 10%) for 5 iterations, we are in a potential well. We want to increase the multiplier to try to get out
- Wait three iterations. Reset the multiplier after we start to decrease again.

However, this isn't working as I might expect. Despite increasing the multiplier, the loss usually has a hard time falling... I was hoping for any advice someone could provide.


ChrisDeGrendele:
image
vis

ChrisDeGrendele:
The code I'm using to test:
`using Revise
 using OrdinaryDiffEq, Plots, Flux
 using NeuralNetDiffEq

 function f(du,u,p,t)
   du[1] = p[1]*u[1] - p[2]*u[1]*u[2]
   du[2] = -p[3]*u[2] + p[4]*u[1]*u[2]
 end
 function f(u,p,t)
   [p[1]*u[1] - p[2]*u[1]*u[2],-p[3]*u[2] + p[4]*u[1]*u[2]]
 end

 p = Float32[1.5,1.0,3.0,1.0]
 u0 = Float32[1.0,1.0]
 prob = ODEProblem(f,u0,(0f0,3f0),p)
 prob_oop = ODEProblem{false}(f,u0,(0f0,3f0),p)

 true_sol = solve(prob,Tsit5())

 opt = ADAM(1e-03)


 chain = Chain(
     x -> reshape(x, length(x), 1, 1),
     MaxPool((1,)),
     Conv((1,), 1=>16, relu),
     Conv((1,), 16=>16, relu),
     Conv((1,), 16=>32, relu),
     Conv((1,), 32=>64, relu),
     Conv((1,), 64=>256, relu),
     Conv((1,), 256=>256, relu),
     Conv((1,), 256=>1028, relu),
     Conv((1,), 1028=>1028),
     x -> reshape(x, :, size(x, 4)),
     Dense(1028, 512, tanh),
     Dense(512, 128, relu),
     Dense(128, 64, tanh),
     Dense(64, 2),
     softmax)

 sol, losses, kicks, multipliers  = solve(prob_oop, NeuralNetDiffEq.NNODE(chain,opt) ,maxiters = 150, verbose = true, dt=1/5f0);

 x = LinRange(1,size(losses)[1], size(losses)[1])
 plot(x, losses, label = "total loss")
 plot!(x,kicks, label = "Size of kick")
 plot!(x, multipliers, label="Multiplier of kick")


 plot(true_sol)
 plot!(sol)`

ChrisDeGrendele:
As of right now, this method is working well with `opt = Flux.ADAM()` but fails to converge with `opt = BFGS()`. I will investigate why this is.

ChrisRackauckas:
Using BFGS on its own usually doesn't work well: I find that you need ADAM to find the well before BFGS can finish it well. Or you need to reduce the initial_stepnorm

ChrisRackauckas:
Also, BFGS needs a deterministic loss function

ChrisDeGrendele:
Does it make sense then to have the following test then in `test/NNODE_tests.jl`?
`
opt = BFGS()
sol = solve(prob, NeuralNetDiffEq.NNODE(chain,opt), dt=1/20f0, verbose = true,
            abstol=1e-10, maxiters = 200)
`

ChrisRackauckas:
The current loss is deterministic. Doesn't make sense anymore if you change it to a stochastic loss.

ChrisRackauckas:
how does this do on LV?

ChrisDeGrendele:
LV

ChrisDeGrendele:
Screen Shot 2020-03-24 at 11 24 13 AM
with:
chain = FastChain(FastDense(1,N,tanh), FastDense(N,N,tanh), FastDense(N,N,tanh), FastDense(N,length(u0)))

ChrisRackauckas:
@KirillZubov you might want to incorporate some of these ideas into the new PINN methods

KirillZubov:
> @KirillZubov you might want to incorporate some of these ideas into the new PINN methods

yeah, thanks, I will definitely return to these ideas

PR #75, opened 2020-03-17
Initialised RODE from the beginning according to new changes in ODE solver.

ashutosh-b-b:


ChrisRackauckas:
This should be ready to merge. Then you know the drill: we need a tutorial and some reference documentation here.

ChrisRackauckas:
Also, the test file needs to be added to runtests.jl

ashutosh-b-b:
> This should be ready to merge. Then you know the drill: we need a tutorial and some reference documentation here.

Yes. I think a tutorial will be necessary here.

PR #76, opened 2020-03-22
Added SDEProblem to Kolmogorov Solver.

ashutosh-b-b:
KolmogorovPDEProblem is now unused and stays there. While the default problem for Kolmogorov Solver is a SDE problem.


src/kolmogorov_solve.jl
```diff
@@ -17,37 +17,56 @@ function DiffEqBase.solve(
...
+    u0 = prob.u0
     #hidden layer
     chain  = alg.chain
     opt    = alg.opt
     sdealg = alg.sdealg
     ps     = Flux.params(chain)
     xi     = rand(xs , d , N[1])
+
+    #Initial Conditions
+    function dirac_delta(u0 , xi)
```
ashutosh-b-b: Used a dirac delta pdf for a constant. Is there any better way to do this?

test/NNKolmogorov_tests.jl
```diff
@@ -1,35 +1,25 @@
...
 tspan = (0.0 , 1.0)
-#f = mu and g  = sigma
-g(u , p , t) = 0.5*u
-f(u , p , t) = 0.5*0.25*u
+g(u , p , t) = 1
+f(u , p , t) = 0
 d = 1
 sdealg = EM()
-prob = KolmogorovPDEProblem(f , g, phi , xspan , tspan, d)
+prob = SDEProblem(f , g , u0 , (0.0 , 1.0) ; xspan = xspan , d = d)
```
ChrisRackauckas: Why was the previous test case removed? It would probably be good to have both?

ashutosh-b-b: I added it again. And kept both the problems now as of the previous discussion

src/kolmogorov_solve.jl
```diff
@@ -17,21 +17,29 @@ function DiffEqBase.solve(
...
-    sigma = prob.sigma
-    μ = prob.μ
-    d = prob.d
+    sigma = prob.g
+    μ = prob.f
+    if prob isa SDEProblem
+        xspan = prob.kwargs.data.xspan
+        d = prob.kwargs.data.d
+        u0 = prob.u0
+        phi(xi) = pdf.(u0 , xi)
```
ChrisRackauckas: this is always going to assume independence. We probably want to handle multivariate distributions as well.

ChrisRackauckas:
Looks good except for the distribution handling comment.

ChrisRackauckas:
We can handle that in a follow up PR.

src/kolmogorov_solve.jl
ashutosh-b-b: Yes I realized this just now when I was conductiing some high dimensional tests. Changing this in the next PR

PR #77, opened 2020-03-25
Support for blackbox optimization coming into DiffEqFlux.jl

ChrisDeGrendele:
https://github.com/JuliaDiffEq/DiffEqFlux.jl/pull/196


src/ode_solve.jl
```diff
@@ -28,7 +29,8 @@ function DiffEqBase.solve(
     adaptive=false,
     abstol = 1f-6,
     verbose = false,
-    maxiters = 100)
+    maxiters = 100,
+    blackbox = false)
```
ChrisRackauckas: this shouldn't be needed: if you just pass an `opt` that's for BBO it'll use BBO. Nothing about this package should be different except the ability to pass bounds.

ChrisRackauckas:
I think this no longer needs to be specialized on because of splatted kwargs, but @ChrisDeGrendele or @KirillZubov can follow this up by adding a test which uses a global optimizer. Cheers!

PR #79, opened 2020-03-30
Added noise_rate_prototype for Non Diagonal Noise.

ashutosh-b-b:
Added noise_rate_prototype as a field for non Diagonal Noise in both SDE and KolmogorovPDE Problems.

### DIFF
--- a/src/kolmogorov_solve.jl
+++ b/src/kolmogorov_solve.jl
@@ -42,7 +42,7 @@ function DiffEqBase.solve(
     ps     = Flux.params(chain)
     xi     = rand(xs ,d ,N[1])
     #Finding Solution to the SDE having initial condition xi. Y = Phi(S(X , T))
-    sdeproblem = SDEProblem(μ,sigma,xi,tspan,noise_rate_prototype)
+    sdeproblem = SDEProblem(μ,sigma,xi,tspan,noise_rate_prototype = noise_rate_prototype)
     sol = solve(sdeproblem, sdealg ,dt=0.01 , save_everystep=false , kwargs...)
     x_sde = sol[end]
     y = phi(x_sde)
--- a/src/kolmogorov_solve.jl
+++ b/src/kolmogorov_solve.jl
@@ -1,9 +1,10 @@
-struct NNKolmogorov{C,O,S} <: NeuralNetDiffEqAlgorithm
+struct NNKolmogorov{C,O,S,E} <: NeuralNetDiffEqAlgorithm
     chain::C
     opt::O
     sdealg::S
+    ensemblealg::E
 end
-NNKolmogorov(chain  ; opt=Flux.ADAM(0.1) , sdealg = EM()) = NNKolmogorov(chain , opt , sdealg)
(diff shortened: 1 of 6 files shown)


src/kolmogorov_solve.jl
```diff
@@ -41,7 +42,7 @@ function DiffEqBase.solve(
     ps     = Flux.params(chain)
     xi     = rand(xs ,d ,N[1])
     #Finding Solution to the SDE having initial condition xi. Y = Phi(S(X , T))
-    sdeproblem = SDEProblem(μ,sigma,xi,tspan)
+    sdeproblem = SDEProblem(μ,sigma,xi,tspan,noise_rate_prototype)
```
ChrisRackauckas, line 45: that's passing it as a parameter?

ChrisRackauckas, line 45:
```suggestion
    sdeproblem = SDEProblem(μ,sigma,xi,tspan,noise_rate_prototype=noise_rate_prototype)
```

ashutosh-b-b: I changed it.

ChrisRackauckas:
This should get a test

src/kolmogorov_solve.jl
```diff
@@ -39,12 +40,23 @@ function DiffEqBase.solve(
...
     #Finding Solution to the SDE having initial condition xi. Y = Phi(S(X , T))
     sdeproblem = SDEProblem(μ,sigma,xi,tspan,noise_rate_prototype = noise_rate_prototype)
-    sol = solve(sdeproblem, sdealg ,dt=0.01 , save_everystep=false , kwargs...)
-    x_sde = sol[end]
+    function prob_func(prob,i,repeat)
+      SDEProblem(prob.f , prob.g , xi[: , i] , prob.tspan ,noise_rate_prototype = prob.noise_rate_prototype)
+    end
+    output_func(sol,i) = (sol[end],false)
+    ensembleprob = EnsembleProblem(sdeproblem , prob_func = prob_func , output_func = output_func)
+    sim = solve(ensembleprob, sdealg, ensemblealg , dt=0.01,trajectories=N[1],adaptive=false)
```
ChrisRackauckas, line 53: you'll still want to do `save_everystep=false,save_start=false` to reduce the memory usage.

test/NNKolmogorov_tests.jl
```diff
@@ -73,4 +71,30 @@ xs2 = xspan2[1]:0.001:xspan2[2]
...
+sdealg3= EM()
+xspan3 = (-10.0 , 10.0)
+tspan3 = (0.0 , 1.0)
+d3 = 2
+prob = SDEProblem(f_noise , g_noise , uo3 , (0.0 , 1.0) ; xspan = xspan3 , d = d3 , noise_rate_prototype=zeros(2,4))
+opt = Flux.ADAM(0.0001)
+m3 = Chain(Dense(d3, 256, elu) ,Dense(256 , 512 , elu), Dense(512 , 1))
+sol3 = solve(prob, NeuralNetDiffEq.NNKolmogorov(m3,opt , sdealg3 , EnsembleThreads()), verbose = true, dt = 0.001,
+            abstol=1e-6, maxiters = 1000)
+println("Non-Diagonal test working.")
```
ChrisRackauckas, line 100: we'l want a full demonstration to check the output values, but this is probably good for now.

PR #84, opened 2020-05-03
Initialising the Optimal Stopping Time Problem

ashutosh-b-b:


kanav99:
You are using so many globals, try to pass them through arguments instead, globals hit performance hard

ashutosh-b-b:
Its an error on my part. All the variables are declared locally. I am changing it in the next commit.

ashutosh-b-b:
Screen Shot 2020-05-12 at 7 34 35 PM

ashutosh-b-b:
@ChrisRackauckas

src/stopping_solve.jl
```diff
@@ -0,0 +1,97 @@
...
+        function Un(n , X )
+            if size(a)[1] >= n
+                return a[n]
+            else
+                if(n == 1)
+                      ans =  first(m(X[1])[1])
+                      a = [ans]
+                      return ans
+                  else
+                      ans =  max(first(m(X[n])[n]) , n + 1 - N)*(1 - sum(Un(i , X ) for i in 1:n-1))
```
ChrisRackauckas, line 54: use a callback for early termination. That should be a lot more efficient: https://docs.sciml.ai/latest/features/callback_functions/#Example-2:-Terminating-an-Integration-1

src/NeuralNetDiffEq.jl
```diff
@@ -56,13 +56,35 @@ function Base.show(io::IO, A::KolmogorovPDEProblem)
   println(io,"Sigma")
   show(io , A.sigma)
 end
+struct OptimalStoppingProblem{ Mu, Sigma, G , U0 , T ,P} <: DiffEqBase.DEProblem
```
ChrisRackauckas, line 59: More generally, this is just an SDE problem with a value function `g`, so I would just make this `SDEProblem` and `g`.

src/stopping_solve.jl
```diff
@@ -0,0 +1,97 @@
...
+    sim = solve(ensembleprob, sdealg, ensemblealg, dt=dt,trajectories=trajectories,adaptive=false)
+    payoff = []
+    times = []
+    iter = 0
+    sump = 0
+    for u in sim.u
+        X = u.u
+        a = []
+        function Un(n , X )
+            if size(a)[1] >= n
```
ChrisRackauckas, line 46: where is `n` defined?

src/stopping_solve.jl
```diff
@@ -0,0 +1,97 @@
...
+    kwargs...
+    )
+
+    tspan = prob.tspan
+    sigma = prob.sigma
+    μ = prob.mu
+    g = prob.g
+    u0 = prob.u0
+    ts = tspan[1]:dt:tspan[2]
+    N = size(ts)[1]
```
ChrisRackauckas, line 27: there's no reason to define `N` here. It just reduces readability because you have to look it up. Instead, just use `length(ts)` where needed.

src/stopping_solve.jl
```diff
@@ -0,0 +1,97 @@
...
+    prob = SDEProblem(μ,sigma,u0,tspan)
+    ensembleprob = EnsembleProblem(prob)
+    sim = solve(ensembleprob, sdealg, ensemblealg, dt=dt,trajectories=trajectories,adaptive=false)
+    payoff = []
+    times = []
+    iter = 0
+    sump = 0
+    for u in sim.u
+        X = u.u
+        a = []
```
ChrisRackauckas, line 44: `a` is a very non-descriptive name

src/stopping_solve.jl
```diff
@@ -0,0 +1,97 @@
...
+    sdealg = alg.sdealg
+    ensemblealg = alg.ensemblealg
+    
+    prob = SDEProblem(μ,sigma,u0,tspan)
+    ensembleprob = EnsembleProblem(prob)
+    sim = solve(ensembleprob, sdealg, ensemblealg, dt=dt,trajectories=trajectories,adaptive=false)
+    payoff = []
+    times = []
+    iter = 0
+    sump = 0
```
ChrisRackauckas, line 41: sumprice? totalprice?

src/stopping_solve.jl
ashutosh-b-b: n is the parameter. its passed here https://github.com/ashutosh-b-b/NeuralNetDiffEq.jl/blob/305a8c0c8a7a867f7c25bb13390ee88c0475d1f6/src/stopping_solve.jl#L62

src/stopping_solve.jl
ashutosh-b-b: I will change this.

src/stopping_solve.jl
ashutosh-b-b: Changes this to un

ashutosh-b-b:
Screen Shot 2020-05-24 at 9 12 55 AM

src/stopping_solve.jl
```diff
@@ -0,0 +1,100 @@
...
+    end
+
+    function loss()
+        reward = 0.00
+        for u in sim.u
+            X = u.u
+            reward = reward + sum(Un(i , X )*g(ts[i] , X[i]) for i in 1 : size(ts)[1])
+            un = []
+        end
+        return 10000 - reward
```
ChrisRackauckas, line 66: why is the 10000 added to it?

ChrisRackauckas:
Awesome, that's looking great! Try a really big train now: see if you can get it almost exact, just to check the convergence.

PR #86, opened 2020-05-22
convert the README to a Documenter documentation

TeAmP0is0N:
Now, we can add tutorials related to `NeuralNetDiffEq`. I have converted `README.md` to `Documenter`. I have not changed anything. If you want any change then please drop a comment.


ChrisRackauckas:
Looks like a great start! I'll merge since it's already an improvement, but before we deploy we'll want to break things apart into pages and all of that.

TeAmP0is0N:
@ChrisRackauckas It's just an initiative for #86. Further, I would like to solve more issues.

ChrisRackauckas:
Awesome. I think the next step would be to split out the tutorials into separate tabs on the side, and make sure they have explanations. Some things from https://mitmath.github.io/18S096SciML/lecture6/sdes and https://mitmath.github.io/18337/lecture15/diffeq_machine_learning can probably be pulled to explain the methods for example. The reference pages should then get descriptions of all of their arguments.

ChrisRackauckas:
The neural network ODE solver has tests (https://github.com/SciML/NeuralNetDiffEq.jl/blob/master/test/NNODE_tests.jl) but doesn't have a tutorial, so one on using that should get written, and the explanation of it is detailed in https://mitmath.github.io/18S096SciML/lecture2/ml

TeAmP0is0N:
@ChrisRackauckas Ok sir. I will complete it soon. Thanks for providing me more details.

PR #87, opened 2020-05-23
described relationship between SDEs and PDE

TeAmP0is0N:
Currently splitting out the tutorials into separate tabs on the side and adding proper explanations to those tutorials.


kanav99:
Awesome stuff, probably change name of the file to not contain spaces, stuff is not good with such names and can lead to unexpected errors. Also a note on curse of dimensionality would be nice too :)

TeAmP0is0N:
> note on curse of dimensionality

Done.

TeAmP0is0N:
> Awesome stuff, probably change name of the file to not contain spaces, stuff is not good with such names and can lead to unexpected errors. Also a note on curse of dimensionality would be nice too :)

done.

docs/src/examples/Relationship_between_SDE_and_PDE.md
```diff
@@ -0,0 +1,93 @@
...
+
+Brownian motion is non-differentiable, and this is fairly straightforward to prove. Take any finite `` h > 0 ``. In that interval, the probability that the Brownian motion has a given value is ``N(0,h)``. Thus it has a positive probability of being positive and a positive probability of being negative. Thus take the sequence ``h/2, h/4, `` etc. to have infinetly many subintervals. With probability 1 at least one of those values will be positive and at least one will be negative, and so for every ``h`` there is a positive or negative value, which means **Brownian Motion** is always neither increasing or decreasing. This means that, if the classical derivative existed, it must be zero, which means it's a constant function, but it's not, so by contradiction Brownian motion cannot have a classical derivative!
+
+However, with a bunch of analysis one can derive an appropriate calculus on SDEs. Take the SDE
+
+```Math
+dx = f(x,t)dt + \sum_{i=1}^{n}g_{i}(x,t)dW_{i}
+```
+
+where ``W_{i}(t)`` is a standard Brownian motion. It0's Rules (the stochastic Chain) rule could be interpreted as:
```
Vaibhavdixit02, line 33:
```suggestion
where ``W_{i}(t)`` is a standard Brownian motion. Ito's Rules (the stochastic Chain) rule could be interpreted as:

```

docs/src/examples/Relationship_between_SDE_and_PDE.md
```diff
@@ -0,0 +1,93 @@
...
+dy=\frac{\partial\psi}{\partial t}dt+\frac{\partial\psi}{\partial x}dx+\frac{1}{2}\frac{\partial^{2}\psi}{\partial x^{2}}\left(dx\right)^{2}
+```
+
+where, if we plug in ``dx``, we get
+
+```math
+dy=d\psi(x,t)=\left(\frac{\partial\psi}{\partial t}+f(x,t)\frac{\partial\psi}{\partial x}+\frac{1}{2}\sum_{i=1}^{n}g_{i}^{2}(x,t)\frac{\partial^{2}\psi}{\partial x^{2}}\right)dt+\frac{\partial\psi}{\partial x}\sum_{i=1}^{n}g_{i}(x,t)dW_{i}
+```
+
+Notice that this is the same as the normal chain rule formula, but now there is a seconf order correction to the main variable. The soltuion is given by the integral form:
```
Vaibhavdixit02, line 62:
```suggestion
Notice that this is the same as the normal chain rule formula, but now there is a seconf order correction to the main variable. The solution is given by the integral form:

```

TeAmP0is0N: Yeah, Sorry. Thanks for reviewing and your feedback.

docs/src/examples/Relationship_between_SDE_and_PDE.md
```diff
@@ -0,0 +1,93 @@
...
+
+```math
+X_n+1 = X_n + f(X_n,t_n)h + sqrt(h)g(X_n,t_n)\zeta
+```
+
+where ``\zeta ~ N(0,1)`` is a standard normal random variable ``(randn())``. Thus a stochastic differential equation is an ordinary differential equation with a small stochastic perturbation at every point in time (continuously!).
+
+There are many definitions of **Brownian Motion**. One of the easiest to picture is that it's the random walk where at every infinitesimal ``dt`` you move by ``N(0,dt)`` (this is formally true in non-standard analysis, see **Radically Elementary Probability Theory** for details). Another way to think of it is as a limit of standard random walks. Let's say you have a random walk were every time step of ``h`` you move ``\Delta x`` to the right or left. If you let ``h/\Delta x = 1`` and send ``h \rightarrow 0``, then the limiting process is a **Brownian Motion**. All of the stuff about normal distributions just comes from the central limit theorem and the fact that you're doing infinitely many movements per second!
+
+## Wiener Process Calculus Summarized
```
ChrisRackauckas, line 23: not sure this is necessary at all? It's somewhat orthogonal to the understanding of what the method is solving and how to use the method.

TeAmP0is0N: @ChrisRackauckas It's just an explanation to tell that there are more Brownian Motion and had tried to give a view about Radically Elementary Probability Theory. If you think that it's not necessary then I will remove it? Waiting for your instruction.

PR #90, opened 2020-05-29
Create NN_ODE_Solver_Using_Test.md

TeAmP0is0N:
A Method for Solving Ordinary Differential Equations with Neural Networks using test


docs/src/examples/NN_ODE_Solver_Using_Test.md
```diff
@@ -0,0 +1,237 @@
...
+
+If `NN(t)` was the true solution, then it would hold that `NN'(t) = f(NN(t),t)` for all `t`. Thus we turn this condition into our loss function. This motivates the loss function:
+
+```math
+L(p) = \sum_i \left(\frac{dNN(t_i)}{dt} - f(NN(t_i),t_i) \right)^2
+```
+
+The choice of `t_i` could be done in many ways: it can be random, it can be a grid, etc. Anyways, when this loss function is minimized (gradients computed with standard reverse-mode automatic differentiation), then we have that `\frac{dNN(t_i)}{dt} \approx f(NN(t_i),t_i)` and thus `NN(t)` approximately solves the differential equation. 
+
+## Quick Question for Understanding: What is an effective way to compute dN/dt?
```
ChrisRackauckas, line 31: this doesn't make sense in documentation

TeAmP0is0N: So, still, we have to replace it or remove it?

TeAmP0is0N: Done!

PR #93, opened 2020-06-02
Add trajectories to Kolmogorov Solver and fix bugs.

ashutosh-b-b:
1. Add trajectories to the solver directly
2. Specify algorithm to stopping_solve.jl and kolmogorov_solve.jl to avoid wrong solver calls for SDEProblem


src/kolmogorov_solve.jl
```diff
@@ -32,7 +33,7 @@ function DiffEqBase.solve(
         phi = prob.phi
     end
     ts = tspan[1]:dt:tspan[2]
-    xs = xspan[1]:0.001:xspan[2]
+    xs = xspan[1]:0.0001:xspan[2]
```
ChrisRackauckas, line 36: this shouldn't be hardcoded.

PR #96, opened 2020-06-16
Add Primal-Dual Algorithm for BSDEs

ashutosh-b-b:


ashutosh-b-b:
@ChrisRackauckas  I did run some simulations over T (for 1.00 , 1.5 , 2.00) and some times the lower limit was higher than the BSDE output. Does this mean the NN wasn't trained correctly?

ashutosh-b-b:
@ChrisRackauckas Also, I am shifting the discussion about the `u0` being less than 1.0 instead of the values being according to the paper.

ChrisRackauckas:
You played with the networks and such and it's just not converging well at all?

@KirillZubov might want to get up to date on this.

ashutosh-b-b:
> You played with the networks and such and it's just not converging well at all?

Umm.. The loss is reducing and arriving at a minima. Its just the u0(x0) should be different.

ChrisRackauckas:
Is it close? Is it converging with different decreasing dt? etc

ashutosh-b-b:
> Is it close? Is it converging with different decreasing dt? etc

But shouldn't it atleast be of that order? it is less than 1 while in paper its greater than 10

ChrisRackauckas:
yes, that's quite odd. This is worth debugging before continuing.

ChrisRackauckas:
Since the tests on the other people seem to be more stringent, I would assume it's generally working, and triple check how the ideas are translated over to here. Maybe there's another sign difference or something.

ashutosh-b-b:
> Since the tests on the other people seem to be more stringent, I would assume it's generally working, and triple check how the ideas are translated over to here. Maybe there's another sign difference or something.

Yes. Is there something about Ucva (the pde form as in the paper) I can dig into? I searched and found some papers by the same author and rechecked if I was doing it correctly.. It would be helpful if you had some resources perhaps?

ChrisRackauckas:
Make sure to read https://www.pnas.org/content/115/34/8505 do be acquainted with the algorithm implemented here.

ashutosh-b-b:
> Make sure to read https://www.pnas.org/content/115/34/8505 do be acquainted with the algorithm implemented here.

Yes I have given this a read earlier. But I ll see this once again.

ashutosh-b-b:
@ChrisRackauckas I read the above paper again. And came back to the orignal paper. (https://papers.ssrn.com/sol3/papers.cfm?abstract_id=3071506) and read the Numerical Example 2. In numerical example 2 The primal problem didnot required a legendre transform. The optimal control was obtained by using ` ∂u/ ∂xi` for each dimension i but we simply have `∇u` instead i.e. the denominator.(the last formula in the image)
Screen Shot 2020-06-19 at 12 15 01 PM

ChrisRackauckas:
> The optimal control was obtained by using ∂u/ ∂xi for each dimension i but we simply have ∇u instead i.e. the denominator.(the last formula in the image)

We should have advection terms in our BSDE?

src/pde_solve.jl
```diff
@@ -59,5 +60,78 @@ function DiffEqBase.solve(
...
+                    t = ts[i]
+                    _σᵀ∇u = σᵀ∇u[i](xsde[i])
+                    dW = sqrt(dt)*randn(d)
+                    U = U .+ f(xsde[i], u, _σᵀ∇u, p, t)*dt .- _σᵀ∇u'*dW
+                end
+                Uo = vcat(Uo , U)
+            end
+            Uo
+        end
+        loss_() = sum(sol_high())/800
```
ChrisRackauckas, line 88: why divide by 800?

ChrisRackauckas: and no norm or abs on this?

ashutosh-b-b: I will change 800 to user defined trajectories. In paper the sum is minimized without abs().

ashutosh-b-b: Do we have to add a norm or abs?

ChrisRackauckas: just took a look at the paper, yeah no abs. Interesting. So just change to number of trajectories

ashutosh-b-b:
@ChrisRackauckas

src/pde_solve.jl
```diff
@@ -59,5 +66,87 @@ function DiffEqBase.solve(
...
+            Uo = []
+            for u in sim.u
+                xsde = u.u
+                U = g(xsde[end])
+                u = u0(X0)[1]
+                for i in length(ts):-1:3
+                    t = ts[i]
+                    _σᵀ∇u = σᵀ∇u[i-1](xsde[i-1])
+                    dW = sqrt(dt)*randn(d)
+                    U = U .+ f(xsde[i-1], U, _σᵀ∇u, p, t)*dt .- _σᵀ∇u'*dW
```
ChrisRackauckas, line 91: are the dW supposed to be pulled from the forward pass here?

src/pde_solve.jl
```diff
@@ -59,5 +66,87 @@ function DiffEqBase.solve(
...
+        ps = Flux.params(u0, σᵀ∇u...)
+        cb = function ()
+            l = loss_()
+            true && println("Current loss is: $l")
+            l < 1e-6 && Flux.stop()
+        end
+        dataS = Iterators.repeated((), maxiters_upper)
+        Flux.train!(loss_, ps, dataS, ADAM(0.01); cb = cb)
+        u_high = loss_()
+        println(u_high)
```
ChrisRackauckas, line 109:
```suggestion
```

src/pde_solve.jl
```diff
@@ -59,5 +66,87 @@ function DiffEqBase.solve(
...
+                    f_matrix = give_f_matrix(X , u_domain, _σᵀ∇u, p, ts[i])
+                    a_ = A[findmax(collect(A).*u .- collect(legendre_transform(f_matrix, a, u_domain) for a in A))[2]]
+                    I = I + a_*dt
+                    Q = Q + exp(I)*legendre_transform(f_matrix, a_, u_domain)
+                end
+                I , Q , X
+            end
+        end
+        u_low = sum(exp(I)*g(X) - Q for (I ,Q ,X) in sol_low())/(m2)
+        println(u_low)
```
ChrisRackauckas, line 148:
```suggestion
```

src/pde_solve.jl
ashutosh-b-b: There is nothing specifically written. The symbols are same though. But problem was that if we take the same the trajectories cannot be taken different..

ChrisRackauckas: oh yes I see, then it'll go right back, so this is quantifying process uncertainty.

ashutosh-b-b: Should I change it then?

ChrisRackauckas: no you're good

src/pde_solve.jl
```diff
@@ -59,5 +66,87 @@ function DiffEqBase.solve(
...
+                    _σᵀ∇u = σᵀ∇u[i-1](xsde[i-1])
+                    dW = sqrt(dt)*randn(d)
+                    U = U .+ f(xsde[i-1], U, _σᵀ∇u, p, t)*dt .- _σᵀ∇u'*dW
+                end
+                Uo = vcat(Uo , U)
+            end
+            Uo
+        end
+
+        loss_() = sum(sol_high())/m1
```
ChrisRackauckas, line 98:
```suggestion
        loss_() = sum(sol_high())/trajectories_upper
```

no need for two names

src/pde_solve.jl
```diff
@@ -59,5 +66,87 @@ function DiffEqBase.solve(
...
-    save_everystep ? iters : u0(X0)[1]
+
+    if give_limit == false
+        save_everystep ? iters : u0(X0)[1]
+    else
+        A = prob.kwargs.data.A
+        u_domain = prob.kwargs.data.u_domain
+
+        ## UPPER LIMIT
+        m1 = trajectories_upper
```
ChrisRackauckas, line 77:
```suggestion
```

src/pde_solve.jl
```diff
@@ -59,5 +66,87 @@ function DiffEqBase.solve(
...
+            end
+            return a
+        end
+        #The legendre transform that uses the precalculated f values.
+        function legendre_transform(f_matrix , a , urange)
+            le = a.*(collect(urange)) .- f_matrix
+            return maximum(le)
+        end
+
+        m2 = trajectories_lower
```
ChrisRackauckas, line 126:
```suggestion
```

src/pde_solve.jl
```diff
@@ -59,5 +66,87 @@ function DiffEqBase.solve(
...
+        function sol_low()
+            map(1:m2) do j
```
ChrisRackauckas, line 128:
```suggestion
            map(1:trajectories_lower) do j
```

src/pde_solve.jl
```diff
@@ -59,5 +66,87 @@ function DiffEqBase.solve(
...
+                u = u0(X0)[1]
+                X = X0
+                I = 0.0
```
ChrisRackauckas, line 131:
```suggestion
                I = zero(eltype(u))
```

and the same on all other zeros. Many times they should be Float32 and not Float64

src/pde_solve.jl
```diff
@@ -59,5 +66,87 @@ function DiffEqBase.solve(
...
+                    X  = X .+ μ(X,p,t)*dt .+ σ(X,p,t)*dW
+                    f_matrix = give_f_matrix(X , u_domain, _σᵀ∇u, p, ts[i])
+                    a_ = A[findmax(collect(A).*u .- collect(legendre_transform(f_matrix, a, u_domain) for a in A))[2]]
+                    I = I + a_*dt
+                    Q = Q + exp(I)*legendre_transform(f_matrix, a_, u_domain)
+                end
+                I , Q , X
+            end
+        end
+        u_low = sum(exp(I)*g(X) - Q for (I ,Q ,X) in sol_low())/(m2)
```
ChrisRackauckas, line 147:
```suggestion
        u_low = sum(exp(I)*g(X) - Q for (I ,Q ,X) in sol_low())/(trajectories_lower)
```

src/pde_solve.jl
```diff
@@ -59,5 +66,87 @@ function DiffEqBase.solve(
...
+        A = prob.kwargs.data.A
+        u_domain = prob.kwargs.data.u_domain
+
+        ## UPPER LIMIT
+        m1 = trajectories_upper
+        sdeProb = SDEProblem(μ , σ , X0 , prob.tspan)
+        ensembleprob = EnsembleProblem(sdeProb)
+        sim = solve(ensembleprob, sdealg, ensemblealg, dt=dt,trajectories=m1,adaptive=false)
+        function sol_high()
+            Uo = []
```
ChrisRackauckas, line 82: This should be strictly typed or else I can see this causing some performance issues. `eltype(sim.u)[]`

src/pde_solve.jl
```diff
@@ -59,5 +66,87 @@ function DiffEqBase.solve(
...
+        end
+        dataS = Iterators.repeated((), maxiters_upper)
+        Flux.train!(loss_, ps, dataS, ADAM(0.01); cb = cb)
+        u_high = loss_()
+        println(u_high)
+        ##Lower Limit
+
+        # Function to precalculate the f values over the domain
+        function give_f_matrix(X,urange,σᵀ∇u,p,t)
+            a = []
```
ChrisRackauckas, line 114: strictly type

ChrisRackauckas: actually, doing this as a map would be easier anyways

src/pde_solve.jl
ChrisRackauckas: doing the following as a map and then `reduce(vcat,U)` could be an easy way to generically handle it.

src/NeuralNetDiffEq.jl
```diff
@@ -4,24 +4,25 @@ using Reexport, Statistics
...
 
-struct TerminalPDEProblem{G,F,Mu,Sigma,X,T,P} <: DiffEqBase.DEProblem
+struct TerminalPDEProblem{G,F,Mu,Sigma,X,T,P,K} <: DiffEqBase.DEProblem
     g::G
     f::F
     μ::Mu
     σ::Sigma
     X0::X
     tspan::Tuple{T,T}
     p::P
```
ChrisRackauckas, line 19: make a section for `A` in here and then have the constructor specifically have a kwarg for `A`.

src/pde_solve.jl
```diff
@@ -59,5 +66,87 @@ function DiffEqBase.solve(
...
+        save_everystep ? iters : u0(X0)[1]
+    else
+        A = prob.kwargs.data.A
+        u_domain = prob.kwargs.data.u_domain
+
+        ## UPPER LIMIT
+        m1 = trajectories_upper
+        sdeProb = SDEProblem(μ , σ , X0 , prob.tspan)
+        ensembleprob = EnsembleProblem(sdeProb)
+        sim = solve(ensembleprob, sdealg, ensemblealg, dt=dt,trajectories=m1,adaptive=false)
```
ChrisRackauckas, line 80: the kwargs should splat into here for the solver.

ChrisRackauckas, line 80:
```suggestion
        sim = solve(ensembleprob, sdealg, ensemblealg, dt=dt,trajectories=trajectories_upper,adaptive=false,kwargs...)
```

ChrisRackauckas:
Looking great! Lots of minor comments, but nothing big. Was this tested on GPUs?

ashutosh-b-b:
> Looking great! Lots of minor comments, but nothing big. Was this tested on GPUs?

No not on gpu. Is BSDE solver compatible with GPU though?

ChrisRackauckas:
yes, since the SDE solvers are. This should get thoroughly tested (and add CI tests for it), and if it fails, let's get it fixed.

ashutosh-b-b:
> yes, since the SDE solvers are. This should get thoroughly tested (and add CI tests for it), and if it fails, let's get it fixed.

The BSDE solver uses SDE in this fashion https://github.com/SciML/NeuralNetDiffEq.jl/blob/6ca75a08b9ae73c0ebe586862cbd0eec867cf73c/src/pde_solve.jl#L48 should we change this to ensembleproblem?

ashutosh-b-b:
@ChrisRackauckas I added `u_domain` field as well. Will open a [WIP] PR for the GPU support. I have opened an issue here FluxML/Zygote.jl#715

src/pde_solve.jl
```diff
@@ -59,5 +66,80 @@ function DiffEqBase.solve(
...
+    if give_limit == false
+        save_everystep ? iters : u0(X0)[1]
+    else
+        A = prob.A
+        u_domain = prob.u_domain
+
+        ## UPPER LIMIT
+        sdeProb = SDEProblem(μ , σ , X0 , prob.tspan)
+        ensembleprob = EnsembleProblem(sdeProb)
+        sim = solve(ensembleprob, sdealg, ensemblealg, dt=dt,trajectories=trajectories_upper,adaptive=false,prob.kwargs...)
```
ChrisRackauckas, line 79: why adaptive=false?

ChrisRackauckas, line 79:
```suggestion
        sim = solve(ensembleprob, EM(), ensemblealg, dt=dt, trajectories=trajectories_upper,prob.kwargs...)
```

I think for NNPDEHan it makes sense to hardcode it as `EM` since the method itself is essentially all about Euler-Maruyama discretizations. For the other deep BSDE method we should do it more generically, but that can follow in a different PR.

ashutosh-b-b: to explicitly provide a `dt`.

PR #97, opened 2020-06-17
Add GPU support Kolmogorov Solver

ashutosh-b-b:


ashutosh-b-b:
@ChrisRackauckas This should provide a GPU support.

src/kolmogorov_solve.jl
```diff
@@ -60,7 +61,14 @@ function DiffEqBase.solve(
         x_sde = hcat(x_sde , u)
     end
     y = phi(x_sde)
+    if use_gpu == true
+        y = y |>gpu
+        xi = xi |> gpu
```
ChrisRackauckas, line 66: Our convention is to match the user intention, so instead adapt it to match `u0`

ashutosh-b-b: But `u0` will either be a distribution  or there exists no `u0` for KolmogorovPDEProblem?

ChrisRackauckas: what's the user data that's an array?

ashutosh-b-b: There's no array input. Two tuples though. `xspan` and `tspan`

ashutosh-b-b: I thought the only way to know is to take a boolean itself to check if the user wants to use a GPU

ChrisRackauckas: I see, this is fine then.

ChrisRackauckas:
@christopher-dG could we get GPU CI setup on this repo? It happened sooner than I thought it would 😆

christopher-dG:
Which test group should be run?

ChrisRackauckas:
Make a new one for GPU. We can add stuff into there.

PR #99, opened 2020-06-17
Set up GitLab CI

christopher-dG:


christopher-dG:
Grrrrr the status is still not being updated (the current one is from the juliadiffeq org which I've just deleted and disabled mirroring). Here's the actual build https://gitlab.com/JuliaGPU/NeuralNetDiffEq.jl/-/jobs/600280877

christopher-dG:
I generated a new token and updated the GitLab settings to use that token and now it works. So I guess I'll just do that for all the other repos too 🙃

PR #102, opened 2020-06-23
HJB example code and dependency install file.

zzj0402:


README.md
```diff
@@ -7,15 +7,21 @@
...
+neural networks (PINNs) and deep BSDE solvers. It utilizes techniques like deep neural networks and neural
 stochastic differential equations to make it practical to solve high dimensional PDEs efficiently through the
 likes of scientific machine learning (SciML).
 
-# Examples
+## Installation
 
-## DeepBSDE Solver
+```bash
+julia ./install.jl
```
ChrisRackauckas, line 17: This is very non-standard and we should stick to standard techniques, e.g. using the Pkg manager REPL (Also, this doesn't work on all operating systems)

ChrisRackauckas:
This should probably move towards populating and deploying the documentation?

https://github.com/SciML/NeuralNetDiffEq.jl/tree/master/docs/src

I'm not sure the README should get much work knowing it will get deleted soon.

ChrisRackauckas:
See https://github.com/SciML/NeuralNetDiffEq.jl/issues/85 for details.

README.md
kanav99: You can make a new environment for `examples` folder and add the dependencies to the Project.toml

ChrisRackauckas: We should probably just have a tutorials section of the documentation like is done in DiffEqFlux, DataDrivenDiffEq, etc.

zzj0402: Okay, working it.

PR #109, opened 2020-07-08
[WIP] Adding Docstrings to the code.

ashutosh-b-b:


src/NeuralNetDiffEq.jl
```diff
@@ -36,6 +47,18 @@ function Base.show(io::IO, A::TerminalPDEProblem)
...
+A standard Kolmogorov PDE Problem.
+# Arguments
+* `f` : The drift function from Feynman-Kac representation of the PDE.
+* `g` : The noise function from Feynman-Kac representation of the PDE.
+* `phi` : The terminal condition for the PDE.
+* `tspan`: The timespan for the problem.
+* `xspan`: The xspan for the problem.
+* `d`: The dimensions of the input x.
+* `noise_rate_prototype`: A prototype type instance for the noise rates, that is the output g.
+"""
```
ChrisRackauckas, line 62:
```suggestion
```

src/NeuralNetDiffEq.jl
```diff
@@ -36,6 +47,18 @@ function Base.show(io::IO, A::TerminalPDEProblem)
   print(io,"timespan: ")
   show(io,A.tspan)
 end
+"""
```
ChrisRackauckas, line 50:
```suggestion
end

"""
```

ashutosh-b-b:
@Chris should I add about the upper and lower limit as well in the `TerminalPDEProblem` or should I add it to the NNPDEHan docstring?

ChrisRackauckas:
It should be documented in `TerminalPDEProblem` and we should make sure to update the other solver to allow for bounds as well.

ashutosh-b-b:
> It should be documented in `TerminalPDEProblem` and we should make sure to update the other solver to allow for bounds as well.

Which other solver?

ChrisRackauckas:
NNPDENS

ChrisRackauckas:
Let's merge and keep moving. @zzj0402 should make note of this.

ChrisRackauckas:
Another PR should add the other missing docstrings and explain the lower/upper bounds parts.

PR #111, opened 2020-07-10
Add Upper and Lower Bounds for the NNPDENS.

ashutosh-b-b:


src/pde_solve_ns.jl
```diff
@@ -102,5 +111,75 @@ function DiffEqBase.solve(
...
+
+    if give_limit == false
+        save_everystep ? iters : re1(p3)(X0)[1]
+    else
+    ## UPPER LIMIT
+        sdeProb = SDEProblem(μ , σ , x0 , tspan , noise_rate_prototype = zeros(Float32,d,d))
+        ensembleprob = EnsembleProblem(sdeProb)
+        sim = solve(ensembleprob, EM(), EnsembleThreads(), dt=dt, trajectories=trajectories_upper,kwargs...)
+        ts = tspan[1]:dt:tspan[2]
+        function sol_high()
```
ChrisRackauckas, line 123: don't write this out: instead write this as an SDEProblem ensemble and solve it with the algorithm the user gives.

src/pde_solve_ns.jl
```diff
@@ -102,5 +111,75 @@ function DiffEqBase.solve(
...
 
-    save_everystep ? iters : re1(p3)(X0)[1]
+
+    if give_limit == false
+        save_everystep ? iters : re1(p3)(X0)[1]
+    else
+    ## UPPER LIMIT
+        sdeProb = SDEProblem(μ , σ , x0 , tspan , noise_rate_prototype = zeros(Float32,d,d))
+        ensembleprob = EnsembleProblem(sdeProb)
+        sim = solve(ensembleprob, EM(), EnsembleThreads(), dt=dt, trajectories=trajectories_upper,kwargs...)
```
ChrisRackauckas, line 121: use the algorithm given by the user

src/pde_solve_ns.jl
```diff
@@ -102,5 +111,75 @@ function DiffEqBase.solve(
...
+        end
+
+        #The legendre transform that uses the precalculated f values.
+        function legendre_transform(f_matrix , a , urange)
+            le = a.*(collect(urange)) .- f_matrix
+            return maximum(le)
+        end
+
+        function sol_low()
+            map(1:trajectories_lower) do j
```
ChrisRackauckas, line 163: again, don't discretize it

ashutosh-b-b: Do u mean that I should use the `EnsembleProblem` for lower bound as well?

ashutosh-b-b:
@ChrisRackauckas

src/pde_solve_ns.jl
```diff
@@ -13,16 +13,25 @@ function DiffEqBase.solve(
     maxiters = 300,
     trajectories = 100,
     alg,
+    dt ,
```
ChrisRackauckas, line 16: a dt is not required

ChrisRackauckas, line 16:
```suggestion
```

src/pde_solve_ns.jl
```diff
@@ -102,5 +110,78 @@ function DiffEqBase.solve(
 
     Flux.train!(loss_n_sde, ps, data, opt; cb = cb)
 
-    save_everystep ? iters : re1(p3)(X0)[1]
+
+    if give_limit == false
+        save_everystep ? iters : re1(p3)(X0)[1]
+    else
```
ChrisRackauckas, line 116: dt is required for the part, so check if `dt in keys(kwargs) || error("`dt` choice is required for lower bound calculation")`

ashutosh-b-b: If I dont pass `dt` it throws this error : `Fixed timestep methods require a choice of dt or choosing the tstops` thats why I added the dt to this https://github.com/SciML/NeuralNetDiffEq.jl/blob/61bbef40706690515dc6022590b8139127f75a0e/src/pde_solve_ns.jl#L85

ChrisRackauckas: oh yes, if you use a fixed timestep method it needs `dt`, but that's why the kwargs are splatted

ashutosh-b-b: Oh, they were there earlier, but even before in all the test cases `dt` was passed which I found strange so I added `dt` as a direct argument..

ashutosh-b-b: The test gave the same error. Should I restore the `dt` ?

ChrisRackauckas: Make `dt` default to `eltype(concrete_prob(_prob).tspan)(0)` and then pass it along, and when you use it explicitly, check if it's zero and throw an error. This will allow the adaptive algorithms to automatically set it but also have it explicit in this code.

ashutosh-b-b: I did passed default as `0.0` but it still gives the same error.

ChrisRackauckas: no, take it as input. Do what you were doing before, but with a default value to `dt` of zero.

ashutosh-b-b: Oh.. ok.

PR #119, opened 2020-07-19
system of pde

KirillZubov:


ChrisRackauckas:
Update for the name change

ChrisRackauckas:
Maybe @AlthausKonstantin can chime in

PR #124, opened 2020-07-21
improve boundary condition

KirillZubov:
in addition to the Dirichlet condition, we can now use any condition in any form.

### DIFF
--- a/src/NeuralPDE.jl
+++ b/src/NeuralPDE.jl
@@ -93,13 +93,15 @@ The PINNs Problem.
 * `train_sets`: Training sets on the domain and boundary spaces
 * `dim`: The dimensions of the problem.
 """
-struct NNPDEProblem{PDEFunction,TS} <:DiffEqBase.DEProblem
+struct NNPDEProblem{PDEFunction,Function,TS} <:DiffEqBase.DEProblem
   pde_func::PDEFunction
+  bc_func:: Array{Function}
   train_sets ::TS
   dim :: Int64
-  NNPDEProblem(pde_func,train_sets,dim) = new{typeof(pde_func),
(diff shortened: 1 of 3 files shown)


src/NeuralPDE.jl
```diff
@@ -93,20 +93,24 @@ The PINNs Problem.
 * `train_sets`: Training sets on the domain and boundary spaces
 * `dim`: The dimensions of the problem.
 """
-struct NNPDEProblem{PDEFunction,TS} <:DiffEqBase.DEProblem
+struct NNPDEProblem{PDEFunction,Function,TS} <:DiffEqBase.DEProblem
```
ChrisRackauckas, line 96: Functions probably shouldn't be held in arrays. This is going to lead to some nasty instabilities. Since the number of BC relations should be fixed, maybe a tuple makes sense? This is a detail though.

ChrisRackauckas:
I think it's fine to merge without that change. That could lead to some performance issues though, but it's best to profile before trying to nail that down since it might not matter when considering the size of neural networks.

src/NeuralPDE.jl
KirillZubov: it can be one function that returns array/tuple

KirillZubov: @ChrisRackauckas but I don't see a difference

ChrisRackauckas: arrays of functions aren't type stable so it'll lead to type inference issues. That said, you shouldn't have tuples of more than length 16, for making it a tuple could be an issue too. This is probably something to test out before working it out.

KirillZubov: ok

PR #127, opened 2020-07-23
Stability for RODE tests

ashutosh-b-b:
This should fix the convergence issue. I think we should run it multiple times on travis first.


ChrisRackauckas:
time failure

ashutosh-b-b:
> time failure

I think `NNODE` tests should be split or reduced perhaps?

ChrisRackauckas:
Yes, probably split out random ODE parts.

PR #130, opened 2020-08-04
Slight refactoring of PINNs

KirillZubov:
NNPDEProblem -> OptimizationProblem

### DIFF
--- a/src/pinns_pde_solve.jl
+++ b/src/pinns_pde_solve.jl
@@ -351,7 +351,7 @@ function DiffEqBase.discretize(pde_system::PDESystem, discretization::PhysicsInf
     # get loss_function
     loss_function = get_loss_function(pde_func,bc_funcs,train_sets)
 
-	return GalacticOptim.OptimizationProblem(loss_function, zeros(dim), p=nothing)
+	return GalacticOptim.OptimizationProblem(loss_function, zeros(dim))
 end
 
 function DiffEqBase.solve(
--- a/src/pinns_pde_solve.jl
+++ b/src/pinns_pde_solve.jl
@@ -293,35 +293,69 @@ function generator_training_sets(domains, discretization, bound_args, dict_indva
     [train_domain_set,train_bound_set]
 end
 
-# Convert a PDE problem into an PINNs problem
+function get_loss_function(pde_func, bc_funcs, train_sets)
+    # the points in the domain and on the boundary
+    train_domain_set, train_bound_set = train_sets
+
+    # coefficients for loss function
+    τb = length(train_bound_set)
+    τf = length(train_domain_set)
+
+    # Loss function for pde equation
+    function inner_domain_loss(phi, x, θ, derivative)
+        sum(pde_func(phi, x, θ, derivative))
+    end
+
+    # Loss function for initial and boundary conditions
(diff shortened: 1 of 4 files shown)


src/pinns_pde_solve.jl
```diff
@@ -293,35 +293,69 @@ function generator_training_sets(domains, discretization, bound_args, dict_indva
...
+    bc_funcs = eval.(extract_bc_funcs)
 
     # generate training sets
     train_sets = generator_training_sets(domains, discretization, bound_args, dict_indvars)
 
-    return NNPDEProblem(_pde_func,_bc_funcs, train_sets, dim)
+    # get loss_function
+    loss_function = get_loss_function(pde_func,bc_funcs,train_sets)
+
+	return GalacticOptim.OptimizationProblem(loss_function, zeros(dim), p=nothing)
```
Vaibhavdixit02, line 354: Not an issue but not passing in `p` as `nothing` should work here

Vaibhavdixit02:
Not sure if `OptimizationFunction` could be used here, where do you keep the derivatives of the loss function?

src/pinns_pde_solve.jl
KirillZubov: yes, thanks

KirillZubov:
> Not sure if `OptimizationFunction` could be used here, where do you keep the derivatives of the loss function?

As long as we use the numeric derivative, this is not necessary, but further to support AD it will probably be convenient.

KirillZubov:
@Vaibhavdixit02 also, I have the problem with add GalactiOptim. We get the error. 
`ERROR: Unknown package Zygote found in the compatibility requirements of GalacticOptim `

Vaibhavdixit02:
@KirillZubov that doesn't look good, can you share your `versioninfo`?  @abhigupta768 have you seen this?

KirillZubov:
@Vaibhavdixit02 
```
julia> versioninfo()
Julia Version 1.4.2
Commit 44fa15b150* (2020-05-23 18:35 UTC)
Platform Info:
  OS: macOS (x86_64-apple-darwin18.7.0)
  CPU: Intel(R) Core(TM) i5-3427U CPU @ 1.80GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-8.0.1 (ORCJIT, ivybridge)
Environment:
  JULIA_EDITOR = atom  -a
  JULIA_NUM_THREADS = 2
  JULIA_NUM_THREADS = 2
```

KirillZubov:
@Vaibhavdixit02  @ChrisRackauckas, I copied the OptimizationProblem and removed the dependence on the GalacticOptim for a while until the problem with add GalacticOptim will solve.

abhigupta768:
I can not recreate the error with master, it's probably an issue with the release. @KirillZubov can you check with the master for your use?

KirillZubov:
@abhigupta768 yes, it works: `add GalacticOptim#master`

ChrisRackauckas:
let's do a 0.2 tag today then

ChrisRackauckas:
The issue was something in the registry: https://github.com/JuliaRegistries/General/pull/19399

ChrisRackauckas:
This isn't quite there yet because this package should not define a solve for an optimization problem, but it's moving in the right direction.

PR #133, opened 2020-08-13
Generalize parser and add access to the inner functions

KirillZubov:
- generalized parser and make it more legible
- added access to the low-level API: `build_loss_function`, `generate_training_sets`, and `get_loss_function` in order to get users more info about inner representation and give the ability to create a custom loss function and training sets. But it is the first step to this way
- added Stochastic Loss Function
- removed NNODE and solver overload
- refactoringdiscretize and PhysicsInformedNN
- use GalacticOptim solver


test/NNPDE_tests.jl
```diff
@@ -19,30 +26,45 @@ bcs = [u(0.,θ) ~ 1.0 , u(1.,θ) ~ 1.202]
...
 pde_system = PDESystem(eq,bcs,domains,[t],[u])
 prob = NeuralPDE.discretize(pde_system,discretization)
-alg = NeuralPDE.NNDE(chain,opt,autodiff=false)
-phi,res = solve(prob,alg,verbose=true, maxiters=1000)
+
+initθ = discretization.initθ
+p = DiffEqBase.NullParameters()
+loss = (θ) -> prob.f(θ,p)
+
+res1 = DiffEqFlux.sciml_train(loss, initθ, Flux.ADAM(0.1); cb = cb, maxiters=1000)
```
Vaibhavdixit02, line 49: Can you open an issue in GalacticOptim for Flux optimisers? 
I had read somewhere they planned on splitting it out to a separate package so we didn't add it in GalacticOptim yet.

test/NNPDE_tests.jl
```diff
@@ -19,30 +26,45 @@ bcs = [u(0.,θ) ~ 1.0 , u(1.,θ) ~ 1.202]
...
+res2 = GalacticOptim.solve(prob,NelderMead(); cb = cb, maxiters=10000)
```
Vaibhavdixit02, line 50: Using a gradient method might hit more coverage?

Vaibhavdixit02, line 50:
```suggestion
res2 = GalacticOptim.solve(prob,BFGS(); cb = cb, maxiters=10000)
```

test/NNPDE_tests.jl
```diff
@@ -61,16 +83,23 @@ domains = [x ∈ IntervalDomain(0.0,1.0),
...
 prob = NeuralPDE.discretize(pde_system,discretization)
-alg = NeuralPDE.NNDE(chain,opt,autodiff=false)
-phi,res  = solve(prob,alg,verbose=true, maxiters=600)
+
+# phi,res  = solve(prob,verbose=true, maxiters=600)
+initθ = discretization.initθ
+p = DiffEqBase.NullParameters()
+loss = (θ) -> prob.f(θ,p)
+
+res = DiffEqFlux.sciml_train(loss, initθ, Flux.ADAM(0.1); cb = cb, maxiters=600)
```
Vaibhavdixit02, line 100: Is the choice of `ADAM` necessary @ChrisRackauckas?

test/NNPDE_tests.jl
KirillZubov: ok

test/NNPDE_tests.jl
KirillZubov: @Vaibhavdixit02 @ChrisRackauckas BFGS() optimizer doesn't work correctly with Stochastic Loss. 
https://github.com/SciML/NeuralPDE.jl/blob/84c7982cf7fea1be94967d01b42c28c835002dd2/src/pinns_pde_solve.jl#L389

ChrisRackauckas: That's expected. BFGS assumes the loss function is deterministic, so you need to use an optimizer robust to stochastic loss functions for that.

test/NNPDE_tests.jl
ChrisRackauckas: oh they weren't added? Just do it like how it was done in sciml_train. It's very straightforward.

test/NNPDE_tests.jl
KirillZubov: ok, thanks. ADAM is one of them but not supported on GalacticOptim yet.

ChrisRackauckas: we need that ASAP @Vaibhavdixit02

src/pinns_pde_solve.jl
```diff
@@ -1,5 +1,68 @@
...
+    end
+
+    PhysicsInformedNN(dx,chain,initθ,phi,autodiff,derivative, training_strategies, kwargs)
+end
+
+struct TrainingStrategies
+    stochastic_loss::Bool
+end
+
+function TrainingStrategies(;stochastic_loss=true)
```
ChrisRackauckas, line 64: instead of true false, make it `strategy = GridTraining()`, or `strategy = StochasticTraining()`, I think we have a list of 5 we want to do now, and that will only grow, so booleans won't manage it well and instead of a dispatch-based way to writing each loss function will be needed.

ChrisRackauckas:
Update GalacticOptim and it should be able to use ADAM to test the stochastic loss.

README.md
```diff
@@ -48,18 +48,26 @@ domains = [x ∈ IntervalDomain(0.0,1.0),
...
 dx = 0.1
-discretization = PhysicsInformedNN(dx)
 
-# Neural network and optimizer
-opt = Flux.ADAM(0.02)
+# Neural network
 dim = 2 # number of dimensions
 chain = FastChain(FastDense(dim,16,Flux.σ),FastDense(16,16,Flux.σ),FastDense(16,1))
 
+training_strategies= TrainingStrategies(stochastic_loss=false)
```
ChrisRackauckas, line 56: this needs to be updated

Vaibhavdixit02 (approved):

README.md
KirillZubov: yeap, thanks

ChrisRackauckas:
cheers, this is looking really good.

PR #135, opened 2020-08-15
Stochastic Loss Function PINNS

KirillZubov:
-added Stochastic Loss Function
-removed `NNODE`  and `solver` overload
-refactoring`discretize` and `PhysicsInformedNN`
-use  GalacticOptim solver

### DIFF
--- a/src/NeuralPDE.jl
+++ b/src/NeuralPDE.jl
@@ -7,6 +7,7 @@ using Flux, Zygote, DiffEqSensitivity, ForwardDiff, Random, Distributions
 using DiffEqFlux, Adapt, DiffEqNoiseProcess, CUDA, StochasticDiffEq
 using ModelingToolkit
 using GalacticOptim
+using Zygote: @nograd
 import Tracker, Optim
 
 abstract type NeuralPDEAlgorithm <: DiffEqBase.AbstractODEAlgorithm end
--- a/src/NeuralPDE.jl
+++ b/src/NeuralPDE.jl
@@ -7,7 +7,6 @@ using Flux, Zygote, DiffEqSensitivity, ForwardDiff, Random, Distributions
 using DiffEqFlux, Adapt, DiffEqNoiseProcess, CUDA, StochasticDiffEq
 using ModelingToolkit
 using GalacticOptim
-using Zygote: @nograd
 import Tracker, Optim
 
 abstract type NeuralPDEAlgorithm <: DiffEqBase.AbstractODEAlgorithm end
@@ -130,7 +129,8 @@ include("pinns_pde_solve.jl")
 
 export NNDE, TerminalPDEProblem, NNPDEHan, NNPDENS, NNRODE,
        KolmogorovPDEProblem, NNKolmogorov, NNStopping,
-       PhysicsInformedNN, discretize,
-       build_loss_function, get_loss_function ,generate_training_sets
+       PhysicsInformedNN,TrainingStrategies, discretize,
+       build_loss_function, get_loss_function ,generate_training_sets,
+       get_phi, get_derivative
 
 end # module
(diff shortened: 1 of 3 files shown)


src/pinns_pde_solve.jl
```diff
@@ -381,7 +398,9 @@ function DiffEqBase.solve(
     args...;
     abstol = 1f-6,
     verbose = false,
-    maxiters = 100)
+    maxiters = 100,
+    stochastic_loss = true,
```
ChrisRackauckas, line 402: why is this not part of the discretization? The setup here is still really odd and should get fixed

src/pinns_pde_solve.jl
```diff
@@ -266,11 +266,27 @@ function get_loss_function(loss_function, train_set)
...
 
-    function loss(θ, phi, derivative)
-        if (loss_function isa Array)
-            1.0f0/τ *sum(sum(abs2,inner_loss(l,phi,x,θ,derivative) for x in set) for (l,set) in zip(loss_function,train_set))
+    if !(loss_function isa Array)
+        loss_function = [loss_function]
+        train_set = [train_set]
+    end
+
+    function loss(θ, phi, derivative, stochastic_loss)
```
ChrisRackauckas, line 274: should probably make this dispatching already , since we will want to have a whole set of training strategies.

src/NeuralPDE.jl
```diff
@@ -7,6 +7,7 @@ using Flux, Zygote, DiffEqSensitivity, ForwardDiff, Random, Distributions
 using DiffEqFlux, Adapt, DiffEqNoiseProcess, CUDA, StochasticDiffEq
 using ModelingToolkit
 using GalacticOptim
+using Zygote: @nograd
```
ChrisRackauckas, line 10: this isn't used?

test/NNPDE_tests.jl
```diff
@@ -247,20 +288,37 @@ dim = length(domains)
...
 
-function loss_function(θ, phi, derivative)
-    return pde_loss_function(θ, phi, derivative) + bc_loss_function(θ, phi, derivative)
+function loss_function(θ)
+    return pde_loss_function(θ) + bc_loss_function(θ)
 end
 
-prob = GalacticOptim.OptimizationProblem(loss_function, zeros(dim))
+initθ = discretization.initθ
+res = DiffEqFlux.sciml_train(loss_function, initθ, Flux.ADAM(0.1); cb = cb, maxiters=1000)
```
ChrisRackauckas, line 318: this just to test sciml_train? Note that will be deprecated for GalacticOptim once it's ready

test/NNPDE_tests.jl
```diff
@@ -247,20 +288,37 @@ dim = length(domains)
...
 
-phi,res  = solve(prob,alg,verbose=true, maxiters=1000)
+# prob = GalacticOptim.OptimizationProblem(loss_function,initθ)
```
ChrisRackauckas, line 320: You might need to set `AutoZygote()` for this.

src/pinns_pde_solve.jl
```diff
@@ -1,5 +1,55 @@
...
+  kwargs::K
+end
+
+function PhysicsInformedNN(dx,
+                           chain,
+                           init_params = nothing;
+                           _phi = nothing,
+                           autodiff=false,
+                           _derivative = nothing,
+                           training_strategies = TrainingStrategies(),
```
ChrisRackauckas, line 18: I think you just want this to just be like `training_strategy = StochasticSampling()` and stuff like that, so you can have training-mode-specific caches and arguments.

ChrisRackauckas: I think that can be handled in a next PR though. This has a lot in it, looks great, so I'd like to merge and make the next diff clearer.

PR #137, opened 2020-08-28
Support gpu

KirillZubov:


ChrisRackauckas:
You need to take that gpu Project.toml as well.

KirillZubov:
but `CUDA`  is already in Project.toml. Or what do you mean?

ChrisRackauckas:
oh then you don't need the gpu environment.

.gitlab-ci.yml
```diff
@@ -1,5 +1,9 @@
 include: https://raw.githubusercontent.com/JuliaGPU/gitlab-ci/master/templates/v6.yml
 
+variables:
+  JULIA_CUDA_VERBOSE: 'true'
```
ChrisRackauckas, line 4: probably don't want that.

docs/src/solvers/pinn.md
```diff
@@ -70,3 +70,7 @@ the subset of points from a full training set.
 Besides the high level of API: `discretize(pde_system, discretization)`, we can also use low-level API methods:  `build_loss_function`, `get_loss_function` ,`generate_training_sets`,`get_phi`, `get_derivative`.
 
 See how this can be used in docs examples or take a look at the tests.
+
+### GPUs
+
+Actually, it is just enough add `|>gpu` after `chain` and all will works.
```
ChrisRackauckas, line 76: More formal. Describe it like, if the Chain is a GPU compatible chain then it's GPU accelerated. For Flux chains then this is done by doing ... For more information on using GPUs with Flux, consult ...

KirillZubov: ok

ChrisRackauckas:
looks like CI is failing?

KirillZubov:
yeap, but isn't depend on GPU. It depends on sciml_train. We need add `progress = false` to `DiffEqFlux.sciml_train(...)` with ADAM opt.

ChrisRackauckas:
how are there so many failures though? What happened?

KirillZubov:
it need to add `progress = false` to  DiffEqFlux.sciml_train(loss, initθ, opt, progress = false; ) when we use `ADAM()`. It should solve the problem for `NNODE` and `NNRODE`.

PR #145, opened 2020-09-08
Adds missing DiffEqFlux import in first example

mdmurbach:
Awesome package, thank you! Just noticed there's a missing `using DiffEqFlux` when directly running the first example in the README.md


ChrisRackauckas:
Thanks!

PR #157, opened 2020-10-02
Improve performance of derivative

KirillZubov:
improve performance of derivative

### DIFF
--- a/src/pinns_pde_solve.jl
+++ b/src/pinns_pde_solve.jl
@@ -39,17 +39,18 @@ function PhysicsInformedNN(dx,
     else
         initθ = init_params
     end
+
     isuinplace = chain.layers[end].out == 1
     dim = chain.layers[1].in
 
     if _phi == nothing
-        phi = get_phi(chain,isuinplace)
+        phi = get_phi(chain)
     else
         phi = _phi
     end
 
     if _derivative == nothing
-        derivative = get_derivative(dim,phi,autodiff,isuinplace)
+        derivative = get_derivative(autodiff,isuinplace)
     else
         derivative = _derivative
     end
@@ -87,11 +88,18 @@ Dict{Symbol,Int64} with 3 entries:
 get_dict_vars(vars) = Dict( [Symbol(v) .=> i for (i,v) in enumerate(vars)])
 
 # Wrapper for _transform_derivative
-function transform_derivative(ex,indvars,depvars,dict_indvars,dict_depvars)
+function transform_derivative(ex,dict_indvars,dict_depvars)
     if ex isa Expr
-        ex.args = _transform_derivative(ex.args,indvars,depvars,dict_indvars,dict_depvars)
+        ex.args = _transform_derivative(ex.args,dict_indvars,dict_depvars)
     end
     return ex
+end
+
+function get_ε(dim, der_num)
+    epsilon = cbrt(eps(Float32))
+    ε = zeros(Float32, dim)
+    ε[der_num] = epsilon
+    ε
 end
 
 """
@@ -101,20 +109,23 @@ Transform the derivative expression to inner representation
 
 1. First derivative of function 'u(x,y)' with respect to x
 
-Take expressions in the form: `derivative(u(x,y,θ), x)` to `derivative(unn, x, y, undv, order, θ)`,
+Take expressions in the form: `derivative(u(x,y,θ), x)` to `derivative(u,unn, [x, y], εs, order, θ)`,
 
 where
+ u - function
  unn - unique number for the function
  x,y - coordinates of point
- undv - uniqie number for derivative variables
+ εs - epsilon mask
  order - order of derivative
  θ - parameters of neural network
 """
-function _transform_derivative(_args,indvars,depvars,dict_indvars,dict_depvars)
+function _transform_derivative(_args,dict_indvars,dict_depvars)
+    dim = length(dict_indvars)
+    εs = [get_ε(dim,d) for d in 1:dim]
     for (i,e) in enumerate(_args)
         if !(e isa Expr)
             if e == :derivative && _args[end] != :θ
-                derivative_variables = []
+                derivative_variables = Symbol[]
                 order = 0
                 while (_args[1] == :derivative)
                     order += 1
@@ -125,11 +136,12 @@ function _transform_derivative(_args,indvars,depvars,dict_indvars,dict_depvars)
                 indvars = _args[2:end-1]
                 depvars_num = dict_depvars[depvar]
                 undv = [dict_indvars[d_p] for d_p  in derivative_variables]
-                _args = [:derivative,depvars_num, indvars...,undv, order, :θ]
+                εs_dnv = [εs[d] for d in undv]
+                _args = [:derivative,  :phi, depvars_num, :([$(indvars...)]), εs_dnv, order, :θ]
                 break
             end
         else
-            _args[i].args = _transform_derivative(_args[i].args,indvars,depvars,dict_indvars,dict_depvars)
+            _args[i].args = _transform_derivative(_args[i].args,dict_indvars,dict_depvars)
         end
     end
     return _args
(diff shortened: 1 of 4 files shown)


src/pinns_pde_solve.jl
```diff
@@ -294,25 +294,24 @@ function generate_training_sets(domains,dx,bcs,indvars,depvars,dict_indvars,dict
...
+        push!(train_set, [points...])
     end
 
-    train_bound_set =  Array{Array{Float32,1}}[]
+    train_bound_set = []
     for bt in bound_args
-        _set = Array{Float32}[]
+        _set = []
         span = [get(dict_var_span, b, b) for b in bt]
         for points in Iterators.product(span...)
```
ChrisRackauckas: don't splat arrays.

src/pinns_pde_solve.jl
```diff
@@ -294,25 +294,24 @@ function generate_training_sets(domains,dx,bcs,indvars,depvars,dict_indvars,dict
...
-            push!(_set, Float32[points...])
+            push!(_set, [points...])
```
ChrisRackauckas: map instead of push!

src/pinns_pde_solve.jl
```diff
@@ -282,80 +293,49 @@ function generate_training_sets(domains,dx,bcs,indvars,depvars,dict_indvars,dict
     spans = [d.domain.lower:dx:d.domain.upper for (d,dx) in zip(domains,dxs)]
     dict_var_span = Dict([Symbol(d.variables) => d.domain.lower:dx:d.domain.upper for (d,dx) in zip(domains,dxs)])
 
-    train_set = []
-    for points in Iterators.product(spans...)
-        push!(train_set, [points...])
-    end
+    train_set = map(points -> [points...], Iterators.product(spans...))
```
ChrisRackauckas, line 296: why splat?

src/pinns_pde_solve.jl
```diff
@@ -282,80 +293,49 @@ function generate_training_sets(domains,dx,bcs,indvars,depvars,dict_indvars,dict
...
-            θ = _x[end]
-            εs_dnv = [εs[d] for d in dnv]
-            return _derivative(x,θ,order,εs_dnv,var_num)
-        end
-        _derivative = (x,θ,order,εs_dnv,var_num) ->
-        begin
-            ε = εs_dnv[order]
+            ε = εs[order]
             if order == 1
                 if isuinplace
```
ChrisRackauckas, line 331: can u ever be in place if we are reverse mode? No I think.

src/pinns_pde_solve.jl
```diff
@@ -282,80 +293,49 @@ function generate_training_sets(domains,dx,bcs,indvars,depvars,dict_indvars,dict
...
-                    return (phi(x+ε,θ) - phi(x-ε,θ))/(2*epsilon)
+                    return (u(x+ε,θ) - u(x-ε,θ))/(2*epsilon)
                 else
-                    return (phi(x+ε,θ)[var_num] - phi(x-ε,θ)[var_num])/(2*epsilon)
+                    return (u(x+ε,θ)[var_num] - u(x-ε,θ)[var_num])/(2*epsilon)
```
ChrisRackauckas, line 334: you're going to repeat derivative calls. The computation should be re-ordered next PR.

src/pinns_pde_solve.jl
```diff
@@ -282,80 +293,49 @@ function generate_training_sets(domains,dx,bcs,indvars,depvars,dict_indvars,dict
...
-    flat_train_bound_set = collect(Iterators.flatten(train_bound_set))
+    flat_train_bound_set = map(x -> x, Iterators.flatten(train_bound_set))
     train_domain_set =  setdiff(train_set, flat_train_bound_set)
 
     [train_domain_set,train_bound_set,train_set]
 end
 
-function get_phi(chain,isuinplace)
+function get_phi(chain)
+    chain isa Chain && "Chain deprecated, use FastChain"
```
ChrisRackauckas, line 310: no, we need to support Chain.

src/pinns_pde_solve.jl
```diff
@@ -282,80 +293,49 @@ function generate_training_sets(domains,dx,bcs,indvars,depvars,dict_indvars,dict
...
-            phi = (x,θ) -> first(chain(adapt(typeof(θ),x),θ))
-        else
-            phi = (x,θ) -> chain(adapt(typeof(θ),x),θ)
-        end
+    if isuinplace
+        phi = (x,θ) -> first(chain(adapt(typeof(θ),x),θ))
     else
-        _,re  = Flux.destructure(chain)
-        if isuinplace
-            phi = (x,θ) -> first(re(θ)(adapt(typeof(θ),x)))
```
ChrisRackauckas, line 317: do you need to destructure here?

src/pinns_pde_solve.jl
KirillZubov: ok

src/pinns_pde_solve.jl
KirillZubov: no

src/pinns_pde_solve.jl
KirillZubov: yes, "in place" is not needed at all

src/pinns_pde_solve.jl
KirillZubov: replaced with `collect(points)` instead of splat.

src/pinns_pde_solve.jl
ChrisRackauckas: oh actually I think you do if you're going to use sciml_train over Flux Chain here.

src/pinns_pde_solve.jl
ChrisRackauckas: what is `points` here: a tuple?

KirillZubov: yes

KirillZubov: but actually, splat is faster

src/pinns_pde_solve.jl
KirillZubov: I confused with the Flux.destructor from PhysicsInformedNN

KirillZubov: yeah, it needs here

src/pinns_pde_solve.jl
```diff
@@ -31,11 +31,7 @@ function PhysicsInformedNN(dx,
                            strategy = GridTraining(),
                            kwargs...)
     if init_params === nothing
-        if chain isa FastChain
-            initθ = DiffEqFlux.initial_params(chain)
-        else
-            initθ,re  = Flux.destructure(chain)
-        end
+        initθ = DiffEqFlux.initial_params(chain)
```
ChrisRackauckas, line 34: is there a test that says this is okay with Flux.Chain? I would think no?

src/pinns_pde_solve.jl
ChrisRackauckas: Really? Show how: the two should be the same if it's a small tuple. But the bigger thing is, why are tuples generated? It should probably go straight to arrays.

src/pinns_pde_solve.jl
KirillZubov: fastchain = FastChain(FastDense(2,16,Flux.σ),FastDense(16,16,Flux.σ),FastDense(16,1))
chain = Chain(Dense(2,16,Flux.σ),Dense(16,16,Flux.σ),Dense(16,1))

initθ1 = DiffEqFlux.initial_params(chain)
initθ2,_  = Flux.destructure(chain)

initθ1 == initθ2

KirillZubov: it is the wrapper : `initial_params(f::Chain) = Flux.destructure(f)[1]` 
https://github.com/SciML/DiffEqFlux.jl/blob/b642d5c917b33e2593165a2e1b076f4080eb6a9b/src/fast_layers.jl#L5

src/pinns_pde_solve.jl
KirillZubov: yes, we should go straight to arrays it would be the best but `Iterators.product(spans...)` create a tuple.

KirillZubov:
```julia
    spans = [0.0:0.25:2.0, 0.0:0.25:2.0, 0.0:0.25:2.0]
    julia> @benchmark map(points -> collect(points), Iterators.product(spans...))
    BenchmarkTools.Trial:
      memory estimate:  86.42 KiB
      allocs estimate:  737
      --------------
      minimum time:     51.168 μs (0.00% GC)
      median time:      59.035 μs (0.00% GC)
      mean time:        88.805 μs (21.89% GC)
      maximum time:     68.412 ms (99.81% GC)
      --------------
      samples:          10000
      evals/sample:     1

    julia> @benchmark map(points -> [points...], Iterators.product(spans...))
    BenchmarkTools.Trial:
      memory estimate:  86.42 KiB
      allocs estimate:  737
      --------------
      minimum time:     48.730 μs (0.00% GC)
      median time:      66.870 μs (0.00% GC)
      mean time:        158.667 μs (13.34% GC)
      maximum time:     76.640 ms (99.82% GC)
      --------------
      samples:          10000
      evals/sample:     1
    ```
    uh,  `collect` is a little faster.

ChrisRackauckas:
Is there a test in the test set that solves like Poisson with a Flux Chain? It would be good to test it through the whole system.

PR #162, opened 2020-10-04
avoid generators to allow Zygote to handle the sum

ChrisRackauckas:


ChrisRackauckas:
@DhairyaLGandhi this seems... really bad. Are iterators in Zygote overloaded to only work with Float32?

DhairyaLGandhi:
No overloading for specific eltypes. Can you see if forwarding generator to broadcasting would be sufficient in this case?
```julia
 @adjoint function Base.Generator(f,iter)
          y, back = Flux.Zygote.pullback(x -> f.(x), iter)
          y, Δ -> begin
            (nothing, back(Δ)...)
          end
        end
```
This is just for if you need to handle generators; a more performant result might need some benchmarking

ChrisRackauckas:
That gives zero gradients.

DhairyaLGandhi:
Is it interacting with custom adjoints that could be misbehaving? 

Using this adjoint got the generator syntax working without much issue
```julia
julia> Flux.Zygote.gradient(x -> sum(i^2 for i in x), 1:3)
([2, 4, 6],)
```

ChrisRackauckas:
Which custom adjoint?

DhairyaLGandhi:
Probably something in `inner_loss`, because once we get the `Base.Generator` forwarded, there's not much else to that's happening in the adjoint. I also tried

```Julia
julia> Flux.Zygote.gradient(x -> sum(abs2, i^2 for i in x), 1:3)
([4, 32, 108],)
```

DhairyaLGandhi:
It can't handle closures cleanly I suppose, but that should return `nothing` instead of zeros, hence I thought its a custom adjoint being invoked.

ChrisRackauckas:
There is one in FastDense but it's fairly basic and just does linear algebra.

PR #174, opened 2020-11-12
Remove θ

KirillZubov:
Removed θ(the neural network parameters ) from the high-level PDE description.


KirillZubov:
Theta has remained reserved and can't be used as a parameter of the equation. I wrote a warning to docs.

ChrisRackauckas:
just use a gensym?

KirillZubov:
> just use a gensym?

oh cool, I didn't know about it

ChrisRackauckas (approved):
anything else needed in this? LGTM

PR #175, opened 2020-11-15
Update Documenter version

mkg33:


ChrisRackauckas:
The lower bound doesn't matter all that much

PR #183, opened 2020-11-29
[WIP] GPU pinns

KirillZubov:


test/NNPDE_tests_gpu.jl
```diff
@@ -43,12 +43,15 @@ dx = 0.25; dy= 0.25; dt = 0.25
 # Neural network
 chain = FastChain(FastDense(3,16,Flux.σ),FastDense(16,16,Flux.σ),FastDense(16,1)) |>gpu
 
+initθ = initial_params(chain) |>gpu
+strategy = QuadratureTraining(algorithm=CubaCuhre(),reltol= 1e-5,abstol= 1e-5,maxiters=50)
```
ChrisRackauckas, line 47: I'm not sure if the quadrature algorithms will sample on GPUs. I don't think they can.

test/NNPDE_tests_gpu.jl
```diff
@@ -43,12 +43,14 @@ dx = 0.25; dy= 0.25; dt = 0.25
 # Neural network
 chain = FastChain(FastDense(3,16,Flux.σ),FastDense(16,16,Flux.σ),FastDense(16,1))
```
ChrisRackauckas, line 44:
```suggestion
const gpuones = gpu(ones(1))
chain = FastChain(FastDense(3,16,Flux.σ),FastDense(16,16,Flux.σ),FastDense(16,1),(u,p)->gpuones .* u)
```

ChrisRackauckas: This is a hacky way to "prove" it's running on the GPU, because multiplying by an array of ones on the GPU would fail if the `u` is on the CPU.

test/NNPDE_tests_gpu.jl
```diff
@@ -87,7 +89,9 @@ domains = [x ∈ IntervalDomain(-2.2,2.2)]
 # Neural network
 chain = FastChain(FastDense(1,12,Flux.σ),FastDense(12,12,Flux.σ),FastDense(12,1))
```
ChrisRackauckas, line 90:
```suggestion
chain = FastChain(FastDense(1,12,Flux.σ),FastDense(12,12,Flux.σ),FastDense(12,1),(u,p)-> gpuones .* u)
```

PR #189, opened 2020-12-05
Quasi random training

KirillZubov:


src/pinns_pde_solve.jl
```diff
@@ -660,9 +657,16 @@ function DiffEqBase.discretize(pde_system::PDESystem, discretization::PhysicsInf
...
-        (pde_loss_function, bc_loss_function, automatic_differentiation)
+
+        (pde_loss_function, bc_loss_function)
+    end
+
+    automatic_differentiation =
+    if strategy isa GridTraining || strategy isa StochasticTraining || strategy isa QuadratureTraining
+        GalacticOptim.AutoZygote()
+    elseif strategy isa SobolTraining
+        GalacticOptim.AutoForwardDiff()
```
ChrisRackauckas, line 669: wait why? The training method shouldn't impact that at all?

ChrisRackauckas: You're sampling outside of the loss function, right?

KirillZubov: it is some problem with Zygote.gradient

KirillZubov:
```julia
using Sobol, Flux, DiffEqFlux, GalacticOptim
lb = [0.0,1.0]
ub = [1.0,2.0]
s =SobolSeq(lb, ub)

chain = FastChain(FastDense(1,12,Flux.σ),FastDense(12,1))
initθ = DiffEqFlux.initial_params(chain)

function f_(θ,p)
	r_point = next!(s)
	sum(θ) + sum(r_point)
end

f = OptimizationFunction(f_, GalacticOptim.AutoZygote())
prob = GalacticOptim.OptimizationProblem(f, initθ)

_loss = function(θ)
      x = prob.f.f(θ, prob.p)
end
_loss(initθ)

ps = Flux.params(initθ)

julia> gs = Flux.Zygote.gradient(ps) do
                         x = _loss(initθ)
                         first(x)
                       end
ERROR: Mutating arrays is not supported
```

ChrisRackauckas: Then you're doing something wrong. What does the generated loss function look like? There should be no mutating arrays: you're just slicing from pre-generated arrays.

ChrisRackauckas: I assume what you're doing incorrectly is calling `sample` inside of the loss function. That will fail exactly like this, but it's unnecessary to do that work in the loss function and try to differentiate it since those are just the evaluation points. Basically think of them as grid points and, like grid points, compute them before the loss function, store them in a constant vector, and index from that vector when you need to ask where to evaluate the PINN. That has no mutation.

Issue #184, opened 2020-11-30
Method for calculating the full Kolmogorov PDE solution


ChrisRackauckas:
https://arxiv.org/pdf/2011.04602.pdf

PR #202, opened 2020-12-15
Full Kolmogorov PDE solution solver

ashutosh-b-b:
Started working on #184 .


ashutosh-b-b:
@ChrisRackauckas is [this](https://github.com/SciML/NeuralPDE.jl/blob/fdf4f2ee6062c85017efe35d31a703318931b527/src/NeuralPDE.jl#L94) a correct way of taking input for different domains?

test/runtests.jl
```diff
@@ -31,7 +31,9 @@ const is_CI = haskey(ENV,"CI")
   if GROUP == "All" || GROUP == "NNRODE"
         @time @safetestset "NNRODE" begin include("NNRODE_tests.jl") end
   end
-
+  if GROUP == "All" || GROUP == "NNPKOLMOGOROV"
+        @time @safetestset "NNParamKolmogorov" begin include("NNParamKolmogorov_tests.jl") end
```
ChrisRackauckas, line 35: How long does this test set take? I don't think it needs to be separated from NNRODE. But if it does, it needs to be added to the Github CI action.

ChrisRackauckas:
Looks good but needs documentation.

test/runtests.jl
ashutosh-b-b: Just in the order of minutes I guess. Yeah we can add it with NNRODE. I ll add this one and check,

ChrisRackauckas:
Still needs docs. Please follow up.

PR #210, opened 2020-12-23
fixed some problems in 2D poisson equation

TeAmP0is0N:
The previous code was causing problems when i was recompiling it in my system. So I am making some changes in the docs and I will complete all the changes by tomorrow evening


TeAmP0is0N:
And also newbie could have difficulty to compile so I thought these changes should be made

PR #225, opened 2021-01-07
QuasiRandom and Quadratute training: argument issues

KirillZubov:


docs/src/solvers/pinn.md
```diff
@@ -55,13 +55,19 @@ List of training strategies that are available now:
...
-  sequences. See the [QuasiMonteCarlo.jl](https://github.com/SciML/QuasiMonteCarlo.jl) for the full
+- `QuasiRandomTraining(sampling_alg,points,minibatch)`: The training set is generated on quasi-random low discrepency
+  sequences.
+  `sampling_alg`is the quasi-Monte Carlo sampling algorithm,
+  `points` is number of quasi-random points in training set,
+  `minibatch` is number of generated training samples.
+  See the [QuasiMonteCarlo.jl](https://github.com/SciML/QuasiMonteCarlo.jl) for the full
   set of quasi-random sampling algorithms which are available.
-- `QuadratureTraining(quadrature_alg)`: The loss is computed as an approximation of the integral of the PDE loss 
+- `QuadratureTraining(quadrature_alg,reltol,abstol,maxiters,batch)`: The loss is computed as an approximation of the integral of the PDE loss
```
ChrisRackauckas, line 65: it's probably reasonable to do reltol and abstol as kwargs, because `abstol=1e-6,reltol=1e-3` should essentially be as accurate as Float32 accuracy can get. Maxiters could probably have an absurd default like `Int(1e5)`, mostly to check for divergence. It's usually adaptive algorithms that get defaults like this because those defaults are probably good for almost every case.

docs/src/solvers/pinn.md
```diff
@@ -55,13 +55,19 @@ List of training strategies that are available now:
    to the spacing per direction.
  - `StochasticTraining(points)`: `points` number of sochastically sampled points from the domain. 
    In each optimization iteration, we randomly select a new subset of points from a full training set.
-- `QuasiRandomTraining(sampling_alg)`: The training set is generated on quasi-random low discrepency
-  sequences. See the [QuasiMonteCarlo.jl](https://github.com/SciML/QuasiMonteCarlo.jl) for the full
+- `QuasiRandomTraining(sampling_alg,points,minibatch)`: The training set is generated on quasi-random low discrepency
```
ChrisRackauckas, line 58: I think minibatch should be a keyword argument. I actually don't quite get what it does from the description here. Is it the number of points per loss evaluation, where `points` is the total points from the quasi-random set, so the minibatch is the number of points to take in a subsample?

KirillZubov: `minibatch` is the number of subsets, where `points` is the number of points in minibatch. The number of the total points is `length(lb) * points * minibatch`, where `lb` is the lower bound and `length(lb)` is the dimensionality. 
On each iteration of training, it is randomly selected one of the minibatch.

```julia
using QuasiMonteCarlo
points = 3
lb = [0, 0]
ub = [1, 1]
sampling_alg = UniformSample()
minibatch = 4
s = QuasiMonteCarlo.generate_design_matrices(points,lb,ub,sampling_alg,minibatch)

julia> size(s)
(4,)
julia> size(s[1])
(2, 3)

```

ChrisRackauckas: >minibatch is the number of subsets, where points is the number of points in minibatch. The number of the total points is length(lb) * points * minibatch, where lb is the lower bound and length(lb) is the dimensionality.
On each iteration of training, it is randomly selected one of the minibatch.

That's perfect. That should go in the docs.

docs/src/solvers/pinn.md
```diff
@@ -55,19 +55,22 @@ List of training strategies that are available now:
...
   at each iteration using [adaptive quadrature methods](https://en.wikipedia.org/wiki/Adaptive_quadrature)
   via the differentiable [Quadrature.jl](https://github.com/SciML/Quadrature.jl).
-  `quadrature_alg` is quadrature algorithm, `reltol` is relative tolerance, `abstol` absolute tolerance,
-  `maxiters`: the maximum number of iterations in quadrature algorithm, `batch`: the preferred number of points to batch.
-  See the Quadrature.jl documentation for the choices of quadrature methods.
+  - `quadrature_alg` is quadrature algorithm,
+  - `reltol`: relative tolerance,
+  - `abstol`: absolute tolerance,
+  - `maxiters`: the maximum number of iterations in quadrature algorithm,
+  - `batch`: the preferred number of points to batch.
```
ChrisRackauckas, line 71: if 0 then?

docs/src/solvers/pinn.md
```diff
@@ -55,19 +55,22 @@ List of training strategies that are available now:
...
-  See the [QuasiMonteCarlo.jl](https://github.com/SciML/QuasiMonteCarlo.jl) for the full
-  set of quasi-random sampling algorithms which are available.
-- `QuadratureTraining(quadrature_alg,reltol,abstol,maxiters,batch)`: The loss is computed as an approximation of the integral of the PDE loss
+ - `QuasiRandomTraining(points;sampling_alg,minibatch)`: The training set is generated on quasi-random low discrepency sequences.`minibatch` is the number of subsets, where `points` is the number of quasi-random points in minibatch.
+  The number of the total points is `length(lb) * points * minibatch`,
+  where `lb` is the lower bound and `length(lb)` is the dimensionality.
+  `sampling_alg` is the quasi-Monte Carlo sampling algorithm.
+  On each iteration of training, it is randomly selected one of the minibatch.
+  See the [QuasiMonteCarlo.jl](https://github.com/SciML/QuasiMonteCarlo.jl) for the full set of quasi-random sampling algorithms which are available.
+- `QuadratureTraining(;quadrature_alg,reltol,abstol,maxiters,batch)`: The loss is computed as an approximation of the integral of the PDE loss
```
ChrisRackauckas, line 64: show the defaults

docs/src/solvers/pinn.md
KirillZubov: if batch = 0, the number of points in the batch is determined adaptively by the algorithm.

PR #228, opened 2021-01-08
fix GPU tests

KirillZubov:


ChrisRackauckas:
Seems like it's getting stuck?

KirillZubov:
> Seems like it's getting stuck?
it's strange. I ran this example on google collab and the training process doesn't get stuck.

ChrisRackauckas:
@pkofod BFGS isn't compatible with GPUs?

KirillZubov:
The PR can be merged. GPU tests work but over runtime in buildkite.

ChrisRackauckas:
I upped the allowed time on buildkite

KirillZubov:
Two hours will still not be enough for the current tests. I will try to finish the type conversion task(Array -> CuArray) by the next week, it will speed up the execution time, but it will still not be fast enough. I think, it needs to replace the tests, the first test will run more than one hours anyway, and the second does not converge with ADAM only with BFGS and BFGS does not work with GPU.  Instead, it needs more simple tests that will run less than 1 hour.

PR #229, opened 2021-01-09
update doc

KirillZubov:


docs/src/solvers/pinn.md
```diff
@@ -26,7 +26,7 @@ discretization = PhysicsInformedNN(chain,
                                    nothing; #init_params
                                    phi = nothing,
                                    derivative = nothing,
-                                   strategy = GridTraining())
+                                   strategy = nothing)
```
ChrisRackauckas, line 29: hmm, if strategy is required, shouldn't it be a required argument?

KirillZubov: yeap, you right

src/pinns_pde_solve.jl
```diff
@@ -20,20 +21,20 @@ struct PhysicsInformedNN{C,P,PH,DER,T,K}
 end
 
 function PhysicsInformedNN(chain,
-                           init_params = nothing;
+                           strategy;
+                           _initθ = nothing,
```
ChrisRackauckas, line 25: Keep the old name. We shouldn't require Unicode in the API otherwise some terminals won't be able to use the argument.

KirillZubov: Actually, it already was `initθ` in arguments of PhysicsInformedNN. Ok, I will rename all the `initθ` to `init_params`.

ChrisRackauckas: Yeah internally is fine, but we have a policy to not expose Unicode externally.

KirillZubov: ok

ChrisRackauckas:
Hmm tests are failing?

PR #233, opened 2021-01-14
[WIP] update documentation

KirillZubov:


docs/src/pinns/training_strategy.md
```diff
@@ -0,0 +1,27 @@
+
+### Training strategy
```
ChrisRackauckas, line 2: The manual should stay on one page. I don't see a reason to split this.

docs/src/pinns/pinns_solver.md
```diff
@@ -0,0 +1,47 @@
+# Physics-Informed Neural Networks
```
ChrisRackauckas, line 1: this should be in the manual section.

docs/src/pinns/low_level_api.md
```diff
@@ -0,0 +1,53 @@
+### Low-level API
```
ChrisRackauckas, line 1: this should be part of the manual section page on PINNs

docs/make.jl
```diff
@@ -10,23 +10,61 @@ makedocs(
...
+
+            "solvers/kolmogorovbackwards_solver.md",
             "examples/kolmogorovbackwards.md",
+
+            "solvers/optimal_stopping.md"
             "examples/optimal_stopping_american.md",
-            "examples/nnrode_example.md",
         ],
-        "Neural-Enhanced PDE Solvers" => Any[
+        "Neural Network ODE Solver Tutorials" => Any[
```
ChrisRackauckas, line 59: API Documentation

ChrisRackauckas:
I did a few style changes. Nothing really major, things like `##` -> `#` to get the titles right and such. I think this is good to merge. Of course it can keep improving, but I quite like it now.

PR #238, opened 2021-01-19
MTK 5

YingboMa:


KirillZubov:
@YingboMa  could you please explain why:
```julia
@parameters t
@variables u(..)
# @derivatives Dt'~t
Dt = Differential(t)^2
ModelingToolkit.toexpr(Dt(u(t)))
```
 return 
```julia
:(_derivative(u(t), t))
```
I thought maybe it's a typo or why an underscore `_` before `derivative`: `_derivative`

YingboMa:
Since `derivative` is already a function in MTK, I renamed the generated code to be `_derivative` to avoid collision.

Issue #239, opened 2021-01-20
SBINNs: Systems Biology INN -- PINNs for ODEs


finmod:
This extension of PINNs to ODE systems for data driven solutions belongs here. For data driven discovery, it would belong to DataDrivenDiffEq. [https://journals.plos.org/ploscompbiol/article?id=10.1371/journal.pcbi.1007575](url)

ChrisRackauckas:
Yup, PINNs on ODEs can already be done. Actually, `NNODE` could do this from 2017! But... it's super inefficient to do this. We can create a tutorial around it, but this is definitely a case where the user should be pointed to DiffEqFlux.jl instead.

Issue #148, opened 2020-09-21
Parameters identification


LucasPa:
Hi,

First of all congratulations for developing the NeuralPDE package!  I have been playing around the different examples and tutorials and although I am new at Julia, I have been able to solve some PDEs that interest me. 

I would like now to do parameters identifications. For instance using PINNs to identify the value of an unknown coefficient in a PDE from measured data. Is it possible to solve this kind of problem ?

Best,

Lucas

ChrisRackauckas:
For now you'll want to use https://diffeqflux.sciml.ai/dev/ directly for that, but this will be something added to the PINNs.

LucasPa:
Thank you Chris for your answer. Then I will have a look at DiffEqFlux and see how I can use that to perform identification as in :

> Raissi, M., P. Perdikaris, and G. E. Karniadakis. "Physics informed deep learning (Part II): Data-driven discovery of nonlinear partial differential equations. arXiv 2017." arXiv preprint arXiv:1711.10566.

ChrisRackauckas:
It won't be the same as though, though it will be an order of magnitude more efficient: PINNs are a very slow way to calculate things. I'd recommend the tutorial https://diffeqflux.sciml.ai/dev/examples/pde_constrained/

LucasPa:
OK I will have a look to this example and see how I can adapt this to my needs.

Thanks a lot !

LucasPa:
Hi Chris,

Thanks for the hint. I have been trough and many other examples but I am not sure that diffeqflux can actually help me. I would like to identify to identify the coefficients of a Robin boundary condition of a two-dimensional PDE (that only involves x- and y-derivatives and t-derivative) from known data. 

Best,

Lucas

ChrisRackauckas:
You'd just track the parameter in the boundary condition. It's no different from any other parameter.

I'm going to reopen this for tracking adding this to the PINNs though, since it's still something we'll want to do for completeness.

PR #256, opened 2021-02-23
Tracking parameters in PINNs

ashutosh-b-b:
Fixes https://github.com/SciML/NeuralPDE.jl/issues/239 and https://github.com/SciML/NeuralPDE.jl/issues/148


src/pinns_pde_solve.jl
```diff
@@ -782,9 +783,9 @@ function DiffEqBase.discretize(pde_system::PDESystem, discretization::PhysicsInf
     end
 
     function loss_function_(θ,p)
```
ChrisRackauckas, line 785: not quite. We want the `p` here of the parameters to be appended to `theta` so that they are optimized simultaneously, not as hyperparameters.

ChrisRackauckas:
This needs tests.

Project.toml
```diff
@@ -20,6 +20,7 @@ IfElse = "615f187c-cbe4-4ef1-ba3b-2fcf58d6d173"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 ModelingToolkit = "961ee093-0014-501f-94e3-6117800e7a78"
 Optim = "429524aa-4258-5aef-a3af-852621145aeb"
+OrdinaryDiffEq = "1dea7af3-3e70-54e6-95c3-0bf5283fa5ed"
```
ChrisRackauckas, line 23: this only needs to be a test dependency

docs/src/pinn/parm_estim.md
```diff
@@ -0,0 +1,67 @@
+# Optimising Parameters of a Lorenz System
+ Consider a Lorenz System ,
+! [lorenzSystem](https://raw.githubusercontent.com/ashutosh-b-b/github-doc-images/master/lorenz_system.png) with Physics-Informed Neural Networks.
```
ChrisRackauckas, line 3: Split line lengths to the standard of the style guide.

docs/src/pinn/parm_estim.md
```diff
@@ -0,0 +1,67 @@
...
+And then the neural networks as,
+```julia
+input_ = length(domains)
+n = 16
+chain1 = FastChain(FastDense(input_,n,Flux.σ),FastDense(n,n,Flux.σ),FastDense(n,n,Flux.σ),FastDense(n,1))
+chain2 = FastChain(FastDense(input_,n,Flux.σ),FastDense(n,n,Flux.σ),FastDense(n,n,Flux.σ),FastDense(n,1))
+chain3 = FastChain(FastDense(input_,n,Flux.σ),FastDense(n,n,Flux.σ),FastDense(n,n,Flux.σ),FastDense(n,1))
+```
+We will add an additional loss term based on the data that we have in order to optimise the parameters.
+Here we simply calculate the solution of the lorenz system with [OrdinaryDiffEq.jl](https://diffeq.sciml.ai/v1.10/tutorials/ode_example.html#In-Place-Updates-1) and pick up some points randomly
```
ChrisRackauckas, line 29: not randomly: based on the adaptivity of the ODE solver. "This is used to introduce non-uniformity to the time series".

docs/src/pinn/parm_estim.md
```diff
@@ -0,0 +1,67 @@
...
+end
+```
+
+```julia
+discretization = NeuralPDE.PhysicsInformedNN([chain1 , chain2, chain3],NeuralPDE.GridTraining(dt), param_estim=true, additional_loss=additional_loss)
+pde_system = PDESystem(eqs,bcs,domains,[t],[x, y, z],[ρ, β], [1.0 ,1.0])
+prob = NeuralPDE.discretize(pde_system,discretization)
+res = GalacticOptim.solve(prob, BFGS(); cb = cb, maxiters=4000)
+p_ = res.minimizer[end-1:end]
+
```
ChrisRackauckas, line 66: the the image to this PR in a comment, and then use the link to it to add  to add the picture to the docs here.

ChrisRackauckas: you should show how close the fit is. Also, show the final parameters.

src/pinns_pde_solve.jl
```diff
@@ -324,6 +328,30 @@ function build_symbolic_loss_function(eqs,indvars,depvars,
...
+    if eq_params != SciMLBase.NullParameters() && param_estim == false
+        params_symbols = Symbol[]
+        expr_params = Expr[]
+        for (i , eq_param) in enumerate(eq_params)
+            push!(expr_params, :(p[$i]))
+            push!(params_symbols, Symbol(:($eq_param)))
+        end
+        params_eq = Expr(:(=), build_expr(:tuple, params_symbols), build_expr(:tuple, expr_params))
+        push!(ex.args,  params_eq)
+    end
```
ChrisRackauckas, line 354: @KirillZubov review. How does this intersect with your current PR?

src/pinns_pde_solve.jl
```diff
@@ -324,6 +328,30 @@ function build_symbolic_loss_function(eqs,indvars,depvars,
...
+
+    if eq_params != SciMLBase.NullParameters() && param_estim == false
+        params_symbols = Symbol[]
+        expr_params = Expr[]
+        for (i , eq_param) in enumerate(eq_params)
+            push!(expr_params, :(p[$i]))
```
ChrisRackauckas, line 349: This might need to use ArrayInterface.allowed_get_index` to handle the GPUs well.

test/NNPDE_tests.jl
```diff
@@ -449,3 +450,55 @@ u_predict  = [first(phi(x,Array(res.minimizer))) for x in xs]
...
+input_ = length(domains)
+n = 16
+chain1 = FastChain(FastDense(input_,n,Flux.σ),FastDense(n,n,Flux.σ),FastDense(n,n,Flux.σ),FastDense(n,1))
+chain2 = FastChain(FastDense(input_,n,Flux.σ),FastDense(n,n,Flux.σ),FastDense(n,n,Flux.σ),FastDense(n,1))
+chain3 = FastChain(FastDense(input_,n,Flux.σ),FastDense(n,n,Flux.σ),FastDense(n,n,Flux.σ),FastDense(n,1))
+discretization = NeuralPDE.PhysicsInformedNN([chain1 , chain2, chain3],NeuralPDE.GridTraining(dt), param_estim=true, additional_loss=additional_loss)
+pde_system = PDESystem(eqs,bcs,domains,[t],[x, y, z],[ρ, β], [1.0 ,1.0])
+prob = NeuralPDE.discretize(pde_system,discretization)
+res = GalacticOptim.solve(prob, BFGS(); cb = cb, maxiters=4000)
+p_ = res.minimizer[end-1:end]
```
ChrisRackauckas, line 501: why not test all 3 parameters?

test/NNPDE_tests.jl
```diff
@@ -449,3 +450,55 @@ u_predict  = [first(phi(x,Array(res.minimizer))) for x in xs]
...
+
+@test abs2(p_[1] - 28.00) < 1.0
```
ChrisRackauckas, line 503: can that bound be tightened easily?

src/pinns_pde_solve.jl
```diff
@@ -782,9 +819,14 @@ function DiffEqBase.discretize(pde_system::PDESystem, discretization::PhysicsInf
     end
 
     function loss_function_(θ,p)
-        return pde_loss_function(θ) + bc_loss_function(θ)
+        if param_estim == false
```
zoemcc, line 822: I can see a lot of cases where the additional_loss function is unrelated to parameter estimation, we should allow the user to pass it in and use it optimization independent of param_estim here.  The if statement should be whether or not `additional_loss isa Nothing`

docs/src/pinn/parm_estim.md
```diff
@@ -0,0 +1,67 @@
...
+sol = solve(prob, Tsit5(), dt=0.1)
+data = []
+data
+indx = rand(1:1:21 , 11)
+for i in indx
+    data = vcat(data , (sol.u[i] , sol.t[i]))
+end
+
+```
+Then we define the additional loss funciton `additional_loss(phi, θ , p)`, the function has three arguments, `phi` the trial solution, `θ` the parameters of neural networks, and `p` optional parameters.
```
zoemcc, line 50: Minor typo: "loss funciton" should be "loss function"

src/pinns_pde_solve.jl
```diff
@@ -324,6 +328,30 @@ function build_symbolic_loss_function(eqs,indvars,depvars,
         vars_phi = Expr(:(=), build_expr(:tuple, phi_nums), build_expr(:tuple, expr_phi))
         push!(ex.args,  vars_phi)
     end
+    #Add an expression for paramater symbols
```
zoemcc, line 331: Minor typo: "paramater" should be "parameter"

ChrisRackauckas: Use the github suggest feature for typos.

```suggestion
    #Add an expression for parameter symbols
```

docs/src/pinn/parm_estim.md
```diff
@@ -0,0 +1,67 @@
...
+
+```julia
+function additional_loss(phi, θ , p)
+	l = length(θ) - 2
+    _loss(u , t; t_ = cu[t]) = sum(abs2, phi[i](t_ , θ[(i*l - l + 1):(i*l)])[1] - (u[i])  for i in 1:1:3)
```
zoemcc, line 55: What is cu[t] here? (and in the test case)

zoemcc: Also, is there a cleaner way to just evaluate the test solution and subtract the data from it than this?

docs/src/pinn/parm_estim.md
```diff
@@ -0,0 +1,67 @@
...
+for i in indx
+    data = vcat(data , (sol.u[i] , sol.t[i]))
+end
+
+```
+Then we define the additional loss funciton `additional_loss(phi, θ , p)`, the function has three arguments, `phi` the trial solution, `θ` the parameters of neural networks, and `p` optional parameters.
+
+```julia
+function additional_loss(phi, θ , p)
+	l = length(θ) - 2
```
zoemcc, line 54: I don't love that the user has to know that the parameters to estimate are appended to the end of the θ vector internally and then slice off those again to evaluate the neural networks they pass in as part of an additional_loss.  Can we provide functions that abstract that for them?

ChrisRackauckas: In the additional loss interface, we could split out the parameters already (call the last argument `hyperp`)

zoemcc: Yeah, I like that. so the loss function internally splits them before it passes to the user-provided function

docs/src/pinn/parm_estim.md
zoemcc, line 50:
```suggestion
Then we define the additional loss function `additional_loss(phi, θ , p)`, the function has three arguments, `phi` the trial solution, `θ` the parameters of neural networks, and `p` optional parameters.
```

docs/src/pinn/parm_estim.md
```diff
@@ -0,0 +1,67 @@
+# Optimising Parameters of a Lorenz System
+ Consider a Lorenz System ,
+! [lorenzSystem](https://raw.githubusercontent.com/ashutosh-b-b/github-doc-images/master/lorenz_system.png) with Physics-Informed Neural Networks.
+Now we would consider the case when we know σ = 10 and we want to optimise β  and ρ.
+We start by defining the the problem,
+```julia
```
KirillZubov, line 6: Add all used dependencies so that the example can be easy copied-past like in other examples

docs/src/pinn/parm_estim.md
```diff
@@ -0,0 +1,67 @@
...
+n = 16
+chain1 = FastChain(FastDense(input_,n,Flux.σ),FastDense(n,n,Flux.σ),FastDense(n,n,Flux.σ),FastDense(n,1))
+chain2 = FastChain(FastDense(input_,n,Flux.σ),FastDense(n,n,Flux.σ),FastDense(n,n,Flux.σ),FastDense(n,1))
+chain3 = FastChain(FastDense(input_,n,Flux.σ),FastDense(n,n,Flux.σ),FastDense(n,n,Flux.σ),FastDense(n,1))
+```
+We will add an additional loss term based on the data that we have in order to optimise the parameters.
+Here we simply calculate the solution of the lorenz system with [OrdinaryDiffEq.jl](https://diffeq.sciml.ai/v1.10/tutorials/ode_example.html#In-Place-Updates-1) and pick up some points randomly
+
+```julia
+function lorenz!(du,u,p,t)
```
KirillZubov, line 32: `using OrdinaryDiffEq`

src/pinns_pde_solve.jl
KirillZubov: I think that the additional loss function should be excluded from the interface, for this there is a Low-level API.

KirillZubov: `additional_loss` is better renamed as `optim_p__loss` or something like that

ChrisRackauckas: >I think that the additional loss function should be excluded from the interface, for this, there is a Low-level API.

I don't get what you mean by that. The low level API is for inspection. This is for injecting code into the loss.

>additional_loss is better renamed as optim_p__loss or something like that

There are many reasons for adding loss terms other than for optimization of parameters, like regularization.

test/NNPDE_tests.jl
```diff
@@ -331,7 +332,7 @@ f_ = OptimizationFunction(loss_function_, GalacticOptim.AutoZygote())
 prob = GalacticOptim.OptimizationProblem(f_, initθ)
 
 cb_ = function (p,l)
-    println("Current losses are: ", pde_loss_function(p), " , ",  bc_loss_function(p))
```
KirillZubov, line 334: it works fine, there's no need to change `cb` here

KirillZubov:
update `symbolic_discretize`

ChrisRackauckas:
> update symbolic_discretize

Is it not just the same code, like `discretize` is not just RuntimeGeneratedFunction'd `symbolic_discretize`?

KirillZubov:
> > update symbolic_discretize
> 
> Is it not just the same code, like `discretize` is not just RuntimeGeneratedFunction'd `symbolic_discretize`?

no, two different,

ChrisRackauckas:
that's something to fix ASAP

src/pinns_pde_solve.jl
KirillZubov: there is no problem with the GPU

KirillZubov: but there is a conflict with PDE systems. Theta and PHI are also splitting there.

KirillZubov:
> that's something to fix ASAP

ok

KirillZubov:
although this is not an entirely different code they both call build_symbolic_loss_function which guarantees the same symbolic representation

src/pinns_pde_solve.jl
KirillZubov: for this task, it seems to be better not to use an additional loss function, but to use a loss function specialized for this task to optimize the parameters. And we can not take it out to the external interface, but generate it under the hood. Add method `get_optimize_p_loss_function`.

KirillZubov: Actually, we have already encountered this https://github.com/SciML/NeuralPDE.jl/pull/136

test/NNPDE_tests.jl
```diff
@@ -449,3 +450,55 @@ u_predict  = [first(phi(x,Array(res.minimizer))) for x in xs]
...
+    _loss(u , t; t_ = cu[t]) = sum(abs2, phi[i](t_ , θ[(i*593 - 593 + 1):(i*593)])[1] - (u[i])  for i in 1:1:3)
+    return sum(abs2, _loss(u,t) for (u , t) in data)/length(data)
+end
+
+input_ = length(domains)
+n = 16
+chain1 = FastChain(FastDense(input_,n,Flux.σ),FastDense(n,n,Flux.σ),FastDense(n,n,Flux.σ),FastDense(n,1))
+chain2 = FastChain(FastDense(input_,n,Flux.σ),FastDense(n,n,Flux.σ),FastDense(n,n,Flux.σ),FastDense(n,1))
+chain3 = FastChain(FastDense(input_,n,Flux.σ),FastDense(n,n,Flux.σ),FastDense(n,n,Flux.σ),FastDense(n,1))
+discretization = NeuralPDE.PhysicsInformedNN([chain1 , chain2, chain3],NeuralPDE.GridTraining(dt), param_estim=true, additional_loss=additional_loss)
```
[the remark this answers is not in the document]
ashutosh-b-b: We wont be sure about the type of loss that we want there. Like Chris said, this could be a regularisation loss as well or another loss defined on the data. So passing the loss would be a better option right?

src/pinns_pde_solve.jl
ashutosh-b-b: How do we pass data in that case? By defining a condition over each point in data?

ashutosh-b-b: @KirillZubov we'll have to define a regularisation loss separately then?

KirillZubov: > How do we pass data in that case? By defining a condition over each point in data?

data can be passed to `PhysicsInformedNN` instead of loss func

KirillZubov: we can even create a separate structure that will store everything that needs to optimize the parameters - data and all the thing

ashutosh-b-b: > > How do we pass data in that case? By defining a condition over each point in data?
> 
> data can be passed to `PhysicsInformedNN` instead of loss func

But then what will our eventual loss function be? It can be L2, L1, or anything that the user might want right?

KirillZubov: PhysicsInformedNN([chain1 , chain2, chain3], GridTraining(dt), EstimatiomParametersProblem(data, ...))

KirillZubov: > @KirillZubov we'll have to define a regularisation loss separately then?

what do you mean by a regularisation loss?

ashutosh-b-b: If we want to add a regularisation term with the loss, such as Ridge Regression. Even ` EstimatiomParametersProblem(data, ...)` after we define this, we'll still need input for an optional/additional loss

ChrisRackauckas: There isn't even 1 loss function for loss against data. There isn't 1 regularization method either. We tried this in DiffEqParamEstim.jl: you will not cover all loss functions. What about if they only have data for 2 of the 3 series? What if they have data only for the derivatives? It's completely impossible to handle like that. We already learned this, so I'm going to say that's a hard no. The user needs a hook to define a loss function which encloses data, there isn't a real alternative there.

ChrisRackauckas:
>although this is not an entirely different code they both call build_symbolic_loss_function which guarantees the same symbolic representation

I see. But how would it not just fail without changing `symbolic_discretize` then? It seems like they got out of sync which shouldn't happen.

docs/src/pinn/parm_estim.md
ChrisRackauckas: that `cu` looks like it's a typo?

ashutosh-b-b: Umm, Yes that'll get removed in the next commits

src/pinns_pde_solve.jl
KirillZubov: ok

KirillZubov: it would be nice to avoid explicit indexing inside the loss function and, in general, to make the interface as clear and compact as possible for the user.
```julia
function get_separete(chains)
    initθs = DiffEqFlux.initial_params.(chains)
    acum =  [0;accumulate(+, length.(initθs))]
    separete = [acum[i]+1 : acum[i+1] for i in 1:length(acum)-1]
end

sep = get_separete([chain1,chain2,chain3])

function additional_loss(phi, θ , p)
    _loss(u, t) = sum(abs2, phi[i](t , θ[s])[1] - (u[i])  for (i,s) in enumerate(sep))
    return sum(abs2, _loss(u,[t]) for (u , t) in data)/length(data)
end
```

KirillZubov: as @zoemcc   already mentioned, if this is an additional general purpose function then it should not depend on `param_estim` if not and this is a function only for parameter optimization then it should be renamed.

KirillZubov: it turns out, that `eqs` and `bcs` are in symbolic representation and there is no additional function. And we get different:  symbolic in one` x(t),  Dt(x(t)) β`, in other in inner  representation  `phi[1](data[i]), derivative(u, phi, ....eps), θ[end-1]`.

KirillZubov: Maybe it makes sense to have the symbolic optimization condition.
Something like that:
```julia
@parameters t_d, x_d, y_d, z_d
optim_param_condition = [x(t_d) - x_d,
                                             y(t_d) - z_d,
                                             z(t_d)) - y_d]
EstimatiomParametersProblem(optim_param_condition,[t_d,x_d,y_d,z_d], data)
```

ChrisRackauckas: We shouldn't add another Problem type. It doesn't match the interface.

ChrisRackauckas: >as @zoemcc already mentioned, if this is an additional general purpose function then it should not depend on param_estim if not and this is a function only for parameter optimization then it should be renamed.

Agreed on this.

ashutosh-b-b:
lorenz_system

plot_lorenz

ChrisRackauckas:
Looks like a real test failure: https://github.com/SciML/NeuralPDE.jl/pull/256/checks?check_run_id=2035703967#step:6:8103

docs/src/pinn/parm_estim.md
ChrisRackauckas: Bump

src/pinns_pde_solve.jl
ChrisRackauckas: This is still probably true.

ChrisRackauckas:
```suggestion
            push!(expr_params, :(ArrayInterface.allowed_getindex(p,$i)))
```

ChrisRackauckas:
You'll need to add an ArrayInterface dep because of the commit I just put in there. That will make parameter estimation GPU-safe, since otherwise it'll failed with not allowing scalar indexing. Otherwise, the last thing is, as @KirillZubov mentioned, the tutorial should just be able to copy paste. Make sure you add all of the `using` statements required to make that true. Once that's done, I think this is good to go.

src/pinns_pde_solve.jl
```diff
@@ -335,28 +363,29 @@ function build_symbolic_loss_function(eqs,indvars,depvars,
...
     depvars,indvars,dict_indvars,dict_depvars = get_vars(_indvars, _depvars)
     bc_indvars = bc_indvars==nothing ? indvars : bc_indvars
     return build_loss_function(eqs,indvars,depvars,
                                dict_indvars,dict_depvars,
                                phi, derivative,initθ,
-                               bc_indvars = bc_indvars)
+                               bc_indvars = bc_indvars, eq_params=eq_params,param_estim=param_estim,default_p=default_p)
 end
 
 function build_loss_function(eqs,indvars,depvars,
```
KirillZubov, line 376: `build_loss_function` should be RuntimeGeneratedFunction's wrapper for `build_symbolic_loss_function` and have the same parameter for both.

ashutosh-b-b:
@ChrisRackauckas Do I have to make any additional changes apart from resolving conflicts?

ChrisRackauckas (approved):

ChrisRackauckas:
I think it's good. @KirillZubov any comments?

ChrisRackauckas:
@ashutosh-b-b see https://github.com/SciML/NeuralPDE.jl/pull/256#discussion_r589249132

ChrisRackauckas:
And yes, resolve the conflicts and rebase the code. This is close.

src/pinns_pde_solve.jl
ashutosh-b-b: > @ashutosh-b-b see [#256 (comment)](https://github.com/SciML/NeuralPDE.jl/pull/256#discussion_r589249132)



> @ashutosh-b-b see [#256 (comment)](https://github.com/SciML/NeuralPDE.jl/pull/256#discussion_r589249132)

Yes I changed them to have the same parameters.

docs/src/pinn/parm_estim.md
```diff
@@ -8,6 +8,7 @@ Now we would consider the case where we want to optimise the parameters σ, β
 We start by defining the the problem,
 
 ```julia
+using OrdinaryDiffEq
```
ChrisRackauckas, line 11: don't you need to use NeuralPDE at least?

src/pinns_pde_solve.jl
```diff
@@ -854,9 +892,19 @@ function DiffEqBase.discretize(pde_system::PDESystem, discretization::PhysicsInf
     end
 
     function loss_function_(θ,p)
-        return pde_loss_function(θ) + bc_loss_function(θ)
+        if additional_loss isa Nothing
+            return pde_loss_function(θ) + bc_loss_function(θ)
+        else
+            function _additional_loss(phi,θ)
+                θ = θ[1:end - length(default_p)]
```
KirillZubov, line 899: it is needed to rename inner `θ`
```
default_p = 2
θ  = [1,2,3,4,5]
julia> θ = θ[1:end - length(default_p)]
4-element Array{Int64,1}:
 1
 2
 3
 4
julia> p = θ[end - length(default_p):end]
2-element Array{Int64,1}:
 3
 4
```

ChrisRackauckas: `_θ`.

test/NNPDE_tests.jl
```diff
@@ -484,3 +485,68 @@ u_predict  = [first(phi(x,res.minimizer)) for x in xs]
...
+    data = []
+    indx = rand(1:1:21 , 12)
+    for i in indx
+        data = vcat(data , (sol.u[i] , sol.t[i]))
+    end
+    return data
+end
+data = getData(sol)
+#Additional Loss Function
+function additional_loss(phi, θ , p)
```
KirillZubov, line 525: θ = [flat_θ; estim_param] 
so it's needed here to not grab all param and only flat_θ

```julia
initθs = DiffEqFlux.initial_params.([chain1,chain2,chain3])
acum =  [0;accumulate(+, length.(initθs))]
sep = [acum[i]+1 : acum[i+1] for i in 1:length(acum)-1]
len = length(data)
function additional_loss(phi, θ , p)
    _loss(u , t) = sum(abs2, phi[i](t , θ[sep[i]])[1] - u[i]  for i in 1:1:3)
    global data
    return sum(abs2, _loss(u,t) for (u , t) in data)/len
end
```

KirillZubov:
I add the vectorized version for add_loss, everything is the same only in vectorized form

docs/src/pinn/parm_estim.md
```diff
@@ -0,0 +1,85 @@
...
+data = []
+indx = rand(1:1:21 , 12)
+for i in indx
+    data = vcat(data , (sol.u[i] , sol.t[i]))
+end
+```
+Then we define the additional loss funciton `additional_loss(phi, θ , p)`, the function has three arguments, `phi` the trial solution, `θ` the parameters of neural networks, and the hyperparameters `p` .
+
+```julia
+function additional_loss(phi, θ , p)
```
KirillZubov, line 55: here, too, correct the `additional loss function` as in the test

test/NNPDE_tests.jl
```diff
@@ -484,3 +485,77 @@ u_predict  = [first(phi(x,res.minimizer)) for x in xs]
...
+#Plotting the system
+# initθ = discretization.init_params
+# acum =  [0;accumulate(+, length.(initθ))]
+# sep = [acum[i]+1 : acum[i+1] for i in 1:length(acum)-1]
+# minimizers = [res.minimizer[s] for s in sep]
+# ts = [domain.domain.lower:dt/10:domain.domain.upper for domain in domains][1]
+# u_predict  = [[discretization.phi[i]([t],minimizers[i])[1] for t in ts] for i in 1:3]
+# plot(sol)
+# plot!(ts, u_predict, label = ["x(t)" "y(t)" "z(t)"])
+p_
```
KirillZubov, line 561: delete it

docs/src/pinn/parm_estim.md
```diff
@@ -0,0 +1,89 @@
...
+cb = function (p,l)
+    println("Current loss is: $l")
+    return false
+end
+res = GalacticOptim.solve(prob, BFGS(); cb = cb, maxiters=5000)
+p_ = res.minimizer[end-2:end] # p_ = [9.93, 28.002, 2.667]
+```
+And then finally some analyisis by plotting.
+```julia
+initθ = discretization.init_params
```
KirillZubov, line 80: update plots code how it is in the test, and image

KirillZubov (approved):

ChrisRackauckas:
I think it looks good. I'm going to merge when tests pass, and any other minor tweaks can be a follow PR at this point. I think it's clear this is correct and it's the "right code", even if small doc changes might follow. Awesome work @ashutosh-b-b

ChrisRackauckas:
@KirillZubov what is this test failure?

https://github.com/SciML/NeuralPDE.jl/pull/256/checks?check_run_id=2095211069#step:6:14775

KirillZubov:
> @KirillZubov what is this test failure?
> 
> https://github.com/SciML/NeuralPDE.jl/pull/256/checks?check_run_id=2095211069#step:6:14775

Example 7, Fokker-Planck equation is unstable. This example worked well with BFGS before vectorization, but now after vectorization, it is unstable and interrupted. I don’t really understand what can be done yet.

ChrisRackauckas:
Okay, can you prioritize that? It also is causing red tests in MTK and Symbolics. I'll take a look if the optimization strategy should be tweaked. ADAM 300 iterations before BFGS right?

KirillZubov:
> Okay, can you prioritize that? It also is causing red tests in MTK and Symbolics. I'll take a look if the optimization strategy should be tweaked. ADAM 300 iterations before BFGS right?

what do you mean, prioritize that? Yeah, alternating between ADAM and BFGS works, but not stable for this example.

ChrisRackauckas:
The BFGS should do `BFGS(initial_stepnorm=0.01)` so it doesn't pop out of the region. It should be like:

ADAM
BFGS initial stepnorm = 0.01
change dx
BFGS initial stepnorm = 0.01
change dx
BFGS initial stepnorm = 0.01

otherwise you'll keep popping out.

KirillZubov:
ok

PR #260, opened 2021-03-12
Remove starter BFGS in Fokker-Plank test?

ChrisRackauckas:


KirillZubov:
This PR can be merged. The Fokker-Planck test should be stable now

PR #274, opened 2021-03-22
[WIP] update parser to new  MTK

KirillZubov:


ChrisRackauckas:
@YingboMa what happened here?

KirillZubov:
> @YingboMa what happened here?

@ChrisRackauckas `toexpr` changed behavior. So it needs update parser in PINNs.

KirillZubov:
```julia
julia> eq
Differential(t)(u(t, x, y)) ~ Differential(x)(Differential(x)(u(t, x, y))) + Differential(y)(Differential(y)(u(t, x, y)))

julia> eq_expr = toexpr(eq)
:(uˍt(t) = (+)(uˍxx(t), uˍyy(t)))
```

KirillZubov:
it looks like a bug, why parameters are reducing.

KirillZubov:
```julia
@parameters x y
@variables u(..)
Dxx = Differential(x)^2
Dyy = Differential(y)^2
Dx = Differential(x)
Dy = Differential(y)
ex = Dx(Dy(u(x,y))) 
julia> toexpr(ex)
ERROR: ArgumentError: non-matching differentials on lhs: y, x
Stacktrace:
 [1] flatten_differential(::Term{Real,Nothing}) at /Users/kirillzubov/.julia/packages/Symbolics/JhAD8/src/utils.jl:103
 [2] diff2term(::Term{Real,Nothing}) at /Users/kirillzubov/.julia/packages/Symbolics/JhAD8/src/utils.jl:93
 [3] function_to_expr(::Differential, ::Term{Real,Nothing}, ::SymbolicUtils.Code.LazyState) at /Users/kirillzubov/.julia/packages/Symbolics/JhAD8/src/diff.jl:499
 [4] toexpr(::Term{Real,Nothing}, ::SymbolicUtils.Code.LazyState) at /Users/kirillzubov/.julia/packages/SymbolicUtils/0Fgyc/src/code.jl:124
 [5] toexpr(::Term{Real,Nothing}) at /Users/kirillzubov/.julia/packages/SymbolicUtils/0Fgyc/src/code.jl:74
 [6] toexpr(::Num) at /Users/kirillzubov/.julia/packages/Symbolics/JhAD8/src/num.jl:188
 [7] top-level scope at none:1
```

YingboMa:
I am confused. The tests pass, so what do you need to update?

Also, if you want to update, then stop trying to transform the generated Julia expressions, instead, just transform on the symbolic expression. We don't guarantee how we generate Julia expressions as long as it correctly implements the symbolic expression. I don't want to have to make sure NeuralPDE doesn't break because it is implemented in this totally rigid way when Symbolics changes some minutia. This gets annoying very quickly. This has happened like 4 times.

YingboMa:
Oops, I see. This is PR is opened last month. I got the notification again because of the new commit.

PR #286, opened 2021-04-08
Add files via upload adaptive function for PINN

Necromuda:


ChrisRackauckas:
This doesn't seem to be relevant.

PR #287, opened 2021-04-08
Add files via upload PINN adaptive function

yadhukartha:
sir this is the function i am talking about


ChrisRackauckas:
Copy pasting the Python code from the paper isn't a solution. It needs to be added to the symbolic parser, i.e. https://github.com/SciML/NeuralPDE.jl/blob/master/src/pinns_pde_solve.jl

yadhukartha:
Sir but that code isn't in any paper, I made the function using 2 functions

PR #298, opened 2021-04-20
[WIP] Added better support for heterogeneous input dependent variables

zoemcc:
Added better support for heterogeneous inputs via a new dictionary dict_depvar_input that contains for each dependent variable symbol: a list of the independent variables it depends on, in order.  This was added to support the SPM model.  This dictionary is created in get_vars by examining if a depvar is applied to variables such as Q(t) or Q(t, r) and then it populates the dict based off of those.  If a dependent variable isn't applied to independent variables then it populates the dictionary by the default of all the independent variables in order.

This makes the SPM model run, I need to verify that all the previous tests are unbroken and add docs/test for the new syntax.  This should in theory be backwards compatible for most models that aren't heterogeneous in input, but will likely break older heterogeneous models that don't include their input variables in the PDESystem dependent variable Nums.  There were a few issues, the biggest two being supporting dependent variables using different orderings of the independent variables than expected in dict_indvars, and also keeping track of substitutions for boundary conditions of these heterogeneous inputs.

Here is an example of the syntax now supported (on a bogus system, but illustrates the heterogeneity):
```julia
@parameters t x y
@variables Q(..) P(..) Z(..)
Dt = Differential(t)
Dx = Differential(x)
Dy = Differential(y)

eqs = [Dt(Q(t)) ~ 1,
           Dt(P(t, x)) ~ Dx(P(t, x)),
           Dt(Z(t, y)) ~ Dy(Z(t, y))]

ics_bcs = [Q(0) ~ 0,
                 P(0, x) ~ x,
                 Z(0, y) ~ y]

t_domain = IntervalDomain(0.0, 1.0)
x_domain = IntervalDomain(0.0, 1.0)
y_domain = IntervalDomain(0.0, 1.0)

ind_vars = [t, x, y]
dep_vars = [Q(t), P(t, x), Z(t,y)]
pde_system = PDESystem(eqs, ics_bcs, domains, ind_vars, dep_vars)
```

### DIFF
--- a/src/pinns_pde_solve.jl
+++ b/src/pinns_pde_solve.jl
@@ -178,11 +178,19 @@ function _transform_expression(ex,dict_indvars,dict_depvars, dict_depvar_input, 
                 num_depvar = dict_depvars[depvar]
                 indvars = _args[2:end]
                 cord = :cord
+                interior_u_args = map(_args[2:end]) do arg
+                    if arg isa Symbol
+                        arg
+                    else
+                        :(fill($arg,(1, $:batch_size)))
+                    end
+                end
                 ex.args = if !(typeof(chain) <: AbstractVector)
-                    [:u, cord, :($θ), :phi]
+                    [:u, :(vcat($(interior_u_args...))), :($θ), :phi]
                 else
                     #[:u, cord, Symbol(:($θ),num_depvar), Symbol(:phi,num_depvar)]
-                    [:u, Symbol(:phi_in,num_depvar), Symbol(:($θ),num_depvar), Symbol(:phi,num_depvar)]
+                    #[:u, Symbol(:phi_in,num_depvar), Symbol(:($θ),num_depvar), Symbol(:phi,num_depvar)]
+                    [:u, :(vcat($(interior_u_args...))), Symbol(:($θ),num_depvar), Symbol(:phi,num_depvar)]
                 end
                 break
             elseif e isa ModelingToolkit.Differential
@@ -289,14 +297,17 @@ to
           end
       end)
 """
-function build_symbolic_loss_function(eqs,_indvars,_depvars,dict_depvar_input, phi, derivative,chain,initθ,strategy; bc_indvars=nothing , eq_params = SciMLBase.NullParameters(), param_estim = false,default_p=nothing)
+function build_symbolic_loss_function(eqs,_indvars,_depvars,dict_depvar_input, phi, derivative,chain,initθ,strategy; 
+    bc_indvars=nothing , eq_params = SciMLBase.NullParameters(), param_estim = false,default_p=nothing, integration_indvars=nothing)
     # dictionaries: variable -> unique number
     depvars,indvars,dict_indvars,dict_depvars,dict_depvar_input = get_vars(_indvars, _depvars)
     bc_indvars = bc_indvars==nothing ? indvars : bc_indvars
+    integration_indvars = integration_indvars==nothing ? indvars : integration_indvars
     return build_symbolic_loss_function(eqs,indvars,depvars,
                                         dict_indvars,dict_depvars,dict_depvar_input,
                                         phi, derivative,chain,initθ,strategy,
-                                        bc_indvars = bc_indvars, eq_params = eq_params, param_estim = param_estim,default_p=default_p)
+                                        bc_indvars = bc_indvars, eq_params = eq_params, param_estim = param_estim,default_p=default_p,
+                                        integration_indvars = integration_indvars)
 end
 
 function get_indvars_ex(bc_indvars)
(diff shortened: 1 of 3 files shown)


ChrisRackauckas:
Yes, this needs to be done to match DiffEqOperators.jl anyways. Is the only reason why this is only WIP because docs and tests need to be updated?

zoemcc:
>  Is the only reason why this is only WIP because docs and tests need to be updated?

Yes, that is correct.  I've only verified it on the SPM model and am focusing on that for the next day.

zoemcc:
The previous version was *definitely* not correct.  It's almost correct now, and it is correct for StochasticTraining (still going through and fixing the inputs for the other trainings).

This was a much larger change than I initially thought due to my misunderstanding in how the inputs are loaded into the loss expressions.  

I had to refactor the way that inputs are loaded in to make it more general, there were a few assumptions baked in.  In particular, there was an assumption that an expression's independent variables could only have one value in each expression, so equations like `q(1) ~ q(-1)` were being parsed and evaluated incorrectly.  

The new process of loading in inputs for an expression is:


1. Find any independent variables that enter an equation symbolically `:x` or `:y`, etc.  These are the integration independent variables.  For example, in the equation `q(x, 1) ~ 0`, for the dependent variable `q(x,y)`, `x` is an integration independent variable but `y` is not.  The function `get_integration_variables` does this process (modified from `get_argument`).  The integration independent variables will form the `cord` for each loss function.
2. During the `transform_expression` function, inputs to dependent variables are converted to `fill` statements of the appropriate size if they are not a symbol.  The input to `u` or `derivative` are not `cord` anymore, but rather a `vcat` of the constants and rows of `cord` that are the inputs to that dependent variable, in the correct order.
3. Within `build_symbolic_loss_function`, make sure that the `let` expression contains the correct rows of `cord` so that they can be `vcat` with each other, and also the size of the batch for the `fill` expressions to be the correct size.

This new focus on integration variables will simplify the interface to each of the training strategies since they can basically all use the same bound information based on the integration variables, and then each phi input can construct the inputs that it needs from the integration variables and the constants embedded in the equation.  

I'm going to finish changing the inputs to each strategy tomorrow so they're using the new integration variables but it might make sense to hold off merging this into master until it can be cleaned up a bit [likely after our deadline].

ChrisRackauckas:
Can we get a version of this finished so we can merge the depvar change?

src/pinns_pde_solve.jl
```diff
@@ -242,20 +262,22 @@ Example:
...
     eq_rhs = isequal(expand_derivatives(eq.rhs), 0) ? eq.rhs : expand_derivatives(eq.rhs)
 
-    left_expr = transform_expression(toexpr(eq_lhs),dict_indvars,dict_depvars,chain,initθ,strategy)
-    right_expr = transform_expression(toexpr(eq_rhs),dict_indvars,dict_depvars,chain,initθ,strategy)
-    left_expr = Broadcast.__dot__(left_expr)
-    right_expr = Broadcast.__dot__(right_expr)
+    left_expr = transform_expression(toexpr(eq_lhs),dict_indvars,dict_depvars, dict_depvar_input,chain,initθ,strategy)
+    right_expr = transform_expression(toexpr(eq_rhs),dict_indvars,dict_depvars, dict_depvar_input,chain,initθ,strategy)
+
+    #left_expr = Broadcast.__dot__(left_expr)   #TODO: ZDM: the new way of filling args & constants to the phi inputs is broken by these
```
KirillZubov, line 278: the problem here, that `__dot__`  is broadcasted all functions that in expression `adapt,parameterless_type,vcat,fill`

```julia
 Expr[:((cord, var"##θ#466", phi, derivative, u, p)->begin
          begin
              let (batch_size,) = (1,)
                  u.(adapt.(DiffEqBase.parameterless_type.(var"##θ#466"), vcat.(fill.(0.0, (1, batch_size)))), var"##θ#466", phi) .- 1.0
              end
          end
      end
```
Simple way redefine broadcasted for these functions like it done for `get_u`:
`Base.Broadcast.broadcasted(::typeof(get_u()), cord, θ, phi) = get_u()(cord, θ, phi)`

src/pinns_pde_solve.jl
```diff
@@ -309,11 +334,26 @@ function get_indvars_ex(bc_indvars)
...
                                       phi, derivative, chain,initθ, strategy; eq_params = SciMLBase.NullParameters(), param_estim = param_estim,default_p=default_p,
-                                      bc_indvars = indvars)
+                                      bc_indvars = indvars,
+                                      integration_indvars = indvars)
+
+    #=
+    this_eq_dict_indvars = get_dict_vars(bc_indvars)
+    
+
+    # this is slightly inefficient, some of these might not get used in the parsed equation, but this is a superset of used dep vars
```
KirillZubov, line 346: yeap, I agree

KirillZubov:
Sorry for the long asnwer. 
@zoemcc could you please explain why `q(1) ~ q(-1) ` is incorrect. this condition for symmetry and periodicity in some cases like `cos(pi) = cos(-pi)`

KirillZubov:
https://github.com/SciML/NeuralPDE.jl/issues/339

killah-t-cell:
Is this PR still active?

src/pinns_pde_solve.jl
killah-t-cell: Does this https://github.com/SciML/NeuralPDE.jl/pull/338/files fix the broadcast piracy problem? What would be the right way to do this now?

Issue #288, opened 2021-04-09
QuasiRandomTraining,  every training step take a new quasirandom sample


KirillZubov:
in QuasiRandomTraining, we first create a minibatch from quasi-random numbers, and at each iteration of training, we take one of them.

it needs to implement option `resample` - every training step takes a new quasirandom sample.

`QuasiRandomTraining(points;sampling_alg = UniformSample(),minibatch=500, resample= true): `

`resample` should guarantee that on each iteration it's used new quasi-random points without reusing one set.

PR #314, opened 2021-06-01
QuasiRandomTraining, on every training iteration, it is generated a new quasirandom sample

KirillZubov:
https://github.com/SciML/NeuralPDE.jl/issues/288

### DIFF
--- a/docs/src/solvers/pinns.md
+++ b/docs/src/solvers/pinns.md
@@ -56,14 +56,11 @@ List of training strategies that are available now:
    to the spacing per direction.
  - `StochasticTraining(points)`: `points` number of stochastically sampled points from the domain.
    In each optimization iteration, we randomly select a new subset of points from a full training set.
- - `QuasiRandomTraining(points;sampling_alg = UniformSample(),minibatch=500)`:
+ - `QuasiRandomTraining(points;sampling_alg = UniformSample(), resampling = true, minibatch=500)`:
    The training set is generated on quasi-random low discrepency sequences.
-   `minibatch` is the number of subsets, where `points` is the number of quasi-random points in minibatch.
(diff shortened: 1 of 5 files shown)


ChrisRackauckas:
Resampling should be an option (true by default though).

KirillZubov:
@ChrisRackauckas almost all methods do not support automatic differentiation except GridSample.

```julia
using QuasiMonteCarlo, Distributions,Zygote
lb = [0.1,-0.5]
ub = [1.0,20.0]
n = 5
d = 2

p = Float32.([1.5,2.0])
function testf(p)
     x = Float32.(QuasiMonteCarlo.sample(n,lb,ub,GridSample([0.1,0.5])))
     sum(sin.(x .* p))
end
dp1 = Zygote.gradient(testf,p)
```
for all other methods, this results in an error "Mutating arrays is not supported"
so `resampling` work only with GridSample([0.1,0.5]) and we need to set the step dx

ChrisRackauckas:
I don't get what you mean. `resample` shouldn't work with GridTraining at all because the points would be fixed?

KirillZubov:
GridSample is one of methods in  QuasiMonteCarlo.jl

KirillZubov:
almost all Sampling Methods in QuasiMonteCarlo.jl  https://github.com/SciML/QuasiMonteCarlo.jl#available-sampling-methods

KirillZubov:
and since they do not support automatic differentiation (Zygote don't work with Mutating array), we while cannot use them for resampling (create at runtime on each iteration)

ChrisRackauckas:
Wrap it in a function and do `@nograd` on it. We shouldn't be differentiating the sampling.

KirillZubov:
yeap, thanks ,it works

Issue #296, opened 2021-04-17
Allow number of boundary points to be chosen by the user by a keyword argument


ChrisRackauckas:
The title should be fairly descriptive here.

PR #316, opened 2021-06-03
Allow number of boundary points

KirillZubov:
https://github.com/SciML/NeuralPDE.jl/issues/296


KirillZubov:
It is not related to this PR.
buildkite build  failed: `Error: CUDA.jl does not yet support CUDA with nvdisasm 11.3.58; please file an issue.`

ChrisRackauckas:
That is not your issue. Your issue is: https://buildkite.com/julialang/neuralpde-dot-jl/builds/388#0241f524-0630-4513-b7e9-d26615f738be/415-686 . I think that scalar indexing could be related to the PR?

KirillZubov:
the same error with Neural PDE v3.11.1. Test failed with GridTraining so it definitely not related to this PR. I think it's better to fix the GPU issue in the next PR.

PR #318, opened 2021-06-04
GPU fix

KirillZubov:


src/pinns_pde_solve.jl
```diff
@@ -600,7 +600,7 @@ function get_numeric_derivative()
     derivative =
         (phi,u,x,εs,order,θ) ->
         begin
-            _epsilon = 1 / (2*cbrt(eps(eltype(θ))))
+            _epsilon = 1.f0 / (2*cbrt(eps(eltype(θ))))
```
ChrisRackauckas, line 603:
```suggestion
            _epsilon = one(eltype(θ)) / (2*cbrt(eps(eltype(θ))))
```

src/pinns_pde_solve.jl
```diff
@@ -631,7 +631,7 @@ function generate_random_points(points, bound)
            fill(Float32(b),(1,points))
        else
            lb, ub =  b[1], b[2]
-           lb .+ (ub .- lb) .* Float32.(rand(1,points))
+           Float32.(lb .+ (ub .- lb) .* (rand(1,points)))
```
ChrisRackauckas, line 634: `Float32` should never be used. It should all be based off of input types.

ChrisRackauckas:
I think something happened with your commit? It should be like `eltype(x)` and `convert.(eltype(x),...)` etc.

src/pinns_pde_solve.jl
```diff
@@ -156,8 +156,8 @@ function transform_expression(ex,dict_indvars,dict_depvars, chain,initθ, strate
 end
 
 function get_ε(dim, der_num)
-    epsilon = cbrt(eps(Float32))
-    ε = zeros(Float32, dim)
+    epsilon = cbrt(eps(eltypeθ))
```
ChrisRackauckas, line 159:
```suggestion
    epsilon = cbrt(eps(eltype(θ)))
```

KirillZubov: we do not have access to θ from this method, and the method is executed before training when assembling the model, so there is no need to do dynamic determine the type of θ.

ChrisRackauckas: got it

ChrisRackauckas: it's never dynamic BTW if type-stable. It's hoisted to compile time.

KirillZubov: yes, it is type-stable always,  and it is not necessary to define the type dynamically as we did before in the loss function(`DiffEqBase.parameterless_type(θ)`), it is enough to do it once during compilation.

ChrisRackauckas: gotcha

KirillZubov:
@ChrisRackauckas  can we calculate in Float64? or the chain still works with accuracy float32 and just convert with noise in the last figures?
 ```julia
 chain = FastChain(FastDense(1,20,Flux.σ),FastDense(20,1))
 chain(rand(1,3),initθ)
1×3 Matrix{Float64}:
 -0.254752  -0.241568  -0.276606
```

ChrisRackauckas:
If the weight and the input are Float64 then the computational is at Float64 accuracy.

KirillZubov:
OK then there needs research that and add tests with Float64

ChrisRackauckas (approved):

PR #326, opened 2021-06-16
Extract num of points for quadrature.jl methods

KirillZubov:
https://github.com/SciML/NeuralPDE.jl/issues/322


KirillZubov:
now the loss function for the quadrature.jl method is calc as `1 / τ *loss(θ)`, where τ - count of points on the last iteration of numerical integration

ChrisRackauckas:
I think dividing by the number of points is a bit weird. To normalize, I think it would make more sense to divide by the area?

KirillZubov:
@ChrisRackauckas could you please review it

Issue #302, opened 2021-04-30
"Systems of PDEs" example in documentation does give the anticipated results


dohter:
Hello,

I am following the example https://neuralpde.sciml.ai/stable/pinn/system/ and I am not getting the same results in. Mainly there is big discrepancy in u2 and u3 outputs
![sol_u2](https://user-images.githubusercontent.com/23319636/116732582-427cb100-a9a8-11eb-981e-3baaa85f687f.png)
![sol_u3](https://user-images.githubusercontent.com/23319636/116732588-44467480-a9a8-11eb-98d3-f33d0aff329b.png)

And this is the code

```
cd(@__DIR__)
using Pkg
Pkg.activate("../")
using NeuralPDE, Flux, ModelingToolkit, GalacticOptim, Optim, DiffEqFlux

@parameters t, x
@variables u1(..), u2(..), u3(..)
Dt = Differential(t)
Dtt = Differential(t)^2
Dx = Differential(x)
Dxx = Differential(x)^2

eqs = [Dtt(u1(t,x)) ~ Dxx(u1(t,x)) + u3(t,x)*sin(pi*x),
       Dtt(u2(t,x)) ~ Dxx(u2(t,x)) + u3(t,x)*cos(pi*x),
       0. ~ u1(t,x)*sin(pi*x) + u2(t,x)*cos(pi*x) - exp(-t)]

bcs = [u1(0,x) ~ sin(pi*x),
       u2(0,x) ~ cos(pi*x),
       Dt(u1(0,x)) ~ -sin(pi*x),
       Dt(u2(0,x)) ~ -cos(pi*x),
       u1(t,0) ~ 0.,
       u2(t,0) ~ exp(-t),
       u1(t,1) ~ 0.,
       u2(t,1) ~ -exp(-t),
       u1(t,0) ~ u1(t,1),
       u2(t,0) ~ -u2(t,1)]


# Space and time domains
domains = [t ∈ IntervalDomain(0.0,1.0),
           x ∈ IntervalDomain(0.0,1.0)]
# Discretization
dx = 0.1
# Neural network
input_ = length(domains)
n = 8
chain1 = FastChain(FastDense(input_,n,Flux.σ),FastDense(n,n,Flux.σ),FastDense(n,1))
chain2 = FastChain(FastDense(input_,n,Flux.σ),FastDense(n,n,Flux.σ),FastDense(n,1))
[...]

PR #329, opened 2021-06-22
Fix PDAE system example and derivative neural network approximation

KirillZubov:
https://github.com/SciML/NeuralPDE.jl/issues/302

### DIFF
--- a/docs/src/pinn/system.md
+++ b/docs/src/pinn/system.md
@@ -50,12 +50,12 @@ n = 15
 chain =[FastChain(FastDense(input_,n,Flux.σ),FastDense(n,n,Flux.σ),FastDense(n,1)) for _ in 1:3]
 initθ = map(c -> Float64.(c), DiffEqFlux.initial_params.(chain))
 
-_strategy = NeuralPDE.QuadratureTraining()
-discretization = NeuralPDE.PhysicsInformedNN(chain, _strategy, init_params= initθ)
+_strategy = QuadratureTraining()
+discretization = PhysicsInformedNN(chain, _strategy, init_params= initθ)
 
 pde_system = PDESystem(eqs,bcs,domains,[t,x],[u1,u2,u3])
-prob = NeuralPDE.discretize(pde_system,discretization)
-sym_prob = NeuralPDE.symbolic_discretize(pde_system,discretization)
+prob = discretize(pde_system,discretization)
+sym_prob = symbolic_discretize(pde_system,discretization)
 
 pde_inner_loss_functions = prob.f.f.loss_function.pde_loss_function.pde_loss_functions.contents
 bcs_inner_loss_functions = prob.f.f.loss_function.bcs_loss_function.bc_loss_functions.contents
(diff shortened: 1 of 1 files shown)


docs/src/pinn/system.md
```diff
@@ -80,26 +82,156 @@ ts,xs = [infimum(d.domain):0.01:supremum(d.domain) for d in domains]
...
+```
+
+![sol_uq1](https://user-images.githubusercontent.com/12683885/122979254-03634e80-d3a0-11eb-985b-d3bae2dddfde.png)
+
+![sol_uq2](https://user-images.githubusercontent.com/12683885/122979278-09592f80-d3a0-11eb-8fee-de3652f138d8.png)
+
+![sol_uq3](https://user-images.githubusercontent.com/12683885/122979288-0e1de380-d3a0-11eb-9005-bfb501959b83.png)
+
+
+## Derivative neural network approximation
```
ChrisRackauckas, line 107: what does that mean?

KirillZubov: we approximate the derivative of the neural network with another neural network `Dt(u1(t,x)) ~ Dtu1(t,x)` thus we avoid taking the second derivative `Dt(Dtu1(t,x)) `.
Since `u3` is only in the first and second equations, that its accuracy during training is determined by the accuracy of the second numerical derivative `Dtt(u1(t,x)) ~ Dxx(u1(t,x)) + u3(t,x)*sin(pi*x)`.
So this is such the trick to not take the second numeric derivative that allows an increase in the marginal precision for optimization.

ChrisRackauckas: ahh this needs more of an explanation.

KirillZubov: https://github.com/SciML/NeuralPDE.jl/blob/Systems_of_PDE_fix/docs/src/pinn/system.md#comparison-of-the-second-numerical-derivative-and-numerical--neural-network-derivative

ChrisRackauckas: But there's no explanation. There's a title, and then there's no text explaining what the section is actually about. There needs to be some introductory text as to what you're doing there and why.

KirillZubov: oh, I have thought you need more explanation :) 
ok, I will add a text explanation to the doc.

KirillZubov:
I used to think that the problem with the systems is that the gradient is not updating correctly and therefore the third equation had a much higher error. It turned out that this was due to a derivative error. The error grows very quickly with the degree of the numerical derivative.
Therefore, we definitely need to update the calc of the numerical derivative and achieve greater accuracy and stability for order > 1 .

ChrisRackauckas:
Yeah, we just need Diffractor.jl done for that. Keno has examples.

KirillZubov:
And the second, regarding the adaptive reweighting,  the problem is not only that the derivatives have frequency scale amplification but also that each degree of the derivative has a large error, which clearly interferes with the gradient descent.

I guess instability gives a splash of the gradient.
max(grad(loss(higher_order_derivatives(u)+big error ))) >>  max(grad(loss(lower-order_derivatives(u))))

but this is only for numeric derivatives, automatic differentiation does not have such problems, as I understand it, although.

ChrisRackauckas:
Also the way the derivative is calculated here isn't great, it should probably use FiniteDifferences.jl if anything because you don't want to nest 1st derivatives and instead adapt the epsilon and all of that.

PR #330, opened 2021-06-23
Add Integro Differential Equations Support

ashutosh-b-b:
1. How do I add support for closure inside the `get_numeric_integral`? `get_numeric_integral` cannot be a fixed function since the number of dependent variables may vary.


KirillZubov:
why can't? , probably better name `parse_integral`(or something like this) then `get_numeric_integral` . You need to call this function for every symbol form of integral in the model and pass dependent variables for every integral.

KirillZubov:
my pseudo code for this :)
```
eqs = integral1(depvars1,...) + integral2(depvars2,...)+ somthing ->
inner_representation_integral1(depvars1) + inner_representation_integral2(depvars2)+ somthing

parser(integral1(depvars1, ...),...) -> inner_representation_integral1
```

src/pinns_pde_solve.jl
```diff
@@ -219,6 +219,14 @@ function _transform_expression(ex,dict_indvars,dict_depvars,chain,eltypeθ,strat
                     [:derivative, Symbol(:phi,num_depvar), :u, cord, εs_dnv, order, Symbol(:($θ),num_depvar)]
                 end
                 break
+            elseif e isa Symbolics.Integral
+                integrating_variable = toexpr(_args[1].x)
+                integrand = transform_expression(_args[2],dict_indvars,dict_depvars,chain,eltypeθ,strategy)
```
KirillZubov, line 224: _transform_expression

src/pinns_pde_solve.jl
```diff
@@ -654,6 +668,45 @@ function get_numeric_derivative()
         end
 end
 
+function get_numeric_integral(strategy, indvars, depvars, dict_indvars, chain, derivative)
+    integral =
+        (u, cord, phi, θ, integrating_variable, integrand, lb, ub, strategy=strategy, indvars=indvars, depvars=depvars)->
+            begin
+                integrand_ = NeuralPDE.build_symbolic_loss_function(nothing, indvars, depvars, phi, derivative, nothing, chain, θ, strategy, integrand = integrand)
```
KirillZubov, line 675: all actions with the parser must be performed before starting training. To call the `build_symbolic_loss`  in `integral` function means that it will be called every time during training, so it will take a very long time.

src/pinns_pde_solve.jl
```diff
@@ -793,12 +860,13 @@ function SciMLBase.symbolic_discretize(pde_system::PDESystem, discretization::Ph
     phi = discretization.phi
     derivative = discretization.derivative
     strategy = discretization.strategy
+    global integral = get_numeric_integral(strategy, pde_system.indvars, pde_system.depvars, dict_indvars, chain, derivative)
```
KirillZubov, line 863: global is bad. it will always lead to trouble. Avoid it

test/NNPDE_tests.jl
```diff
@@ -553,3 +553,35 @@ p_ = res.minimizer[end-2:end]
...
 # plot!(ts, u_predict, label = ["x(t)" "y(t)" "z(t)"])
+
+@parameters t
+@variables i(..)
+Di = Differential(t)
+Ii = Integral(t, DomainSets.ClosedInterval(0, t))
+eq = Di(i(t)) + 2*i(t) + 5*Ii(i(t)) ~ 1
+bcs = [i(0.) ~ 0.0]
+domains = [t ∈ Interval(0.0,2.0)]
+chain = FastChain(FastDense(1,15,Flux.σ),FastDense(15,1))
```
KirillZubov, line 564: it's some trouble with `FastChain` after the new release of `Flux`, just use `Chain` for a test while `FastChain` won't have been fixed.

ChrisRackauckas: what's up?

KirillZubov: `initial_params` change behavior after the update in Flux and now it works incorrectly.
```julia
chain = FastChain(FastDense(1,12,Flux.σ),FastDense(12,1))
julia> initθ = DiffEqFlux.initial_params(chain)
26-element Vector{Union{Nothing, Float32}}:
  0.44163412f0
  0.3313627f0
 -0.24543865f0
  0.16519924f0
  0.18937507f0
 -0.047894925f0
  ⋮
 -0.50703824f0
 -0.13543157f0
  0.036089323f0
  0.0010503983f0
   nothing
```

ChrisRackauckas: Can you try and narrow that down?

KirillZubov: ok, I will open the issue in DiffEqFlux as soon as I figure out what exactly is the problem with `DiffEqFlux.initial_params`.

src/pinns_pde_solve.jl
ashutosh-b-b: > in this case I needed to use integral outside the `symbolic_discretize` and `discretize`. And hence global declaration. What are other ways to do it?

test/NNPDE_tests.jl
```diff
@@ -553,3 +553,32 @@ p_ = res.minimizer[end-2:end]
...
+initθ = Float64.(DiffEqFlux.initial_params(chain))
+strategy_ = NeuralPDE.GridTraining(0.05)
+discretization = NeuralPDE.PhysicsInformedNN(chain,
+                                             strategy_;
+                                             init_params = nothing,
+                                             phi = nothing,
+                                             derivative = nothing,
+                                             )
+pde_system = PDESystem(eq,bcs,domains,[t],[i])
+prob = NeuralPDE.discretize(pde_system,discretization)
```
KirillZubov, line 574: it work in forward but
```julia
prob.f.f.loss_function(initθ)
1.1225828958232524
```
donesn't work in backward
```julia
julia> Zygote.gradient(prob.f.f.loss_function,initθ)
ERROR: MethodError: no method matching get(::Tuple{Int64, Int64}, ::Int64, ::Int64)
```
I suppose it is related to integral function.
So you need  that work in both way:
`integral = get_numeric_integral(strategy, _indvars, _depvars, dict_indvars, chain, derivative)`

```julia
integral(initθ,... )
integral_ = (θ) -> integral(θ, ...)
Zygote.gradient(integral_,initθ)
```
All symbolic things should do before calculation because `Zygote.gradient` probably can't handle all of it.

ChrisRackauckas:
Is it still WIP?

test/NNPDE_tests.jl
```diff
@@ -553,3 +553,55 @@ p_ = res.minimizer[end-2:end]
...
+
+analytic_sol_func(t) = 1/2*(exp(-t))*(sin(2*t))
+u_real  = [analytic_sol_func(t) for t in ts]
+u_predict  = [first(phi([t],res.minimizer)) for t in ts]
+@test Flux.mse(u_real, u_predict) < 0.001
+
+## Simple Integral Test
+@parameters x
+@variables u(..)
+Ix = Integral(x, DomainSets.ClosedInterval(0, x))
```
ChrisRackauckas, line 587: needs a multidimensional test

ashutosh-b-b: Yes, regarding that I wanted to ask this, for a multidimensional integral, I wanted to change definition of the symbolic integral to have a vector of symbols. For eg. I wanted it to be like this : `I = Integral([x , y, z], UnitCube())`. Should I proceed?

ChrisRackauckas: It should be done in the same way we're describing sets, i.e. `(x , y, z) \in UnitCube()`

ashutosh-b-b: I am talking about the `Integral`.

ChrisRackauckas: yes. `I = Integral(x^2 + y^2 + z^2,(x,y,z) in UnitCube())`

KirillZubov:
I think it will need to split tests files and move tests with integral to integro_differential_test.jl

test/NNPDE_tests.jl
ashutosh-b-b: We first define the `Integral` as `Integral(integrating_variable, domain)` and then pass the function. What I meant was using vector of integrating_variables instead. Is it fine? Or do I have to change it?

src/pinns_pde_solve.jl
```diff
@@ -334,19 +380,24 @@ end
...
+                                      )
     if chain isa AbstractArray
         eltypeθ = eltype(initθ[1])
     else
         eltypeθ = eltype(initθ)
     end
-
-    loss_function = parse_equation(eqs,dict_indvars,dict_depvars,chain,eltypeθ,strategy)
-    vars = :(cord, $θ, phi, derivative,u,p)
+    if integrand isa Nothing
```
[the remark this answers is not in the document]
ashutosh-b-b: So we can re use the `build_symbolic_loss_function` for getting the integral function in our inner representation.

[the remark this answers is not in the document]
KirillZubov: got it

Project.toml
```diff
@@ -45,7 +49,7 @@ DiffEqNoiseProcess = "5.1"
 DiffEqSensitivity = "6.5"
 Distributions = "0.23, 0.24, 0.25"
 DocStringExtensions = "0.8"
-Flux = "0.10.1, 0.11, 0.12"
+Flux = "0.10.1, 0.11, < 0.12.5"
```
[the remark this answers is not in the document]
KirillZubov: just press the button “resolve conflict” and merge it

src/pinns_pde_solve.jl
```diff
@@ -658,7 +709,60 @@ function get_numeric_derivative()
         end
 end
 
+function get_numeric_integral(strategy, _indvars, _depvars, chain, derivative)
+    depvars,indvars,dict_indvars,dict_depvars = get_vars(_indvars, _depvars)
+    integral =
+        (u, cord, phi, integrating_var_id, integrand_func, lb, ub, θ ;strategy=strategy, indvars=indvars, depvars=depvars, dict_indvars=dict_indvars, dict_depvars=dict_depvars)->
+            begin
+                flat_θ = if (typeof(chain) <: AbstractVector) reduce(vcat,θ) else θ end
```
KirillZubov, line 717: why do you need it? θ should already be flat

src/pinns_pde_solve.jl
```diff
@@ -658,7 +709,60 @@ function get_numeric_derivative()
...
+                if !ub_definite
+                    ub_ = ub(cord , flat_θ, phi, derivative, nothing, u, nothing)
+                else
+                    if ub isa Number
+                        ub_ = fill(ub, 1 , size(cord)[2])
+                    else
+                        ub_ = repeat(ub, 1, size(cord)[2])
+                    end
+                end
+                for i in 1:size(cord)[2]
```
KirillZubov, line 754: if `cord = ones(2,300)`
so then 
julia> size(cord)[2]
300
why we calculate it for each index?  it is very slow

ashutosh-b-b: in order to give the vectorise output. We have each index is a cord point in the data, so we need to calculate integral seperately for each point

src/pinns_pde_solve.jl
ashutosh-b-b: I checked it, it wasnt. I ll recheck

src/pinns_pde_solve.jl
ashutosh-b-b: I did try `mapslices` but it threw a `Mutating arrays not supported` error

src/pinns_pde_solve.jl
```diff
@@ -658,7 +709,60 @@ function get_numeric_derivative()
         end
 end
 
+function get_numeric_integral(strategy, _indvars, _depvars, chain, derivative)
+    depvars,indvars,dict_indvars,dict_depvars = get_vars(_indvars, _depvars)
+    integral =
+        (u, cord, phi, integrating_var_id, integrand_func, lb, ub, θ ;strategy=strategy, indvars=indvars, depvars=depvars, dict_indvars=dict_indvars, dict_depvars=dict_depvars)->
```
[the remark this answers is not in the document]
ashutosh-b-b: These are used inside the integral function, so shouldn't we pass them?

src/pinns_pde_solve.jl
```diff
@@ -793,12 +898,13 @@ function SciMLBase.symbolic_discretize(pde_system::PDESystem, discretization::Ph
     phi = discretization.phi
     derivative = discretization.derivative
     strategy = discretization.strategy
+    global integral = get_numeric_integral(strategy, pde_system.indvars, pde_system.depvars, chain, derivative)
```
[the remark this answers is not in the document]
ashutosh-b-b: Done.

src/pinns_pde_solve.jl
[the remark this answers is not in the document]
KirillZubov: we don't use these for initializing function inside `get_numeric_integral` so we don't need to pass them.

src/pinns_pde_solve.jl
KirillZubov: it is a vector we do not need an index, we need a broadcast here for fast performance (see how it works in `get_numeric_derivative`).
In any case, it works fully correctly, but slow

KirillZubov:
```julia
using BenchmarkTools

f(x) =x^2

function f_index(x)
    for i in 1:size(x)[2]
        x[:,i].^2
    end
end


julia> @benchmark f.(data) setup=(data=rand(2,1000))
BenchmarkTools.Trial: 10000 samples with 9 evaluations.
 Range (min … max):  2.525 μs …  3.818 ms  ┊ GC (min … max):  0.00% … 99.82%
 Time  (median):     3.732 μs              ┊ GC (median):     0.00%
 Time  (mean ± σ):   7.412 μs ± 85.571 μs  ┊ GC (mean ± σ):  28.22% ±  2.44%

  ▁ ▂ █▅▄▇▂ ▁                 ▃▄▅▄▄▄▃▃▃▃▂▂▂▁▁                ▂
  █▄█████████▆▅▄▅▄▅▆▅▅▅▃▅▆▅▅▅▇█████████████████▇▇▇▇▆▄▅▅▆▅▅▅▅ █
  2.52 μs      Histogram: log(frequency) by time     12.2 μs <

 Memory estimate: 15.75 KiB, allocs estimate: 1.

julia> @benchmark f_index(data) setup=(data=rand(2,1000))
BenchmarkTools.Trial: 10000 samples with 1 evaluation.
 Range (min … max):   56.758 μs … 36.600 ms  ┊ GC (min … max):  0.00% … 99.53%
 Time  (median):      78.680 μs              ┊ GC (median):     0.00%
 Time  (mean ± σ):   118.139 μs ±  1.100 ms  ┊ GC (mean ± σ):  29.35% ±  3.15%

  ▄▃ ▆█▆▆▆▄▅█▇▃▃▂▂▃▂▃▃▃▂▂▂▂▁▁▁▁▁▁▁▁▁  ▁▁                       ▂
  ████████████████████████████████████████▇▇█▇▇▇▇▇▇▆▆▆▇▆▆▄▅▄▅▅ █
  56.8 μs       Histogram: log(frequency) by time       188 μs <

 Memory estimate: 187.50 KiB, allocs estimate: 2000.
```

Project.toml
```diff
@@ -15,13 +15,15 @@ DiffEqNoiseProcess = "77a26b50-5914-5dd7-bc55-306e6241c503"
...
 DocStringExtensions = "ffbed154-4ef7-542d-bbb7-c09d3a79fcae"
+DomainSets = "5b8099bc-c8ec-5219-889f-1d9e522a28bf"
 Flux = "587475ba-b771-5e3f-ad9e-33799f191a9c"
 ForwardDiff = "f6369f11-7733-5829-9624-2563aa707210"
 GalacticOptim = "a75be94c-b780-496d-a8a9-0878b188d577"
 IfElse = "615f187c-cbe4-4ef1-ba3b-2fcf58d6d173"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 ModelingToolkit = "961ee093-0014-501f-94e3-6117800e7a78"
 Optim = "429524aa-4258-5aef-a3af-852621145aeb"
+Plots = "91a5bcdd-55d7-5caf-9e0b-520d859cae80"
```
ChrisRackauckas, line 26: why is this added?

ashutosh-b-b: Oh that got pushed by mistake. I ll remove it.

test/NNPDE_tests.jl
```diff
@@ -760,3 +760,112 @@ res  = GalacticOptim.solve(prob,BFGS(),maxiters=500)
...
+chain = Chain(Dense(2,15,Flux.σ),Dense(15,1))
+initθ = Float64.(DiffEqFlux.initial_params(chain))
+strategy_ = NeuralPDE.GridTraining(0.05)
+discretization = NeuralPDE.PhysicsInformedNN(chain,
+                                             strategy_;
+                                             init_params = nothing,
+                                             phi = nothing,
+                                             derivative = nothing,
+                                             )
+pde_system = PDESystem(eq,bcs,domains,[x,y],[u])
```
KirillZubov, line 863: @named pde_system = PDESystem(eq,bcs,domains,[x,y],[u])

test/NNPDE_tests.jl
```diff
@@ -760,3 +760,112 @@ res  = GalacticOptim.solve(prob,BFGS(),maxiters=500)
...
```
KirillZubov, line 836: @named

test/NNPDE_tests.jl
```diff
@@ -760,3 +760,112 @@ res  = GalacticOptim.solve(prob,BFGS(),maxiters=500)
...
+chain = Chain(Dense(1,15,Flux.σ),Dense(15,1))
+initθ = Float64.(DiffEqFlux.initial_params(chain))
+strategy_ = NeuralPDE.GridTraining(0.05)
+discretization = NeuralPDE.PhysicsInformedNN(chain,
+                                             strategy_;
+                                             init_params = nothing,
+                                             phi = nothing,
+                                             derivative = nothing,
+                                             )
+pde_system = PDESystem(eq,bcs,domains,[x],[u])
```
KirillZubov, line 809: @named

docs/src/pinn/integro_diff.md
```diff
@@ -0,0 +1,71 @@
...
+We use the grid training strategy and then use the `PhysicsInformedNN` interface to define the optimization problem.
+```julia
+strategy_ = NeuralPDE.GridTraining(0.05)
+discretization = NeuralPDE.PhysicsInformedNN(chain,
+                                             strategy_;
+                                             init_params = nothing,
+                                             phi = nothing,
+                                             derivative = nothing,
+                                             )
+pde_system = PDESystem(eq,bcs,domains,[t],[i])
```
KirillZubov, line 45: `@named`

test/NNPDE_tests.jl
```diff
@@ -760,3 +760,112 @@ res  = GalacticOptim.solve(prob,BFGS(),maxiters=500)
...
+chain = Chain(Dense(1,15,Flux.σ),Dense(15,1))
+initθ = Float64.(DiffEqFlux.initial_params(chain))
+strategy_ = NeuralPDE.GridTraining(0.05)
+discretization = NeuralPDE.PhysicsInformedNN(chain,
+                                             strategy_;
+                                             init_params = nothing,
+                                             phi = nothing,
+                                             derivative = nothing,
+                                             )
+pde_system = PDESystem(eq,bcs,domains,[t],[i])
```
KirillZubov, line 782: and here `@named`

docs/src/pinn/integro_diff.md
```diff
@@ -0,0 +1,71 @@
...
+prob = NeuralPDE.discretize(pde_system,discretization)
```
KirillZubov, line 46: add
```julia
cb = function (p,l)
    println("Current loss is: $l")
    return false
end

```

docs/src/pinn/integro_diff.md
```diff
@@ -0,0 +1,71 @@
...
+```julia
+analytic_sol_func(t) = 1/2*(exp(-t))*(sin(2*t))
+u_real  = [analytic_sol_func(t) for t in ts]
+```
+Plotting the final solution and analytical solution
+```julia
+using Plots
+plot(u_real, ts, label = "Analytical Solution")
+plot!(u_predict, ts, label = "PINN Solution")
+```
```
KirillZubov, line 71: IDE

KirillZubov: `IDE`

docs/src/pinn/integro_diff.md
```diff
@@ -0,0 +1,71 @@
+# Integro Differential Equations
```
KirillZubov, line 1: I don't think it needs to describe how to define `Differential` and etc, this is already clear and is described in the rest of the documentation. Here it needs to describe in more detail how to define the integral that it was clear how to use it.
 
My draft suggestion: 

# Integro Differential Equations

The integral of function u(x),

```math
\int_{0}^{t}u(x)dx
```

where x is variable of integral and t is variable of integro differential equation,

is defined as

```julia
using ModelingToolkit
@parameters t
@variables i(..)
Ii = Symbolics.Integral(t, t in DomainSets.ClosedInterval(0, t))
```

In multidimensional case,

```julia
Ix = Integral([x,y], (x,y) in DomainSets.UnitSquare())
```

///describe how it work  ///

```julia
Ix = Integral([x,y], (x,y) in DomainSets.ProductDomain(ClosedInterval(0 ,1), ClosedInterval(0 ,x)))
```


## 1-dimensional example

Lets take an example of an integro differential equation:
```math
\frac{∂}{∂x} u(x)  + 2u(x) + 5 \int_{0}^{x}u(t)dt = 1 for x \geq 0
```
and boundary condition
```math
u(0) = 0
```

```julia
using NeuralPDE, Flux, ModelingToolkit, GalacticOptim, Optim, DiffEqFlux, DomainSets

@parameters t
@variables i(..)
Di = Differential(t)
Ii = Integral(t, t in DomainSets.ClosedInterval(0, t))
eq = Di(i(t)) + 2*i(t) + 5*Ii(i(t)) ~ 1
bcs = [i(0.) ~ 0.0]
domains = [t ∈ Interval(0.0,2.0)]
chain = Chain(Dense(1,15,Flux.σ),Dense(15,1))

strategy_ = GridTraining(0.05)
discretization = PhysicsInformedNN(chain,
                                   strategy_;
                                   init_params = nothing,
                                   phi = nothing,
                                   derivative = nothing)
pde_system = PDESystem(eq,bcs,domains,[t],[i])
cb = function (p,l)
    println("Current loss is: $l")
    return false
end
res = GalacticOptim.solve(prob, BFGS(); cb = cb, maxiters=100)
```

Plotting the final solution and analytical solution

```julia
ts = [infimum(d.domain):0.01:supremum(d.domain) for d in domains][1]
phi = discretization.phi
u_predict  = [first(phi([t],res.minimizer)) for t in ts]

analytic_sol_func(t) = 1/2*(exp(-t))*(sin(2*t))
u_real  = [analytic_sol_func(t) for t in ts]
using Plots
plot(u_real, ts, label = "Analytical Solution")
plot!(u_predict, ts, label = "PINN Solution")
```

IDE

test/NNPDE_tests.jl
```diff
@@ -772,3 +774,136 @@ res  = GalacticOptim.solve(prob,BFGS(),maxiters=500)
...
+# p1 = plot(xs,ys,u_real,linetype=:contourf,label = "analytic")
+# p2 = plot(xs,ys,u_predict,linetype=:contourf,label = "predict")
+# p3 = plot(xs,ys,error_,linetype=:contourf,label = "error")
+# plot(p1,p2,p3)
+
+@parameters x,y
+@variables u(..)
+Dx = Differential(x)
+Dy = Differential(y)
+Ix = Integral([x,y], (x,y) in DomainSets.ProductDomain(UnitInterval(),ClosedInterval(0 ,x)))
```
ChrisRackauckas, line 881: why is this two outputs? The equations below don't have the right size.

test/NNPDE_tests.jl
```diff
@@ -772,3 +774,136 @@ res  = GalacticOptim.solve(prob,BFGS(),maxiters=500)
...
+#simple multidimensitonal integral test
+println("simple multidimensitonal integral test")
+
+@parameters x,y
+@variables u(..)
+Dx = Differential(x)
+Dy = Differential(y)
+Ix = Integral([x,y], (x,y) in DomainSets.UnitSquare())
+eq = Ix(u(x,y)) ~ 1/3
+bcs = [u(0., 0.) ~ 1, Dx(u(x,y)) ~ -2*x , Dy(u(x ,y)) ~ -2*y ]
```
ChrisRackauckas, line 849: same sizing issue.

ChrisRackauckas (approved):

ChrisRackauckas:
@sdwfrost

Issue #333, opened 2021-06-24
Neural adapter


KirillZubov:
- [x] first step, simple retrain 
```julia
chain1 = FastChain(FastDense(2,16,Flux.σ),FastDense(16,1))
result1 = GalacticOptim.solve(prob1,...)
chain2 = FastChain(FastDense(2,16,Flux.σ),
                   FastDense(16,32,Flux.σ),
                   FastDense(32,64,Flux.σ),
                   FastDense(64,1))
retrain_weight_for_chain2 = neurural_adapter(chain1,result,chain2,...)
```

- [x] time decomposition [chain1,chain2,...,chain_n] -> total_time_domain_chain

PR #336, opened 2021-06-30
Neural adapter

KirillZubov:
https://github.com/SciML/NeuralPDE.jl/issues/333


src/pinns_pde_solve.jl
```diff
@@ -1,5 +1,5 @@
 import Base.Broadcast
-# Base.Broadcast.dottable(x::Function) = true
+Base.Broadcast.dottable(x::Function) = true
```
KirillZubov, line 2: to use @register, I needed this again

KirillZubov: https://github.com/SciML/NeuralPDE.jl/issues/353

ChrisRackauckas: `ModelingToolkit.@register`?

docs/src/pinn/neural_adapter.md
```diff
@@ -0,0 +1,252 @@
+## Neural Adapter
+
+Using the example of 2D Poisson equation, it is shown how, using method neural_adapter, to retrain the prediction of one neural network to another.
```
ChrisRackauckas, line 3:
```suggestion
## Transfer Learning with Neural Adapter

Using the example of 2D Poisson equation, it is shown how, using method neural_adapter, to retrain the prediction of one neural network to another.
```

ChrisRackauckas, line 3: This needs more of an introduction. What is transfer learning? Why would you want to use it? What are the pros? The cons?

docs/src/pinn/neural_adapter.md
```diff
@@ -0,0 +1,252 @@
...
+
+initθ2 =Float64.(DiffEqFlux.initial_params(chain2))
+
+function loss(cord,θ)
+    chain2(cord,θ) .- phi(cord,res.minimizer)
+end
+
+strategy = NeuralPDE.GridTraining(0.02)
+
+prob_ = NeuralPDE.neural_adapter(loss, initθ2, pde_system, strategy)
```
ChrisRackauckas, line 60: why is this not using the params from `res`?

src/pinns_pde_solve.jl
KirillZubov: yes

ChrisRackauckas: Can you show a minimal example of what it's needed for now?

KirillZubov:
```julia
@parameters θ
@variables u(..)
Dθ = Differential(θ)
fx(θ) = θ
@register fx(θ)

eq = Dθ(u(θ)) ~ 2*fx(θ)

expr =ModelingToolkit.toexpr(eq.rhs)
```


```julia
julia> dump(expr)
Expr
  head: Symbol call
  args: Array{Any}((3,))
    1: * (function of type typeof(*))
    2: Int64 2
    3: Expr
      head: Symbol call
      args: Array{Any}((2,))
        1: fx (function of type typeof(fx))
        2: Symbol θ
```
`fx` is a function inside `expr` if we convert it in the symbol(in the main branch), 
@register function is not working and can't be called in an inner representation 
as a function because it converts to a symbol.
Convert in symbol was done for avoid broadcast piracy `Base.Broadcast.dottable(x::Function) = true`

https://github.com/SciML/NeuralPDE.jl/blob/d42dbd54072567096407e990c8685a63ea7d1b13/src/pinns_pde_solve.jl#L188-L190

here broadcast avoid function
```julia
Base.Broadcast.dottable(x::Function) = false

julia> Broadcast.__dot__(expr)
:((*)(2, (fx)(θ)))
```
here functions aslo broadcasted
```julia
Base.Broadcast.dottable(x::Function) = true

Broadcast.__dot__(expr)
:((*).(2, (fx).(θ)))
```
so if we want to use ModelingToolKit@register and broadcast work correct is need `Base.Broadcast.dottable(x::Function) = true`

and macro @. only works with direct insert.
Maybe some trick is needed here which I don't know

```julia
julia>  expr
:((*)(2, (fx)(θ)))

julia> @. expr
:((*)(2, (fx)(θ)))

julia> @. :((*)(2, (fx)(θ)))
:((*).(2, fx.(θ)))
```

docs/src/pinn/neural_adapter.md
KirillZubov: it is using in 54th line 
```julia
function loss(cord,θ)
    chain2(cord,θ) .- phi(cord,res.minimizer)
end
```

KirillZubov:
@zoemcc could you please review this PR

ChrisRackauckas (approved):

KirillZubov:
🎉

Issue #337, opened 2021-07-02
 Broadcast piracy


KirillZubov:
Get rid  of override dottable for PINNs  `Base.Broadcast.dottable(x::Function) = true`
https://github.com/SciML/NeuralPDE.jl/blob/master/src/pinns_pde_solve.jl#L2

PR #338, opened 2021-07-02
fix broadcast piracy

KirillZubov:
https://github.com/SciML/NeuralPDE.jl/issues/337

### DIFF
--- a/src/pinns_pde_solve.jl
+++ b/src/pinns_pde_solve.jl
@@ -195,9 +195,9 @@ function _transform_expression(ex,dict_indvars,dict_depvars,chain,eltypeθ,strat
                 indvars = _args[2:end]
                 cord = :cord
                 ex.args = if !(typeof(chain) <: AbstractVector)
-                    [:u, cord, :($θ), :phi]
+                    [:($u), cord, :($θ), :phi]
                 else
-                    [:u, cord, Symbol(:($θ),num_depvar), Symbol(:phi,num_depvar)]
+                    [:($u), cord, Symbol(:($θ),num_depvar), Symbol(:phi,num_depvar)]
                 end
                 break
             elseif e isa ModelingToolkit.Differential
@@ -217,9 +217,9 @@ function _transform_expression(ex,dict_indvars,dict_depvars,chain,eltypeθ,strat
                 undv = [dict_indvars[d_p] for d_p  in derivative_variables]
                 εs_dnv = [εs[d] for d in undv]
                 ex.args = if !(typeof(chain) <: AbstractVector)
-                    [:derivative, :phi, :u, cord, εs_dnv, order, :($θ)]
+                    [:($derivative), :phi, :u, cord, εs_dnv, order, :($θ)]
                 else
-                    [:derivative, Symbol(:phi,num_depvar), :u, cord, εs_dnv, order, Symbol(:($θ),num_depvar)]
+                    [:($derivative), Symbol(:phi,num_depvar), :u, cord, εs_dnv, order, Symbol(:($θ),num_depvar)]
                 end
                 break
             end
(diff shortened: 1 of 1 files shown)


src/pinns_pde_solve.jl
```diff
@@ -185,6 +185,9 @@ where
 function _transform_expression(ex,dict_indvars,dict_depvars,chain,eltypeθ,strategy)
     _args = ex.args
     for (i,e) in enumerate(_args)
+        if e in [+,-,*,/,^,inv]
```
ChrisRackauckas, line 188: why is this necessary?

ashutosh-b-b: Inorder to disable `__dot__` for specific functions.

ChrisRackauckas: but why these functions, and why here? I don't get it.

KirillZubov: sorry for the long answer, 
the problem is that `toexpr` transforms not as symbols but as functions of all arithmetic operators. But in Broadcast.jl the function is used as a marker for not dotting (which is generally weird)
https://github.com/JuliaLang/julia/blob/4270d3b44e53e9fa68dcdbb6cb416d232615db4c/base/broadcast.jl#L1295


```julia
@parameters x y
@variables u(..)
eq =  u(x)+ u(x)*x
expr = toexpr(eq)
julia> dump(expr)
Expr
  head: Symbol call
  args: Array{Any}((3,))
    1: + (function of type typeof(+))
    ...

Base.Broadcast.dottable(x::Function) = false
julia> Broadcast.__dot__(expr)
:((+)(u.(x), (*)(x, u.(x))))

Base.Broadcast.dottable(x::Function) = true
julia> Broadcast.__dot__(expr)
:((+).(u.(x), (*).(x, u.(x))))
```

KirillZubov: so to avoid piracy  and broadcasted full expression with arithmetic in need convert them in symbol

KirillZubov: in general, this is a crutch, using functions inside an expression to avoid dotting. 
https://github.com/JuliaLang/julia/blob/4270d3b44e53e9fa68dcdbb6cb416d232615db4c/base/broadcast.jl#L1295

KirillZubov: leading to the fact that we have to exclude functions in order to correct broadcast expr, which is kind of bad

ChrisRackauckas:
```suggestion
        if e isa Function
```

?

KirillZubov: it doesn't work,   `Differential` is also function( .
```julia
Dθ = Differential(θ)
eq = Dθ(u(θ))
expr = toexpr(eq)
julia> dump(expr)
Expr
  head: Symbol call
  args: Array{Any}((2,))
    1: Differential(θ) (function of type Differential)
...
```

KirillZubov: to avoid all this, in the end, we need our own `toexpr` in NeuralPDE

ChrisRackauckas: `if e isa Function && e !== Differential`

ChrisRackauckas: >to avoid all this, in the end, we need our own toexpr in NeuralPDE

That seems unnecessary.

ChrisRackauckas: Why would it require an entire new `toexpr` instead of just omitting one function?

KirillZubov: Since `ModelingToolKit.toexpr` does not guarantee the same behavior, many times it was necessary to rewrite the parser after updating ModelingToolKit.jl

but it’s not urgent, as long as it works)

ChrisRackauckas (approved):

PR #361, opened 2021-08-04
[wip] Coordinate transform example

vpuri3:
To do:
1. fix latex rendering
2. add image


docs/src/pinn/poisson_annulus.md
```diff
@@ -0,0 +1,228 @@
...
+## Coordinate Transform and Differential Operators
+
+We represent the field vairable, ``u``, and physical coordinates ``x,y`` in terms of reference variables ``r,\theta``:
+
+```math
+u(r,\theta), \, x(r,\theta), \, y(r,\theta)
+```
+
+```julia
+@paameters r,\theta
```
KirillZubov, line 158: @parameters

vpuri3: fixed :D

KirillZubov:
very cool example

vpuri3:
Thanks. Could you help me with the writeup? LaTeX isn't compiling in the markdown, and I don't know where to upload an image of the solution.

KirillZubov:
> Thanks. Could you help me with the writeup? LaTeX isn't compiling in the markdown, and I don't know where to upload an image of the solution.
documenter.jl compiles in LaTeX after merge. so it's fine. the markdown don't do it.

KirillZubov:
the same in `main` branch https://github.com/SciML/NeuralPDE.jl/blob/master/docs/src/pinn/2D.md but in docs it is already compiled

vpuri3:
Thanks. Where can I upload the image?
poisson_annulus

vpuri3:
Nevermind. Dragging and dropping into the markdown worked!

vpuri3:
@ChrisRackauckas , I think this commit is ready to merge.

ChrisRackauckas:
Yeah, just put images into the github and then use `` to put it into the docs.

docs/src/pinn/poisson_annulus.md
```diff
@@ -0,0 +1,228 @@
+# Poisson Equation
+
+In this example, we solve the Poisson equation on an annulus to demostrate how NeuralPDE can solve differential equations in complex geometries.
```
ChrisRackauckas, line 3: using coordinate maps, etc. This should get a bit more explanation at the beginning, with a nice source that goes into more depth, etc.

ChrisRackauckas:
We should really find a good way to do such a coordinate transformation automatically symbolically.

PR #366, opened 2021-08-09
ModelingToolkit" to "6" update

KirillZubov:


src/pinns_pde_solve.jl
```diff
@@ -460,8 +460,8 @@ function build_loss_function(eqs,indvars,depvars,
 end
 
 function get_vars(indvars_, depvars_)
-    depvars = [nameof(value(d)) for d in depvars_]
-    indvars = [nameof(value(i)) for i in indvars_]
+    depvars = [nameof(d.f) for d in depvars_]
+    indvars = [nameof(i.val) for i in indvars_]
```
ChrisRackauckas, line 464: it should be accessed using the accessor functions, not through `.`

KirillZubov: ok

Issue #390, opened 2021-09-08
Broadcast integral calculation and increase performance


KirillZubov:
the current calculation of the integral is in for each index in the loop, which is very slow.

https://github.com/SciML/NeuralPDE.jl/blob/2442835ea24fd2e2dbb4663c8d05e69296ac4a50/src/pinns_pde_solve.jl#L870-L873
Need implementation with broadcast for the vector. This will greatly reduce the execution time. 
and will also give the ability to use the GPU for calculation integral.

https://github.com/SciML/NeuralPDE.jl/pull/330#discussion_r686313128

killah-t-cell:
Do you think this is a fast issue? Adding a integral makes the whole thing super slow right now.

how do you think we should implement this?

killah-t-cell:
I played with this a little and actually solved it for 1D. All you have to do is do `return integration_.(cord, lb_, ub_, Ref(flat_θ))` instead of that loop. And then `vcat(...)` a couple of things inside of `integration_`. When we get to multi dimensions though this becomes a bit harder. Currently, a multidimensional integration_ gets 

```julia
cord[:, i] = Float32[0.8, 1.0]
ub__ = [1.0, 1.0]
lb__ = [0.0, 0.0]
```

as arguments. But a broadcast `integration_.(cord, lb_, ub_, Ref(flat_θ))` gives only one dimensional inputs.
```
cord = 0.4f0
lb = 0.0
ub = 1.0
```

I have no idea how to do multidimensional broadcasting. LMK if you have any ideas here @KirillZubov

KirillZubov:
- [ ] minimize the number of unnecessary operations that are called during optimization at each iteration

The next thing that will increase the performance of calculating the integral is to take out all the lines that can be executed once before training. Instead of calling it on each training iteration.

https://github.com/SciML/NeuralPDE.jl/blob/8996411723544bc65e5ee76ff755ee0209b16576/src/pinns_pde_solve.jl#L840-L868

for example:  `typeof(chain)` is static so it can be called one time before
` flat_θ = if (typeof(chain) <: AbstractVector) reduce(vcat,θ) else θ end`

killah-t-cell:
We could just place that line before the map like this, right?

```julia
flat_θ = if (typeof(chain) <: AbstractVector) reduce(vcat,θ) else θ end
[...]

PR #393, opened 2021-09-11
Increase integral performance with map

killah-t-cell:
https://github.com/SciML/NeuralPDE.jl/issues/390 I changed the loop into a map to improve integral performance. Nothing else is different since the map gives the same inputs to `integration_` (but hopefully it does so a bit more efficiently). If you think there are other quick wins here, let me know. Happy to help.

I will debug this a bit but wanted to open the PR now.

### DIFF
--- a/src/pinns_pde_solve.jl
+++ b/src/pinns_pde_solve.jl
@@ -867,7 +867,7 @@ function get_numeric_integral(strategy, _indvars, _depvars, chain, derivative)
                     end
                 end
 
-                integration_arr = map((cord__,lb__,ub__) -> integration_(cord__,lb__,ub__,flat_θ),  eachcol(cord__), eachcol(lb_), eachcol(ub_))
+                integration_arr = map((cord__,lb__,ub__) -> integration_(cord__,lb__,ub__,flat_θ),  eachcol(cord), eachcol(lb_), eachcol(ub_))
                 return reshape(integration_arr, :, length(integration_arr)) 
             end
 end
--- a/src/pinns_pde_solve.jl
+++ b/src/pinns_pde_solve.jl
@@ -867,7 +867,7 @@ function get_numeric_integral(strategy, _indvars, _depvars, chain, derivative)
                     end
                 end
 
-                integration_arr = map((cord__,lb__,ub__) -> integration_(cord__,lb__,ub__,flat_θ), @view cord[:,1], eachcol(lb_), eachcol(ub_))
+                integration_arr = map((cord__,lb__,ub__) -> integration_(cord__,lb__,ub__,flat_θ),  eachcol(cord__), eachcol(lb_), eachcol(ub_))
                 return reshape(integration_arr, :, length(integration_arr)) 
             end
 end
--- a/src/pinns_pde_solve.jl
+++ b/src/pinns_pde_solve.jl
@@ -839,7 +839,7 @@ function get_numeric_integral(strategy, _indvars, _depvars, chain, derivative)
             begin
                 flat_θ = if (typeof(chain) <: AbstractVector) reduce(vcat,θ) else θ end
                 function integration_(cord, lb, ub, flat_θ)
-                    cord_ = cord #vcat maybe
+                    cord_ = cord
                     function integrand_(x , p)
                         @Zygote.ignore @views(cord_[integrating_var_id]) .= x
                         return integrand_func(cord_, p, phi, derivative, nothing, u, nothing)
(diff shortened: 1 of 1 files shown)


src/pinns_pde_solve.jl
```diff
@@ -867,12 +867,7 @@ function get_numeric_integral(strategy, _indvars, _depvars, chain, derivative)
                     end
                 end
 
-                for i in 1:size(cord)[2]
-                    ub__ = @Zygote.ignore getindex(ub_, :,  i)
-                    lb__ = @Zygote.ignore getindex(lb_, :,  i)
-                    integration_arr = hcat(integration_arr ,integration_(cord[:, i], lb__, ub__, flat_θ))
-                end
-                return integration_arr
+                return map((cord__,lb__,ub__) -> integration_(cord__,lb__,ub__,flat_θ), eachcol(cord), eachcol(lb_), eachcol(ub_))
```
ChrisRackauckas, line 870: eachcol with views would be good. Does it automatically use views? How does the profile look?

killah-t-cell: How do you see views being used here? 

Another question: I might need to add another line here to convert my `map((cord__,lb__,ub__) -> integration_(cord__,lb__,ub__,flat_θ), eachcol(cord), eachcol(lb_), eachcol(ub_))` which returns an X element Vector into a 1 by X Matrix. This feels silly but the convergence tests break if I don't do this. What is the most efficient way to do this conversion? 

For now I am doing `reshape(integration_arr, :, length(integration_arr)) ` but this doesn't feel right

killah-t-cell: @ChrisRackauckas (forgot to tag)

ChrisRackauckas: the column iterators, are they slicing per column or viewing?

src/pinns_pde_solve.jl
```diff
@@ -866,8 +866,7 @@ function get_numeric_integral(strategy, _indvars, _depvars, chain, derivative)
                         @Zygote.ignore ub_[i, :] = u_(cord , flat_θ, phi, derivative, nothing, u, nothing)
                     end
                 end
-
-                integration_arr = map((cord__,lb__,ub__) -> integration_(cord__,lb__,ub__,flat_θ),  eachcol(cord), eachcol(lb_), eachcol(ub_))
+                integration_arr = map((cord__,lb__,ub__) -> integration_(cord__,lb__,ub__,flat_θ), @views(eachcol(cord)), @views(eachcol(lb_)), @views(eachcol(ub_)))
```
ChrisRackauckas, line 869: no, it would be internal to the iterator.

killah-t-cell: Sorry, I am confused. Maybe I have a knowledge gap here (it is my first optimization PR). But isn't each time I call eachcol() already a `view`? So they take a slice per viewing? Or do you mean I should do something to the `integration_` function? 

MWE
```
matrix = Float32[0.1 0.2 0.3 0.4 0.5 0.6 0.7 0.8 0.9 1.0 0.1 0.2 0.3 0.4 0.5 0.6 0.7 0.8 0.9 1.0 0.1 0.2 0.3 0.4 0.5 0.6 0.7 0.8 0.9 1.0 0.1 0.2 0.3 0.4 0.5 0.6 0.7 0.8 0.9 1.0 0.1 0.2 0.3 0.4 0.5 0.6 0.7 0.8 0.9 1.0 0.1 0.2 0.3 0.4 0.5 0.6 0.7 0.8 0.9 1.0 0.1 0.2 0.3 0.4 0.5 0.6 0.7 0.8 0.9 1.0 0.1 0.2 0.3 0.4 0.5 0.6 0.7 0.8 0.9 1.0 0.1 0.2 0.3 0.4 0.5 0.6 0.7 0.8 0.9 1.0 0.1 0.2 0.3 0.4 0.5 0.6 0.7 0.8 0.9 1.0; 0.1 0.1 0.1 0.1 0.1 0.1 0.1 0.1 0.1 0.1 0.2 0.2 0.2 0.2 0.2 0.2 0.2 0.2 0.2 0.2 0.3 0.3 0.3 0.3 0.3 0.3 0.3 0.3 0.3 0.3 0.4 0.4 0.4 0.4 0.4 0.4 0.4 0.4 0.4 0.4 0.5 0.5 0.5 0.5 0.5 0.5 0.5 0.5 0.5 0.5 0.6 0.6 0.6 0.6 0.6 0.6 0.6 0.6 0.6 0.6 0.7 0.7 0.7 0.7 0.7 0.7 0.7 0.7 0.7 0.7 0.8 0.8 0.8 0.8 0.8 0.8 0.8 0.8 0.8 0.8 0.9 0.9 0.9 0.9 0.9 0.9 0.9 0.9 0.9 0.9 1.0 1.0 1.0 1.0 1.0 1.0 1.0 1.0 1.0 1.0]

function f(cord)
    @show typeof(cord)
    return
end

map(f, eachcol(cord))

# typeof(cord) = SubArray{Float32, 1, Matrix{Float32}, Tuple{Base.Slice{Base.OneTo{Int64}}, Int64}, true}
```

ChrisRackauckas: That's what I was asking. If they are already views internally that's fine.

killah-t-cell: Ohh ok! I misunderstood. Well if you are happy with this we can merge.

KirillZubov (approved):

KirillZubov:
All looks good for me, it's ready for the merge.

PR #401, opened 2021-09-20
WIP: Integrate in infinite interval

killah-t-cell:
We forgot that we need to also change the lb and ub in `_transform_expression` so it matches the transformation done by Quadrature.jl – namely changing the ub and lb from Inf to a real number. I just took the logic from Quadrature and added it to the `_transform_expression` equation and it worked.

Is it worth writing a convergence test for this?

### DIFF
--- a/src/pinns_pde_solve.jl
+++ b/src/pinns_pde_solve.jl
@@ -332,12 +332,6 @@ function _transform_expression(ex,indvars,depvars,dict_indvars,dict_depvars,dict
                     end
                 end
 
-                #forget the logic here for now
-                lb_ = [-1.0]
-                ub_ = [1.0]
-
-                @show lb_
-                @show ub_
                 integrand_func = @RuntimeGeneratedFunction(integrand)
                 ex.args = [:($(Expr(:$, :integral))), :u, Symbol(:cord, num_depvar), :phi, integrating_var_id, integrand_func, lb_, ub_,  :($θ)]
                 break
--- a/src/pinns_pde_solve.jl
+++ b/src/pinns_pde_solve.jl
@@ -300,6 +300,19 @@ function _transform_expression(ex,indvars,depvars,dict_indvars,dict_depvars,dict
                 ub = toexpr.(ub)
                 ub_ = []
                 lb_ = []
+
+                if -Inf in lb || Inf in ub
+                    lbb = lb .== -Inf
+                    ubb = ub .== Inf
+                    _none = .!lbb .& .!ubb
+                    _inf = lbb .& ubb
+                    _semiup = .!lbb .& ubb
+                    _semilw = lbb  .& .!ubb
+                
+                    lb = 0.00.*_semiup + -1.00.*_inf + -1.00.*_semilw +  _none.*lb
+                    ub = 1.00.*_semiup + 1.00.*_inf  + 0.00.*_semilw  + _none.*ub
+                end
+
                 for l in lb
                     if l isa Number
                         push!(lb_, l)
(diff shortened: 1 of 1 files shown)


src/pinns_pde_solve.jl
```diff
@@ -300,6 +300,19 @@ function _transform_expression(ex,indvars,depvars,dict_indvars,dict_depvars,dict
...
+                if -Inf in lb || Inf in ub
+                    lbb = lb .== -Inf
+                    ubb = ub .== Inf
+                    _none = .!lbb .& .!ubb
+                    _inf = lbb .& ubb
+                    _semiup = .!lbb .& ubb
+                    _semilw = lbb  .& .!ubb
+                
+                    lb = 0.00.*_semiup + -1.00.*_inf + -1.00.*_semilw +  _none.*lb
+                    ub = 1.00.*_semiup + 1.00.*_inf  + 0.00.*_semilw  + _none.*ub
```
ChrisRackauckas, line 313: You can't just change the bounds, you also have to change the integrand.

killah-t-cell: Oh so I should do the transformation
`function v_inf(t)
	return t ./ (1 .- t.^2)
end
...
`
 that Quadrature.jl does in `_transform_expression` as well?

ChrisRackauckas: yes, but why isn't Quadrature.jl doing this automatically? It should whenever it sees infinite bounds?

killah-t-cell: It does but I think it might have to do with the order of how we call this. When we get to `get_numeric_integral`, we first call 

```julia
lb_ = zeros(size(lb)[1], size(cord)[2])
ub_ = zeros(size(ub)[1], size(cord)[2])
for (i, l) in enumerate(lb)
    if l isa Number
        @Zygote.ignore lb_[i, :] = fill(l, 1, size(cord)[2])
    else
        @Zygote.ignore lb_[i, :] = l(cord , θ, phi, derivative, nothing, u, nothing)
    end
end
...
```

At that point the bounds are still infinite. Only after this do we call `integration_(cord__,lb__,ub__,θ)` which transforms the bounds. But then the cords are built and GalacticOptim complains. That is why I thought transforming the bounds sooner would do the trick. Everything else is handled.

I did not look too deeply into this so I might just be wrong.

killah-t-cell: This is where Quadrature.jl handles it https://github.com/SciML/Quadrature.jl/blob/f913f39d9766a5d15f3829c51155ab3b655269d3/src/Quadrature.jl#L152

As you can see we do call `DiffEqBase.solve(prob::QuadratureProblem...` so it should just do it

killah-t-cell: @ChrisRackauckas on second thought this is wrong. If I change the bounds before solve is called it doesn’t recognize that it is an infinite bound and doesn’t adapt. It needs to receive infinite bounds but the rest of numeric_integration needs to work with the non-infinite bounds if I understand this correctly.

I could change the bounds in get_numeric_integration and pass the infinite bounds to solve but I am not sure if that is right. Do you have any intuition around at which level I should fix this? Would appreciate some guidance :)

ChrisRackauckas: Pass the infinite bounds to the quadrature solve and 🤞

killah-t-cell: it didn't do the trick 😭. Apparently the error `LoadError: MethodError: no method matching (::Zygote.var"#pairs_namedtuple#351"{(), NamedTuple{(), Tuple{}}})(::Tuple{})` shows up when integration_ calls QuadratureProblem with infinite bounds. The weird thing is that the solve actually handles the transformation and even outputs all solutions before failing. 

Screenshot 2021-09-22 at 08 10 33

killah-t-cell:
https://github.com/SciML/Quadrature.jl/issues/73 continued work on this here

Issue #406, opened 2021-09-23
Test on a simple integrodifferential PDE


dapias:
One of the simplest integrodifferential PDEs comes from the so-called [Bouchaud trap model](https://arxiv.org/pdf/cond-mat/9601012.pdf). The equation is

![CodeCogsEqn (2)](https://user-images.githubusercontent.com/5333281/134465182-a9d7eaa6-8fbf-4daa-b12f-f5c442a7b64a.gif)

P(E,t) is the probability of finding the system at a trap with energy E and at time t, and beta is a parameter related to the inverse of temperature. The equation has a solution in the Laplace domain that can be compared with a numerical solution. 

So far, in the [documentation](https://neuralpde.sciml.ai/stable/pinn/integro_diff/#Integro-Differential-Equations), there is only one example of an ordinary integrodifferential equation (ordinary in contrast to partial) that cannot be enough to cover more complex cases. I think we can use this model as a harder validation problem.

So far, what I have tried is (also posted in [stackoverflow](https://stackoverflow.com/questions/69290327/integrodifferential-pde-with-neuralpdes-julia) ):
```julia
using NeuralPDE, Flux, ModelingToolkit, GalacticOptim, Optim, DiffEqFlux, DomainSets
import ModelingToolkit: Interval, infimum, supremum

@parameters t E
@variables i(..)

Dt = Differential(t)
DE = Differential(E)
IE = Integral(E in DomainSets.OpenInterval(0,Inf))

beta = 2.0;
eq = Dt(i(t, E)) + exp(-beta*E)*i(t, E) - IE(i(t,E)*exp(-beta*E))*exp(-E) ~ 0
bcs = [i(0., E) ~ exp(-E)]
domains = [t ∈ Interval(0.0,100.0), E ∈ Interval(0.0, 10.0)]

chain = Chain(Dense(1,15,Flux.σ),Dense(15,1))
initθ = Float64.(DiffEqFlux.initial_params(chain));

strategy_ = GridTraining(0.05)
discretization = PhysicsInformedNN(chain,
                                   strategy_;
                                   init_params = nothing,
                                   phi = nothing,
                                   derivative = nothing)

@named pde_system = PDESystem(eq,bcs,domains,[t, E],[i(t, E)])

prob = NeuralPDE.discretize(pde_system,discretization)

```

This piece of code leads to the error: 
> KeyError: key * not found

[...]

PR #409, opened 2021-10-01
Support for compound integrals

killah-t-cell:
Some compound integrals are working. Others show different errors ranging from problems with the inner representation to ChainRulesCore errors. Will investigate and try to fix it.

### DIFF
--- a/src/NeuralPDE.jl
+++ b/src/NeuralPDE.jl
@@ -185,7 +185,7 @@ export NNODE, TerminalPDEProblem, NNPDEHan, NNPDENS, NNRODE,
        GridTraining, StochasticTraining, QuadratureTraining, QuasiRandomTraining
        build_loss_function, get_loss_function,
        generate_training_sets, get_variables, get_argument, get_bounds
-       get_phi, get_numeric_derivative,
+       get_phi, get_numeric_derivative, get_numeric_integral
        build_symbolic_equation, build_symbolic_loss_function, symbolic_discretize
 
 end # module
--- a/src/pinns_pde_solve.jl
+++ b/src/pinns_pde_solve.jl
@@ -325,28 +325,29 @@ function _transform_expression(ex,indvars,depvars,dict_indvars,dict_depvars,dict
                     if l isa Number
                         push!(lb_, l)
                     else
-                        l = NeuralPDE.build_symbolic_loss_function(nothing, indvars,depvars,
+
+                        l_expr = NeuralPDE.build_symbolic_loss_function(nothing, indvars,depvars,
                                                                    dict_indvars,dict_depvars,
                                                                    dict_depvar_input, phi, derivative_,
-                                                                   nothing, chain, θ, strategy,
+                                                                   nothing, chain, initθ, strategy,
                                                                    integrand = l, integrating_depvars=integrating_depvars,
                                                                    param_estim =false, default_p = nothing)
-                        l = @RuntimeGeneratedFunction(l)
-                        push!(lb_, l)
+                        l_f = @RuntimeGeneratedFunction(l_expr)
+                        push!(lb_, l_f)
                     end
                 end
                 for u_ in ub
                     if u_ isa Number
                         push!(ub_, u_)
                     else
-                        u_ = NeuralPDE.build_symbolic_loss_function(nothing, indvars,depvars,
+                        u_expr = NeuralPDE.build_symbolic_loss_function(nothing, indvars,depvars,
                                                                     dict_indvars,dict_depvars,
                                                                     dict_depvar_input, phi, derivative_,
-                                                                    nothing, chain, θ, strategy,
+                                                                    nothing, chain, initθ, strategy,
                                                                     integrand = u_, integrating_depvars=integrating_depvars,
                                                                     param_estim =false, default_p = nothing)
-                        u_ = @RuntimeGeneratedFunction(u_)
-                        push!(ub_, u_)
+                        u_f = @RuntimeGeneratedFunction(u_expr)
+                        push!(ub_, u_f)
                     end
                 end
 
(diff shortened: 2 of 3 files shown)


killah-t-cell:
Ok we now support integrals of the type `Ix(u(x)*x)`, `Ix(u(x)*const)`. We do not support `Ix(u(x) + const)` right now and I don't think we need to just because `Ix(u(x) + const) = Ix(u(x)) + const`

There are two problems remaining in this PR:
1. I did some work to support `Ix(u(x)*w(x))`- type of integrals when I noticed a problem. We pass the cord number to the whole integral in this line: `ex.args = [:($(Expr(:$, :integral))), :u, Symbol(:cord, num_depvar), :phi, integrating_var_id, integrand_func, lb_, ub_,  :($θ)]` I don't know if that makes sense if we have two different dependent variables inside the integral. The inner representation ends up looking like this. 

```julia
(Expr[:((cord, var"##θ#330", phi, derivative, integral, u, p)->begin
          #= /Users/gabrielbirnbaum/.julia/dev/NeuralPDE/src/pinns_pde_solve.jl:620 =#
          #= /Users/gabrielbirnbaum/.julia/dev/NeuralPDE/src/pinns_pde_solve.jl:620 =#
          begin
              let (x,) = (cord[[1], :],)
                  begin
                      cord2 = vcat(x)
                      cord1 = vcat(x)
                  end
                  integral(u, cord2, phi, [1], RuntimeGeneratedFunctions.RuntimeGeneratedFunction{(:cord, Symbol("##θ#330"), :phi, :derivative, :integral, :u, :p), Main.NeuralPDE.var"#_RGF_ModTag", Main.NeuralPDE.var"#_RGF_ModTag", (0xad043f06, 0x65200796, 0x30dd12fc, 0xaa20459c, 0xb967c5c2)}(quote
    #= /Users/gabrielbirnbaum/.julia/dev/NeuralPDE/src/pinns_pde_solve.jl:620 =#
    #= /Users/gabrielbirnbaum/.julia/dev/NeuralPDE/src/pinns_pde_solve.jl:620 =#
    begin
        let (x,) = (cord[[1], :],)
            begin
                cord2 = vcat(x)
                cord1 = vcat(x)
            end
            (+)(u(cord1, var"##θ#330", phi), u(cord2, var"##θ#330", phi))
        end
    end
end),...
```

2. I was testing the new feature with the compound integral `Ix(u(x)*exp(x))`. It worked with no problem, but then I ran into a strange error when I ran a compound integral with `Ix(u(x)*sin(x))`, `Ix(u(x)*cos(x))`, and other real functions. 
```julia
LoadError: MethodError: no method matching real(::ChainRulesCore.Tangent{Any, NamedTuple{(:data, :uplo), Tuple{Matrix{Float64}, ChainRulesCore.ZeroTangent}}})
Closest candidates are:
real(::Complex) at complex.jl:63
real(::LinearAlgebra.Symmetric{var"#s832", S} where {var"#s832"<:Real, S<:(AbstractMatrix{var"#s832"} where var"#s832"<:var"#s832")}) at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.6/LinearAlgebra/src/symmetric.jl:353
real(::LinearAlgebra.Symmetric) at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.6/LinearAlgebra/src/symmetric.jl:355
```
This might have something to do with Quadrature? Or another part of the ecosystem that is upstream of NeuralPDE? 

I would love to hear your thoughts on this @KirillZubov

killah-t-cell:
@KirillZubov bump :) your feedback would really help me move on with this.

KirillZubov:
inf and compound expr are supported. Is it work now?  https://github.com/SciML/NeuralPDE.jl/issues/406
 @killah-t-cell  could you please check?

killah-t-cell:
@KirillZubov do you see any typos in https://github.com/SciML/NeuralPDE.jl/issues/406?

I keep getting a `ERROR: LoadError: DimensionMismatch("A has dimensions (12,1) but B has dimensions (2,402000)")`

I thought this had to do with compound integrals, so I removed the compound integral, then infinite interval, and then the integral altogether, and it still failed with the same error. Then I tested it in master, pulled, `] up`ed and still nothing.

Could you test this on your end and see if it runs?

```julia
using NeuralPDE, Flux, ModelingToolkit, GalacticOptim, Optim, DiffEqFlux, DomainSets

@parameters t E
@variables i(..)

Dt = Differential(t)
DE = Differential(E)
IE = Integral(E in DomainSets.ClosedInterval(0,Inf))

beta = 2.0;
eq = Dt(i(t, E)) + exp(-beta*E)*i(t, E) - IE(i(t,E)*exp(-beta*E))*exp(-E) ~ 0
bcs = [i(0., E) ~ exp(-E)]
domains = [t ∈ Interval(0.0,100.0), 
           E ∈ Interval(0.0, 10.0)]

           numhid=16
chain = FastChain(FastDense(1, numhid, Flux.σ), FastDense(numhid, numhid, Flux.σ), FastDense(numhid, 1))

strategy_ = GridTraining(0.05)
discretization = PhysicsInformedNN(chain,
                                   strategy_)

@named pde_system = PDESystem(eq,bcs,domains,[t, E],[i(t, E)])

cb = function (p,l)
    println("Current loss is: $l")
    return false
end
res = GalacticOptim.solve(prob, ADAM(0.1); cb = cb, maxiters=100)
```

killah-t-cell:
If you don't have time, it is ok, I can write another example tomorrow and see if it runs.

KirillZubov:
chain input size should be =2

killah-t-cell:
Oh thanks! Then this works and is probably ready to be merged. Should we add a compound integral test?

KirillZubov:
integrant with two variables work?  `Ix(u(x)*w(x))` need test with it in forward test .

killah-t-cell:
Yeah it should work. Do you have any systems in mind we could use to test?

KirillZubov:
@killah-t-cell  I will add the test myself

PR #422, opened 2021-10-30
check master

ChrisRackauckas:


KirillZubov:
we can rollback Zygote to the latest stable version and release a working release with NeuralPDE for while problems with Zygote will not be fixed.

PR #440, opened 2021-11-22
Change length of dif

killah-t-cell:
I noticed here that if we set dif to be length of domains, then it may be that it fails if the length of bcs > length of domains. I don't know if this is intended, but I thought it was wiser to set this to the length of the iterator to avoid errors.


killah-t-cell:
Opened an issue instead because I found more problems

PR #442, opened 2021-11-24
Fix reduce bug in GPU

killah-t-cell:
Right now `flat_initθ = if (typeof(chain) <: AbstractVector) reduce(vcat,initθ) else initθ end` is erroring when I run a system with a chain of type AbstractVector on the GPU. Tim helped me debug this here https://julialang.slack.com/archives/C689Y34LE/p1637573211277100

It's a small change but it should fix the problem. Here is an MWE

```julia
using CUDA

chain = [1]
initθ = cu(rand(40))
flat_initθ = if (typeof(chain) <: AbstractVector) reduce(vcat,initθ) else initθ end

# LoadError: GPUArrays.jl needs to know the neutral element for your operator `vcat`. Please pass it as an explicit argument to (if possible), or register it globally your operator by defining `GPUArrays.neutral_element(::typeof(vcat), T)`.
```

```julia
using CUDA

chain = [1]
initθ = cu(rand(40))
flat_initθ = if (typeof(chain) <: AbstractVector) reshape(initθ, :) else initθ end

# works
```

### DIFF
--- a/src/pinns_pde_solve.jl
+++ b/src/pinns_pde_solve.jl
@@ -1094,7 +1094,7 @@ function SciMLBase.discretize(pde_system::PDESystem, discretization::PhysicsInfo
 
     chain = discretization.chain
     initθ = discretization.init_params
-    flat_initθ = if (typeof(chain) <: AbstractVector) reduce(vcat,initθ) else  initθ end
+    flat_initθ = if (typeof(chain) <: AbstractVector) reshape(initθ, :) else  initθ end
     eltypeθ = eltype(flat_initθ)
     parameterless_type_θ =  DiffEqBase.parameterless_type(flat_initθ)
 
--- a/src/pinns_pde_solve.jl
+++ b/src/pinns_pde_solve.jl
@@ -1031,7 +1031,7 @@ function SciMLBase.symbolic_discretize(pde_system::PDESystem, discretization::Ph
 
     chain = discretization.chain
     initθ = discretization.init_params
-    flat_initθ = if (typeof(chain) <: AbstractVector) reduce(vcat,initθ) else initθ end
+    flat_initθ = if (typeof(chain) <: AbstractVector) reshape(initθ, :) else initθ end
     eltypeθ = eltype(flat_initθ)
     parameterless_type_θ =  DiffEqBase.parameterless_type(flat_initθ)
     phi = discretization.phi
--- a/src/pinns_pde_solve.jl
+++ b/src/pinns_pde_solve.jl
@@ -1031,7 +1031,7 @@ function SciMLBase.symbolic_discretize(pde_system::PDESystem, discretization::Ph
 
     chain = discretization.chain
     initθ = discretization.init_params
-    flat_initθ = if (typeof(chain) <: AbstractVector) reshape(initθ, :) else initθ end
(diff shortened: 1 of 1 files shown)


src/pinns_pde_solve.jl
```diff
@@ -1031,7 +1031,7 @@ function SciMLBase.symbolic_discretize(pde_system::PDESystem, discretization::Ph
 
     chain = discretization.chain
     initθ = discretization.init_params
-    flat_initθ = if (typeof(chain) <: AbstractVector) reshape(initθ, :) else initθ end
+    flat_initθ = if (typeof(chain) <: AbstractVector) vcat(initθ...) else initθ end
```
ChrisRackauckas, line 1034:
```suggestion
    flat_initθ = if (typeof(chain) <: AbstractVector) reduce(vcat,initθ) else initθ end
```

src/pinns_pde_solve.jl
```diff
@@ -1094,7 +1094,7 @@ function SciMLBase.discretize(pde_system::PDESystem, discretization::PhysicsInfo
 
     chain = discretization.chain
     initθ = discretization.init_params
-    flat_initθ = if (typeof(chain) <: AbstractVector) reshape(initθ, :) else  initθ end
+    flat_initθ = if (typeof(chain) <: AbstractVector) vcat(initθ...) else  initθ end
```
ChrisRackauckas, line 1097:
```suggestion
    flat_initθ = if (typeof(chain) <: AbstractVector) reduce(vcat,initθ) else  initθ end
```

ChrisRackauckas:
Wait, why use the splatting form? That seems like it would be less efficient? @maleadt

src/pinns_pde_solve.jl
killah-t-cell: Tim said one can't reduce vectorized expressions on the GPU (which is what caused the error). `vcat(A...)`ing is not ideal (and the GPU also warns that this is slow), but I couldn't figure out a more efficient way to do this). `reshape` would be great, but it doesn't work because initθ can be a Vector{Vector{Float64}}

maleadt:
The MWE posted on slack wasn't a `Vector{Vector{...}}`, so I didn't suggest a splat.

ChrisRackauckas:
What is it then? It should be `Vector{Vector{Float32}}` I would've thought?

maleadt:
The MWE at the start of this PR just did `cu(rand(40))`, in which case the `reduce(vcat)` is useless and can be avoided. That's what I commented on Slack. But apparently the actual use case is different.

killah-t-cell:
Yeah that was my bad, I sillily wrote the MWE with a different data type because the goal was to show `reduce(vcat,initθ)` errored in a system with a chain of type AbstractVector on the GPU.

The data type is `Vector{Vector{Float64}}` and the suggestion to `vcat(initθ...)` came from me (Tim just kindly helped navigate the MWE). The TL;DR is: we want to turn a `Vector{Vector{Float64}}` into a `Vector{Float64}`. If we use `reduce(vcat,initθ)`, it errors on the GPU, so we need a fix for that.

killah-t-cell:
Oh as long as the vector is not already flattened, this works. We used to have a line in get_numeric_integral() that broke (it was already flat), but this was fixed. Thanks for the help and sorry for the confusion!

PR #482, opened 2022-01-20
Adaptive loss merge take two

zoemcc:
Contents:

At long last, finished merging adaptive loss reweighting schemes. This includes TensorBoardLogger logging functionality for outputting the internal state of the adaptive reweighting schemes, which is basically necessary for debugging, since this is hard to find/opaque from the user's perspective.  They could find the necessary data in the adaptive loss struct but this handles it much cleaner.  The logging functionality is provided by passing a complete TensorBoardLogger struct and optional LogOptions struct to the PINN constructor.

The only thing left is a tutorial, which i want to do for SPM, but that could potentially come in a future PR (along with the hyperparameter & experiment manager stuff)


killah-t-cell:
This is pretty cool :) out of curiosity, what are the benefits of adaptive loss reweighing?

zoemcc:
> This is pretty cool :) out of curiosity, what are the benefits of adaptive loss reweighing?

It's useful for challenging PDEs where the training doesn't converge correctly with the standard techniques.  A common case is where the interior to the PDE learns some intermediate solution that satisfies the interior domain equations but it doesn't learn to match the boundary/initial conditions, or where the solution learns the boundary conditions but not the interior.  There's a competitive dynamic between the interior and the boundary and this helps mediate between them.

zoemcc:
@ChrisRackauckas @KirillZubov 
This could use a review (I can't request reviewers).  I think I'll put the tutorial in a separate PR after I've run some hyperparameter experiments to find a really good set of hyperparameters for the tutorial.  

I'm not sure what's going on with the tests.  There are some issues with my tests locally, as well.

ChrisRackauckas:
Spawned a master test https://github.com/SciML/NeuralPDE.jl/pull/487 to see what happens.

zoemcc:
They're all uniformly cancelled (or failing) after 17 minutes, is there some sort of compute/time limit that it's hitting?

ChrisRackauckas:
https://github.com/SciML/NeuralPDE.jl/pull/487 master seems to pass so there is something to look into here. Tests will cancel after the first one fails.

ChrisRackauckas:
@KirillZubov Please don't write tests/docs that rely on internal captured values 😅 

https://github.com/SciML/NeuralPDE.jl/pull/482/files#diff-23fb9a337993c3237d7576c8dfb831cda11b3be9fbec32b61da2687dcb6d154eR1469

Okay, so what's going on is that in a closure `g = (x) -> f(x)` it turns out that you have `g.f` defined. This is an internal detail that's not guaranteed to be true. 

https://github.com/SciML/NeuralPDE.jl/blob/master/src/pinns_pde_solve.jl#L1215

`loss_function_` uses `loss_function` so it exists in there. 

https://github.com/zoemcc/NeuralPDE.jl/blob/adaptive_loss_merge_2/test/forward_tests.jl#L100

This test is looking for the variable named `loss_function`. But `inner_loss =prob.f.f.pde_loss_function.pde_loss_functions.contents[1].loss_function` is probably find because it's at the top level. 

That said, this should probably just be using something like `get_loss_function` from the lower level API to be more robust (and test the lower level API)

https://neuralpde.sciml.ai/dev/solvers/pinns/#Low-level-API

ChrisRackauckas:
I set fail_fast to false so that all tests will run to completion, even if another fails. Normally we won't want this because of compute times, but let's take a peak.

zoemcc:
Ok I removed the use of the `.contents` captures in tests.  Those tests were the ones that failed locally for me as well.  I refactored `discretize` into a function `discretize_full_functions` that does everything the old `discretize` did but returns a namedtuple of all the intermediate/inner functions as well, and then `discretize` calls `discretize_full_functions` and then wraps the returned full loss function in the `OptimizationProblem` and returns that.  `discretize_full_functions` doesn't have to be the name of that function, I'll think about a better name more this weekend.

src/pinns_pde_solve.jl
```diff
@@ -110,7 +135,31 @@ struct PhysicsInformedNN{isinplace,C,T,P,PH,DER,PE,AL,K} <: AbstractPINN{isinpla
         else
             _derivative = derivative
         end
-        new{iip,typeof(chain),typeof(strategy),typeof(initθ),typeof(_phi),typeof(_derivative),typeof(param_estim),typeof(additional_loss),typeof(kwargs)}(chain,strategy,initθ,_phi,_derivative,param_estim,additional_loss,kwargs)
+
+        if typeof(adaptive_loss) <: AbstractAdaptiveLoss 
+            _adaptive_loss = adaptive_loss
+        elseif adaptive_loss == nothing
+            _adaptive_loss = NonAdaptiveLoss{Float64}()
```
ChrisRackauckas, line 142: why not just do `adaptive_loss = NonAdaptiveLoss{Float64}()` above? And why default to `Float64`? This should probably match `eltype(init_params)`.

src/pinns_pde_solve.jl
```diff
@@ -110,7 +135,31 @@ struct PhysicsInformedNN{isinplace,C,T,P,PH,DER,PE,AL,K} <: AbstractPINN{isinpla
...
+            _log_options = log_options
+        end
+
+        if iteration == nothing
+            _iteration = [1]
+            self_increment = true
+        else
+            _iteration = iteration
+            self_increment = false
+        end
```
ChrisRackauckas, line 159: Same for these, I think just use the normal kwarg handling?

src/pinns_pde_solve.jl
```diff
@@ -1206,13 +1360,113 @@ function SciMLBase.discretize(pde_system::PDESystem, discretization::PhysicsInfo
...
+        if logger isa TBLogger
+            log_value(logger, "$(name)", s, step=iteration[1])
+        end
+    end
+
+    # this is the function that gets called to do the adaptive reweighting, a function specific to the 
+    # type of adaptive reweighting being performed. 
+    # TODO: I'd love to pull this out and then dispatch on it via the AbstractAdaptiveLoss, so that users can implement their own
+    #       currently this is kind of tricky since the different methods need different types of information, and the loss functions
+    #       are generated internal to the code
```
ChrisRackauckas, line 1398: Open an issue and let's do that in a further iteration.

ChrisRackauckas:
Just fairly minor revisions requested, mostly just on simplifying the kwarg handling. It looks great.

ChrisRackauckas:
How "heavy" is the TensorboardLogger dep? Should that be made optional?

zoemcc:
> How "heavy" is the TensorBoardLogger dep? Should that be made optional?

I'm not sure how to test/measure the heaviness.  

Its Project.toml file has these dependencies:
```
CRC32c = "8bf52ea8-c179-5cab-976a-9e18b702a9bc"
FileIO = "5789e2e9-d7fb-5bc7-8068-2c6fae9b9549"
ImageCore = "a09fc81d-aa75-5fe9-8630-4744c3626534"
ProtoBuf = "3349acd9-ac6a-5e09-bcdb-63829b23a429"
Requires = "ae029012-a4dd-5104-9daa-d747884805df"
StatsBase = "2913bbd2-ae8a-5f71-8c99-4fb6c76f3a91"
```

and when I make a new project and add only TensorBoardLogger, these are all the packages/artifacts that get installed:

```
Installed ProtoBuf ─ v0.11.5
    Updating `~/.julia/dev/test_tblogger/Project.toml`
  [899adc3e] + TensorBoardLogger v0.1.18
    Updating `~/.julia/dev/test_tblogger/Manifest.toml`
  [621f4979] + AbstractFFTs v1.1.0
  [79e6a3ab] + Adapt v3.3.3
  [d360d2e6] + ChainRulesCore v1.13.0
  [9e997f8a] + ChangesOfVariables v0.1.2
  [3da002f7] + ColorTypes v0.11.0
  [c3611d14] + ColorVectorSpace v0.9.8
  [5ae59095] + Colors v0.12.8
  [34da2185] + Compat v3.41.0
  [9a962f9c] + DataAPI v1.9.0
  [864edb3b] + DataStructures v0.18.11
  [ffbed154] + DocStringExtensions v0.8.6
  [5789e2e9] + FileIO v1.13.0
  [53c48c17] + FixedPointNumbers v0.8.4
  [a2bd30eb] + Graphics v1.1.1
  [a09fc81d] + ImageCore v0.9.3
  [3587e190] + InverseFunctions v0.1.2
  [92d709cd] + IrrationalConstants v0.1.1
  [692b3bcd] + JLLWrappers v1.4.1
  [2ab3a3ac] + LogExpFunctions v0.3.6
  [dbb5928d] + MappedArrays v0.4.1
  [e1d29d7a] + Missings v1.0.2
  [e94cdb99] + MosaicViews v0.3.3
  [77ba4419] + NaNMath v0.3.7
  [6fe1bfb0] + OffsetArrays v1.10.8
  [bac558e1] + OrderedCollections v1.4.1
  [5432bcbf] + PaddedViews v0.5.11
  [21216c6a] + Preferences v1.2.4
  [3349acd9] + ProtoBuf v0.11.5
  [189a3867] + Reexport v1.2.2
  [ae029012] + Requires v1.3.0
  [a2af1166] + SortingAlgorithms v1.0.1
  [276daf66] + SpecialFunctions v2.1.4
  [cae243ae] + StackViews v0.1.1
  [82ae8749] + StatsAPI v1.2.1
  [2913bbd2] + StatsBase v0.33.16
  [899adc3e] + TensorBoardLogger v0.1.18
  [62fd8b95] + TensorCore v0.1.1
  [efe28fd5] + OpenSpecFun_jll v0.5.5+0
  [c7845625] + protoc_jll v3.19.4+0
  [0dad84c5] + ArgTools
  [56f22d72] + Artifacts
  [2a0f44e3] + Base64
  [8bf52ea8] + CRC32c
  [ade2ca70] + Dates
  [8bb1440f] + DelimitedFiles
  [8ba89e20] + Distributed
  [f43a241f] + Downloads
  [b77e0a4c] + InteractiveUtils
  [b27032c2] + LibCURL
  [76f85450] + LibGit2
  [8f399da3] + Libdl
  [37e2e46d] + LinearAlgebra
  [56ddb016] + Logging
  [d6f4376e] + Markdown
  [a63ad114] + Mmap
  [ca575930] + NetworkOptions
  [44cfe95a] + Pkg
  [de0858da] + Printf
  [3fa0cd96] + REPL
  [9a3f8284] + Random
  [ea8e919c] + SHA
  [9e88b42a] + Serialization
  [1a1011a3] + SharedArrays
  [6462fe0b] + Sockets
  [2f01184e] + SparseArrays
  [10745b16] + Statistics
  [fa267f1f] + TOML
  [a4e569a6] + Tar
  [8dfed614] + Test
  [cf7118a7] + UUIDs
  [4ec0a83e] + Unicode
  [e66e0078] + CompilerSupportLibraries_jll
  [deac9b47] + LibCURL_jll
  [29816b5a] + LibSSH2_jll
  [c8ffd9c3] + MbedTLS_jll
  [14a3606d] + MozillaCACerts_jll
  [4536629a] + OpenBLAS_jll
  [05823500] + OpenLibm_jll
  [83775a58] + Zlib_jll
  [8e850b90] + libblastrampoline_jll
  [8e850ede] + nghttp2_jll
  [3f19e933] + p7zip_jll
  ```
  
The most different dependency than what we have is probably the ProtoBuf dependency since that's Google's file format for serialization/deserialization, and possibly ImageCore?

src/pinns_pde_solve.jl
zoemcc: https://github.com/SciML/NeuralPDE.jl/issues/489#issue-1154686102

src/pinns_pde_solve.jl
zoemcc: I had mostly been following the style that was already there.  Good suggestion on the `eltype(init_params)`.  I simplified it a bit but if I'm using `eltype(init_params)` to choose the type then it needs to be default initialized after init_params is calculated so I can get the `eltype`.

src/pinns_pde_solve.jl
zoemcc: I added LogOptions to the default kwarg handling but I think iteration still needs some initialization logic here since I need to know if the user passed in an iteration vector or not as a proxy for whether or not the adaptive reweighting loss needs to estimate its own optimization increment counter.

ChrisRackauckas:
can we do the logging as a subpackage?

zoemcc:
Ok if this is green I believe it should be good to go.

The architecture that I ended up with for the tests is:
Logging tests are in the NeuralPDELogging subpackage, and they don't really test accuracy, just that logs are being generated by TBLogger.
Logging tests use the in-repo develop version of NeuralPDE so that you can add features to NeuralPDE and also logging tests for them simultaneously (needed for my current PR to not fail, for instance).
There's two more Groups to the main NeuralPDE tests:
    AdaptiveLoss: tests ~accuracy for the adaptive losses on a simple PDE to make sure they're each in the right ballpark, but doesn't test any logging
    Logging:  runs the in-repo subpackage NeuralPDELogging tests: this is so that you can run the superpackage tests and also get test readouts for the logging subpackage.  This is slightly different than the KernelAbstractions (which was used as an example for subpackage development), since our subpackage is mostly for organization and to keep dependencies off the main repo, whereas the subpackage delineation in KernelAbstractions was due to different hardware, but this allows us to keep track of regressions using the main Pkg.test
    
And then the only difference to CI is that I added those two more groups to the strategy matrix.  The interaction with the subpackage is all handled in the respective test sets so that it's isolated

killah-t-cell:
@ChrisRackauckas can we get a tag for this? Would love to try it out.

Issue #501, opened 2022-03-21
Neural Tangent Kernel


zoemcc:
Implementing the Neural Tangent Kernel adaptive loss method proposed in the "When and Why PINNs Fail to Train: A Neural Tangent Kernel Perspective" [paper](https://arxiv.org/pdf/2007.14527.pdf) by Sifan Wang, Xinling Yu, Paris Perdikaris.  There is a [github repo](https://github.com/PredictiveIntelligenceLab/PINNsNTK) that should guide implementation.

The algorithm is Algorithm 1 in the paper. The algorithm should be implemented as a concrete subtype of `AbstractAdaptiveLoss` so that it fits within our pre-existing code gen infrastructure in the `discretize_inner_functions` function.  The definition of the `K` kernels is in Lemma 3.1.

(i.e.)
```
struct NeuralTangentKernelAdaptiveLoss <: AbstractAdaptiveLoss
...
end
```

This paper is slightly harder than some of the other adaptive loss methods to implement in our system, but not that much harder.  The definition of `K` requires a selection of points from each domain, and so that could be generated via a grid or stochastic or quasi-random.  The implementation provided on their github seems to have used a Grid strategy, but I don't see why that must always be the case for this quantity (it seems arbitrary).  Thus, most of the difficulty in implementation is just figuring out the best way to have this own type maintain its own samples that are possibly different from the main PDE domain samplers for the toplevel PINN, and then calculating the kernel quantities using those points and the internal generated PDE functions.

Parvfect:
I have been going through the code written in the `discretize_inner_functions' and there seems to be some additional code for each adaptive loss method specifically of the form,
```

if adaloss isa GradientScaleAdaptiveLoss
        weight_change_inertia = discretization.adaptive_loss.weight_change_inertia
        function run_loss_gradients_adaptive_loss(θ)
            if iteration[1] % adaloss.reweight_every == 0
                # the paper assumes a single pde loss function, so here we grab the maximum of the maximums of each pde loss function
                pde_grads_maxes = [maximum(abs.(Zygote.gradient(pde_loss_function, θ)[1])) for pde_loss_function in pde_loss_functions]
                pde_grads_max = maximum(pde_grads_maxes)
                bc_grads_mean = [mean(abs.(Zygote.gradient(bc_loss_function, θ)[1])) for bc_loss_function in bc_loss_functions]

[...]

PR #506, opened 2022-04-05
Add NeuralTangentKernel Loss

Parvfect:
Adapted from paper - https://arxiv.org/pdf/2007.14527.pdf
Treating Issue #501 

Issues 
1. Still not completely sure if values used in struct are to be calculated in the struct or somewhere else 
2. Relies on ForwardDiff.compute_jacobian
3. Tests and dimension check 
4. Not completely clear on how reweighing takes place

### DIFF
--- a/src/pinns_pde_solve.jl
+++ b/src/pinns_pde_solve.jl
@@ -336,7 +336,61 @@ SciMLBase.@add_kwonly function MiniMaxAdaptiveLoss(reweight_every; pde_max_optim
 end
 
 
+"""
 
+A way of adaptively reweighing the components of the loss function by using the values of the Jacobian of the 
+NTK predictions at the current point (infinite width assumption). 
+
+* 'u_pred' : the predictions of the Network on the training data at current state
+* 'r_pred' : the predictions of the Network on the boundary conditions at current state
+* 'kernel_size' : the size of the kernel used to compute the Jacobian for the NTK matrix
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
> Relies on ForwardDiff.compute_jacobian

That's fine for now. The literature doesn't sem to say how to optimize this more, even though it's quite obvious it can be improved. Seems like a good paper idea, but for now just do it by the book.

> Not completely clear on how reweighing takes place

Feel free to ask questions on the Slack! Zoe should be around?

> Tests and dimension check

This is essential. See the other adaptive reweighing tests

Parvfect:
Haven't been able to get help from Zoe, I'll try Slack again later. 

I'll go through the tests that are written already for other adaptive losses. Is it worth reproducing results from the paper?

ChrisRackauckas:
> I'll go through the tests that are written already for other adaptive losses. Is it worth reproducing results from the paper?

It would be good to try. While it might be too expensive to be a unit test, it would make a good tutorial.

zoemcc:
I'm working on a project deadline for this Tuesday.  I'll go through and add more detailed comments after that.

ChrisRackauckas:
This should get rebased due to changes in https://github.com/SciML/NeuralPDE.jl/pull/553. That would hopefully make it much cleaner too.

PR #508, opened 2022-04-08
Test Flux 0.13

ChrisRackauckas:


ChrisRackauckas:
@KirillZubov for the neural adaptor tests, are the analytical results looking correct? They look like they are essentially zero which is weird.

PR #518, opened 2022-05-22
Fix variables in 1-dimensional IDE example

### LINKED ISSUE #517: Dummy variable and upper limit of integral should be swapped
https://github.com/SciML/NeuralPDE.jl/blob/474ef28efb8b9b7681e8e07d4dc69d7f09fca56c/docs/src/pinn/integro_diff.md?plain=1#L37-L39

In the spirit of #455, `t` and `x` should be swapped in the integral as well.

elisno:
In this PR:

- Integral variables have been swapped.
- Render "for" as plain text.

### Before fix

$$
... \int_{0}^{x}u(t)dt = 1 for t \geq 0 
$$

### After fix

$$
... \int_{0}^{t}u(x)dx = 1 \ \text{for} \ t \geq 0
$$

Closes #517

### DIFF
--- a/docs/src/pinn/integro_diff.md
+++ b/docs/src/pinn/integro_diff.md
@@ -35,7 +35,7 @@ Ix = Integral((x,y) in DomainSets.ProductDomain(ClosedInterval(0 ,1), ClosedInte
 Lets take an example of an integro differential equation:
 
 ```math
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
Thanks!

PR #557, opened 2022-06-30
[WIP] use Diffractor with NNODE

ChrisRackauckas:


src/ode_solve.jl
```diff
@@ -155,15 +155,15 @@ end
 function ode_dfdx(phi::ODEPhi{C, T, U}, t::Number, θ,
                   autodiff::Bool) where {C, T, U <: AbstractVector}
     if autodiff
-        ForwardDiff.jacobian(t -> phi(t, θ), t)
+        Diffractor.PrimeDerivativeFwd(t -> phi(t, θ))(t)
```
oscardssmith, line 158: `Diffractor.jacobian`

src/ode_solve.jl
```diff
@@ -146,7 +146,7 @@ function ode_dfdx end
 function ode_dfdx(phi::ODEPhi{C, T, U}, t::Number, θ,
                   autodiff::Bool) where {C, T, U <: Number}
     if autodiff
-        ForwardDiff.derivative(t -> phi(t, θ), t)
+        Diffractor.PrimeDerivativeFwd(t -> phi(t, θ))(t)
```
oscardssmith, line 149: can this use reverse diff? Diffractor reversediff is in much better shape.

ChrisRackauckas: Given the sizes, Forward mode makes more sense here.

PR #559, opened 2022-06-30
[WIP] Allow for setting boundary conditions as constraints

ChrisRackauckas:
This should error @Vaibhavdixit02 because both AutoZygote and BFGS don't do constraints, so...


Vaibhavdixit02:
This should error now (for BFGS) and AutoZygote works with constraints now. Is there something to do here for finishing up the supporting?

Issue #549, opened 2022-06-29
Clean up the nomenclature and document `PINNLossFunctions`


ChrisRackauckas:
`_pde_loss_functions` vs `pde_loss_functions`... yeah don't do that haha.

We need a better name for them. `datafree_pde_loss_function` vs `data_embedded_pde_loss_function`? @zoemcc any comments on the name here? If there's a clear name, then we document it for `PINNLossFunctions` and use those names everywhere, and that part of the code will be much more clear.

In the future, @KirillZubov please avoid naming schemes like that. You should be able to look at the names and know what you're dealing with, see https://github.com/SciML/SciMLStyle

PR #560, opened 2022-06-30
Restructure the documentation

ChrisRackauckas:
Fixes https://github.com/SciML/NeuralPDE.jl/issues/549


ChrisRackauckas:
@KirillZubov @ashutosh-b-b @vboussange note that this PR removes all of the things migrating to HighDimPDE.jl in order to build a fully PINN-based doc setup. I wanted to do this and get the v5 breaking update tagged, so all of the old bits will be available still just by looking at v4.x

PR #566, opened 2022-07-03
script to solve linear sde

hpieper14:
This is a script that solves a linear SDE using the KKL expansion for the Wiener process. A couple things -- 
1. This script creates individual variables for the modes. Ideally, the number of modes would be `n`, where `n` is specified by the user, something like
> order = n
> @parameters t z[1:n] 
> eq = Dt(u(t,z)) ~ f(u, t) + g(u,t)*dW(t,z)
I tried this and ModelingToolkit doesn't support differentiation of array variables. I'm not sure if there's a workaround or if that functionality needs to be extended in ModelingToolkit. 
2. I couldn't figure out how to write this script using an `ODEProblem` for the SDE. With the expansion we are using, we have time as a variable as well as `n` other variables from the expansion. I used a `PDEProblem` instead.

### DIFF
--- a/test/NNSDE_wp_tests.jl
+++ b/test/NNSDE_wp_tests.jl
@@ -1,0 +1,92 @@
+using Test, Flux, Optim, DiffEqFlux, Optimization
+using Random, NeuralPDE, DifferentialEquations
+using Statistics, Distributions
+import ModelingToolkit: Interval 
+import DomainSets: UnitInterval
+Random.seed!(100)
+
+@parameters t z1 z2
+@variables u(..)
+Dt = Differential(t)
+
+α = 1.2
+β = 1.1
+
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
This looks very much on track.

> I tried this and ModelingToolkit doesn't support differentiation of array variables. I'm not sure if there's a workaround or if that functionality needs to be extended in ModelingToolkit.

Turn it into an array of symbolic variables via `z = collect(z)`. 

> I couldn't figure out how to write this script using an ODEProblem for the SDE. With the expansion we are using, we have time as a variable as well as n other variables from the expansion. I used a PDEProblem instead.

I meant SDEProblem (sorry if I wrote ODEProblem somewhere). I meant create a version of https://github.com/SciML/NeuralPDE.jl/blob/master/src/ode_solve.jl which takes in an SDEProblem and does this solve on it. It should look very similar to the ODEProblem solver NNODE, except with the bits for handling the Wiener expansions.

Issue #574, opened 2022-07-15
Defining an array of symbolic variables causes error with NeuralPDE.discretize()


hpieper14:
I am trying to solve a PDE with variables [t, z1, …, zn]. Here’s my code:
```Julia 
using Test, Flux, Optim, DiffEqFlux, Optimization
using Random, NeuralPDE, DifferentialEquations
using Statistics, Distributions, LinearAlgebra
import ModelingToolkit: Interval 
import DomainSets: UnitInterval
Random.seed!(100)

order = 3

@parameters t z[1:order]
@variables u(..)
Dt = Differential(t)
z = collect(z)

α = 1.2
β = 1.1

function dW(t, z...)
    val = 0 
    for i in 1:length(z)
        val += z[i]*cos((i-1/2)π*t)
    end
    val = √2*val
end

eq  = Dt(u(t,z...))  ~ α*u(t,z...) + β*u(t,z)*dW(t,z...)
bcs = [u(0, z...) ~ 1.0]

# Space and time domains
domains = Vector{Symbolics.VarDomainPairing}(undef, order + 1)
domains[1] = t ∈ Interval(0.0, 1.0)
for i in 1:order
    domains[i+1] = z[i] ∈ Interval(0.0, 1.0)
end

# number of dimensions
dim = order + 1
chain = Flux.Chain(Dense(dim,16,Flux.σ),Dense(16,16,Flux.σ),Dense(16,1))
# Initial parameters of Neural network
initθ = Float64.(DiffEqFlux.initial_params(chain))

# Discretization
dx = 0.05
discretization = PhysicsInformedNN(chain,GridTraining(dx),init_params =initθ)

@named pde_system = PDESystem(eq,bcs,domains,[t,z...],[u(t,z...)])
prob = discretize(pde_system,discretization)
```
This gives the error 
```
ERROR: MethodError: no method matching nameof(::Term{Real, Base.ImmutableDict{DataType, Any}})
Closest candidates are:
  nameof(::Sym) at ~/.julia/packages/SymbolicUtils/vnuIf/src/types.jl:144
  nameof(::ModelingToolkit.AbstractSystem) at ~/.julia/packages/ModelingToolkit/iHLWM/src/systems/abstractsystem.jl:139
  nameof(::DataType) at /Applications/Julia-1.7.app/Contents/Resources/julia/share/julia/base/reflection.jl:223
  ...
Stacktrace:
[...]

PR #580, opened 2022-08-05
Array Variable interface [WIP]

hpieper14:
The goal of this PR is to extend the interface to accept `PDESystem`s that are initialized with array variables (#574) The changes so far occur in `symbolic_utilities.jl` and I expect that there will be changes that need to occur in `discretize.jl` as well. 

To be more precise, currently in `symbolic_utilities.jl`, there is a function `get_vars()` that creates `Vector{Symbol}` for both the independent and dependent variables as well as dictionaries for both sets of variables that whose (key, value) pairs correspond to (variable name, order). These objects are constructed using the name of the symbolic variables via `get_name`. Most of the other functions in this file take the symbolic arrays or dictionaries generated by `get_vars` as inputs. For example, if the independent variables are initialized as 
```Julia 
@parameters t z1 z2
```
then the functions in symbolic utilities work with something like `indvars = [:t, :z1, :z2]`. 

If we want to initialize our independent or dependent variables as arrays of symbols, this breaks down since `get_name` does not return unique names for entries of a symbolic array. For example, if our independent variables are initialized as
```Julia 
@parameters t z[1:2]
```
then `get_vars()` would return `indvars = [:t, :z, :z]` which leads to errors in the other functions since the independent variables do not have unique names. 

The goal is to not use `get_name` at all and rework the functions in `symbolic_utilities.jl` to accept the symbolic variables directly, i.e. `get_vars()` should return something like `indvars = [t, z[1], z[2]]'.

### DIFF
--- a/src/symbolic_utilities.jl
+++ b/src/symbolic_utilities.jl
@@ -360,6 +360,7 @@ function get_vars(indvars_, depvars_)
 
     dict_indvars = get_dict_vars(indvars)
     dict_depvars = get_dict_vars(depvars)
+
     return depvars, indvars, dict_indvars, dict_depvars, dict_depvar_input
 end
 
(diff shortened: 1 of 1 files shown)


r0uv3n:
Was this closed for being stale? Has there been some other work on this? I would love to be able to use symbolic arrays.

ChrisRackauckas:
Yes this PR is stale and would probably need to be restarted.

Issue #563, opened 2022-07-01
RODE handling disabled


ChrisRackauckas:
It was disabled because it was done improperly. It defined the RODE as having a fixed Wiener process, which means it would not generalize to giving solutions from the full distribution but instead only deterministically based on a single one. It needs to sample the space of noise processes in order to be more robust, for example like https://github.com/SciML/NeuralPDE.jl/issues/531 . In practice, it does work on some simple examples because you can, with a long enough process, "learn how `f` works" from just one really long noise trajectory, but only in some circumstances, and only for some lucky cases where you sample long enough.

This should get re-enabled after setting up sampling schemes over the Wiener space.

@ashutosh-b-b

PR #585, opened 2022-08-16
Enabling RODE handling

ashutosh-b-b:
Solves #563 on a first pass.

### DIFF
--- a/src/rode_solve.jl
+++ b/src/rode_solve.jl
@@ -19,9 +19,25 @@ mutable struct RODEPhi{C, T, U, S}
     chain::C
     t0::T
     u0::U
-    function RODEPhi(re::Optimisers.Restructure, t, u0)
-        new{typeof(re), typeof(t), typeof(u0), Nothing}(re, t, u0)
+    st::S
+
+    function RODEPhi(chain::Lux.AbstractExplicitLayer, t::Number, u0, st)
+        new{typeof(chain), typeof(t), typeof(u0), typeof(st)}(chain, t, u0, st)
     end
+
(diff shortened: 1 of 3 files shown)


ChrisRackauckas:
format

src/rode_solve.jl
```diff
@@ -1,38 +1,191 @@
...
+    fs = reduce(hcat, [f(out[:, i], p, arrt[i], W[i]) for i in 1:size(out, 2)])
+    dxdtguess = Array(rode_dfdx(phi, t, W, θ, autodiff))
+    sum(abs2, dxdtguess .- fs) / length(t)
+end
+
+function generate_loss(strategy::GridTraining, phi, f, autodiff::Bool, tspan, W, p, batch)
+    ts = tspan[1]:(strategy.dx):tspan[2]
 
+    # sum(abs2,inner_loss(t,θ) for t in ts) but Zygote generators are broken
+    println(typeof(W))
```
ChrisRackauckas, line 163:
```suggestion
```

avik-pal:
added in #882

ChrisRackauckas:
Is it correct though? The original formulation here is not correct as it's only with respect to a single brownian path

avik-pal:
To clarify, I re-enabled the old implementation updating it from the FastChain days and ran the existing tests on it

ChrisRackauckas:
Yes that is not a correct implementation and should not be added.

avik-pal:
commented it out

ChrisRackauckas:
Just delete it. It needs to be started fresh.

PR #603, opened 2022-09-11
change sigmoid to tanh

YichengDWu:


KirillZubov:
why actually `tanh` better `sigmoid` for PINNs issue? I didn't see it in the experiment and some theories that it can explain.
@YichengDWu could you explain or link to the paper?

YichengDWu:
https://stats.stackexchange.com/questions/330559/why-is-tanh-almost-always-better-than-sigmoid-as-an-activation-function

KirillZubov:
"tanh is almost always better than sigmoid" but Pinns problem is out from that "almost always" :) .
tanh is quite good if it need only the first degree of derivative in order to update the weight but in PINNs problem we take derivate >2  degree because use the derivative operator >1 degree  and then update weight with a gradient  +1 from this operator.

and actually, tanh is quite a bad choice, since the value of the derivative grows rapidly with the degree of the derivative for tanh. sigmoid is also a bad but little bit better. 

one of the main big problems for Pinns is what is called "gradient pathologies" and one main specific between "PINNS" and  "classic deep learning"
 https://arxiv.org/abs/2001.04536v1

- since we have a main operator and boundary conditions that exist in different dimensions and it is difficult to ensure a uniform update of the weights
- and the fact that the differential operator has different degrees of derivative, which greatly affects the amount of instability of the weight update.

if one of the gradients is much larger, then it makes the gradient of other parts invisible for updating, which leads to instability and the inability to train the neural network.

what is clear for now is that the gradient for all parts of the complete operator must be comparable in the magnitude of the gradient.
D(D1,D2,..., Dn) =f  
grad(D1) ≈ grad(D2) ≈ ... ≈ grad(Dn)

```julia
using Flux
using Zygote
using Plots

x = collect(-10:0.1:10)

σ_ = Flux.σ.(x)
tanh_ = Flux.tanh.(x)

plot(x, σ_)
plot!(x, tanh_)

#first derivative
dσ(x) = Flux.σ(x) .* (1 .- Flux.σ(x))
dtanh(x) = 1 .- Flux.tanh.(x) .^ 2
dσ_ = dσ(x)
dtanh_ = dtanh(x)

plot(x, dσ_)
plot!(x, dtanh_)

#second  derivative
ddσ(x) = Flux.σ(x) .* (1 .- Flux.σ(x)) .* (1 .- 2 * Flux.σ(x))
ddtanh(x) = -2 .* Flux.tanh.(x) .* (1 .- Flux.tanh.(x) .^ 2)

ddσ_ = ddσ(x)
ddtanh_ = ddtanh(x)

plot(x, ddσ_)
plot!(x, ddtanh_)

#third derivative

dddσ_ = Zygote.gradient(x -> sum(ddσ(x)), x)[1]
dddtanh_ = Zygote.gradient(x -> sum(ddtanh(x)), x)[1]

plot(x, dσ_, label = "dσ")
plot!(x, ddσ_, label = "ddσ")
plot!(x, dddσ_, label = "dddσ", title = "σ")


plot!(x, dtanh_, label = "dtanh")
plot!(x, ddtanh_, label = "ddtanh")
plot!(x, dddtanh_, label = "dddtanh")
```

file

 'tanh' does differently not do it and sigmoid also actually.

in summary, I completely disagree that tanh will be better, just for PINNs tanh is almost always a bad choice, and I see this both in experimental observations and in theory

It looks like this is a common misconception that tanh is almost always better, but this is true only for the classic ML and DL, but not for SciML.

This is still an open question and much is not clear which activation function or general approach  is better for one or another type of problem - dimension, degree of derivative, type of boundary conditions, and type pde (elliptic, hyperbolic, parabolic and etc)

YichengDWu:
It is not a misconception, even for PINNs, just more true in some ways and less true in very few other ways.

> in PINNs problem we take derivate >2 degree because use the derivative operator >1 degree and then update weight with a gradient +1 from this operator.

This part is true. The differential operator amplifies the sensitivity of the PDE loss to the parameters, but this is an inherent property of differential equations. If you want the network to eventually approximate the true solution, this cannot be changed anyway. It just suggests that you should add more weight to the bounds, not that sigmoid is in any sense better. The tanh just approaches the derivative faster than it does. Remember that the gradient to the parameters depends on both the gradient of the network and the size of the residuals. Also the paper you linked actually confirms that the weights of the loss function are very stable and they don't need to change much once they reach the right size.

In my experiments, tanh is almost always better. You can run https://github.com/SciML/NeuralPDE.jl/blob/master/test/adaptive_loss_tests.jl and see.

Thanks for the reply, it made me more sure of my original idea. If it doesn't convince you yet, feel free to just close this pr :).

YichengDWu:
Oh thanks for the graph also. It actually shows dσ >0

YichengDWu:
Yeah. That is a typical example where the eigenfunctions are sines and cosines, so you have `sin'(kpi x)=kpi*cos(kpi*x)` and the second derivative being `-k^2*pi^2*sin(k*pix*x)`.

KirillZubov:
> The differential operator amplifies the sensitivity of the PDE loss to the parameters, but this is an inherent property of differential equations. If you want the network to eventually approximate the true solution, this cannot be changed anyway.

The differential operator is more sensitive, and that's the problem. We are trying to predict some physical law which is described by a differential operator. and the neural network and activation function must satisfy the mathematical properties of the problem and provide smooth learning so that the sensitivity of the differential operator is minimized and does not lead to instability and unconvergence.

 In addition to the fact that the neural network must satisfy the solution, but also its derivatives too.
therefore, how the derivatives of the activation function behave is also important.

As you can see in the pic above, `tanh` is bad because different degrees of the derivative is not normalized and can differ by an order of magnitude. Which leads to non-uniform updating of the weights and that gradient pathology.

Of course, the behavior of the operator from the degree of the derivative can differ greatly, but the neural network and the activation function must satisfy this behavior and also provide a uniform update of the weights, i.e. a commensurate gradient value in backpropagation.

And in this sense, both `tanh` and sigmoid, just by luck are suitable for this or that task but do not have any universal good properties.
According to my experimental observations, this is still just a fine-tuning, sometimes `tanh` is faster to converge than a sigmoid, or sometimes `tanh` diverges and a sigmoid does not, and it also depends on the size of the neural network and many other parameters.

We cannot do anything about the inherent property of differential equations, but we can make a model that satisfy the inherent property of differential equations.

> It just suggests that you should add bigger weights to the boundary conditions, not that sigmoid is in any sense better. The tanh just approaches the derivative faster than it does. Remember that the gradient to the parameters depends on both the gradient of the network and the size of the residuals. Also the paper you linked actually confirms that the weights of the loss function are very stable and they don't need to change much once they reach the right size.

I link to the article because they proposed a term that describes some phenomenon that the gradient is unstable for PINNs relative to the classic DL. The methodology itself suggests regularization of gradient pathologies in an attempt to smooth out the error but without explaining why this pathology arises and tries to fix the consequence of the problem but not the problem itself. They also consider the problem from the point of view of dividing into boundary conditions and the main equation and trying to regulate the loss function between, but not the differential operator itself. Probably the reason is just that the boundary conditions either do not have a derivative of the first degree (Dirichlet and Neumann) and the main PDE has a higher degree of the derivative and its combination. Also, boundary conditions exist in a less dimension, but this is another topic.

My hypothesis is that gradient pathology exists first of all because of classical Dl approach is not work good with PINNs - FFNN and activation function: sigmoid and `tanh` is not satisfy physical law(mathematical properties of the deferential operator and the type and degree)  in general and universal, only sometimes with fine-tuning.
And we see that PINNs(with `ffnn` and `tanh/sigmoid`) work well only with hyperbolic pde because they have the same degree of derivative in the operator and are better with the Dirichlet boundary condition because don't have a derivative.

YichengDWu:
> As you can see in the pic above, tanh is bad because different degrees of the derivative is not normalized and can differ by an order of magnitude.

This is the reason I would use that pic to explain why tanh is almost always better. Since it has a large range of derivatives. And I wouldn't assume that the derivatives are always at their peaks in the graph. The weights can adjust that. Sensitivity is relative and the sensitivity is a stable property embedded in the differential operator. That paper may said it's unstable, but Figure 5, Figure 8, and Figure 16 all have clear mean values of weights during training. You can just assign bigger weights to the boundary conditions, so they are balanced. 

> we can make a model that satisfy the inherent property of differential equations

The inherent property is that the solution almost always has a growing magnitude of derivatives,   unless you are dealing with a boring DE of which the solution is very flat, then sigmoid might be better.

> I link to the article because they proposed a term that describes some phenomenon that the gradient is unstable for PINNs relative to the classic DL

Where the imbalance comes from is special for PINNs. But the existence of imbalance itself is a common thing in classical multi-task learning.

> The methodology itself suggests regularization of gradient pathologies in an attempt to smooth out the error 

I don't get you. The paper says only about weighting. The " regularization" is that paper is just the pde constraint.

>  sometimes tanh diverges and a sigmoid does not
 
sigmoid does regularize the output, but it has a bigger spectral bias, which is another serious issue.

> because they have the same degree of derivative in the operator 

I guess you are saying that the higher derivatives dominate the lower ones inside of PDE?  That is something I never thought about. Can you provide some examples?

ChrisRackauckas:
Let's settle this empirically in the SciMLBenchmarks. @YichengDWu can you make a a section that goes through a few different choices of loss functions with the same optimizer? I would think that `tanh` is more stable for the numerous reasons people argue, though I am curious what the effect of other non-saturating smooth functions like `gelu` would do in PINN contexts.

KirillZubov:
Would be cool write benchmark which have all set of example in docs and tests that we have now where we can setup the same parameters , nn optimizer and etc and check what going on , with logger and analyzes , it will be help to developing new approaches and compare in experiment what exactly is better work and try understand why and share this result/prove

KirillZubov:
Actually what we have in test system but with analyzes of result and the same setup for all example

YichengDWu:
From sigmoid to tanh is a representative example. Discussing other activation functions is actually a bit redundant for me.

Simply switching from one activation function to another is not actually a fair comparison, and such an experiment doesn't really tell the story.

For example, if you have `chain_a` and `chain_b`, with activation functions `a` and `b` respectively, and all other settings are identical, and you find that chain_a performs better than chain_b, does that mean that a is better than b? Not really, because the initialization of chain_a and chain_b may not be optimal. If the two perform similarly with separate optimal initializations, then you cannot simply say that a is better than b.

This also applies to tanh and sigmoid. "Simply replacing sigmoid with tanh performs better" does not mean that tanh is better than sigmoid.

What's really going on under the hood? Suppose the bandwidth of the real solution is F, the initialization bandwidth of chain_a is F_a, and chain_b is F_b, we have F_b<F_a<F. So chain_a initializes better than chain_b, and of course it converges faster, there is no mystery. Why does this not mean that a is better than b? Because we can adjust the initialization of chain_b to adjust the bandwidth so that F_a=F_b and they will have almost the same performance.

But swapping b for a is still the easiest way to improve network bandwidth.

This of course has the derived problem of how to adjust the bandwidth without creating the problem of vanishing gradients and exploding gradients. For the former, it is easy to do by increasing the variance of the weights in the first layer. For the latter, unfortunately, Flux never seems to care about it and just uses the same initialization for all activation functions. In pytorch there is a table of initializations for different activation functions https://pytorch.org/docs/stable/nn.init.html?highlight=gain#torch.nn.init.calculate_gain, which represents the edges of chaos.
Less than these values correspond to vanishing gradients and more than them correspond to exploding gradients. I did not find an equivalent in Flux. But if these two requirements are met, I don't think there is any significant difference between the different activation functions.

Personally, I just use Siren or BACON directly, they don't have the problem of disappearing and exploding gradients, and I can easily adjust the bandwidth.

ChrisRackauckas:
Such a benchmark would need to tweak more than just the activation, and use multiple runs to factor out potential initialization issues. There's enough compute time to do all of that.

YichengDWu:
For hidden layers, there is basically a unique optimal way for initialization, which is the edge of chaos. That table from pytorch could be useful. [FullyConnected](https://yichengdwu.github.io/Sophon.jl/dev/#Sophon.FullyConnected-Union{Tuple{T},%20Tuple{N},%20Tuple{Tuple{Vararg{T,%20N}},%20Function}}%20where%20{N,%20T%3C:Int64}) in my package is consistent with pytorch, which is better than `Chain` in this sense. 

The only thing that could change is the input layer, for which you can use positional embedding, larger weights, or other activation functions, as long as you can tune the bandwidth with it. I think the activation function in sequential layers is way less important.

I don't know when I would have time to write a such benchmark.  I need to finish my thesis this semester in order to look for a job next semester. Sophon.jl is part of my thesis, so I put a lot of time into developing it.

KirillZubov:
if in simple words, we cannot carry out an experiment that would show that one activation function is better than another, because it is not true :) 

or if it can only be worked under rather specific conditions, like optimal weights initialization, means that  "tangent is almost always better than sigmoid" is also not true.

And the hypothesis that would give a better prediction of PINNs lies beyond the choice of just activation function.

Issue #445, opened 2021-11-28
Kuramoto–Sivashinsky equation


cruzchue:
Dear SciML/NeuralPDE;

First, many thanks for providing your code.  I hope I can fit it for our research problem.

I have a question regarding your Kuramoto–Sivashinsky example. I basically copied/pasted your [code](https://github.com/SciML/NeuralPDE.jl/blob/master/docs/src/pinn/ks.md)  (also tried the same example provided at [neuralPDE website](https://neuralpde.sciml.ai/stable/pinn/ks/#Kuramoto%E2%80%93Sivashinsky-equation) ).

However, the prediction is different than the one reported (see figure - top left).  That is, the prediction looks sigmoidal instead of gaussian.  Any hints or suggestions about why this is happening?.  As mentioned above, the code is the same as yours.

![KS](https://user-images.githubusercontent.com/20096271/143727742-b7411b5b-4559-42c3-9eb6-184d56f01abc.png)

Regards;

Eduardo

ChrisRackauckas:
Don't use direct BFGS with GridTraining. Use QuasiRandomSample with the ADAM->BFGS mixed optimizer default of GalacticOptim and you'll generally see much better results.

cruzchue:
I did two changes following your comment.  

1.- The [line](https://github.com/SciML/NeuralPDE.jl/blob/a19c52cbf53a5ab9cbfb1609722a3fe10b2aa4ae/docs/src/pinn/ks.md?plain=1#L64): 
> discretization = PhysicsInformedNN(chain, GridTraining([dx,dt]))

was commented out and replaced by:

> discretization = PhysicsInformedNN(chain,QuasiRandomTraining(50))

2.- also, the [lines](https://github.com/SciML/NeuralPDE.jl/blob/a19c52cbf53a5ab9cbfb1609722a3fe10b2aa4ae/docs/src/pinn/ks.md?plain=1#L73-L74): 

> opt = Optim.BFGS()
> res = GalacticOptim.solve(prob,opt; cb = cb, maxiters=2000)

were commented out and replaced by:

> res = GalacticOptim.solve(prob, ADAM(0.01); cb = cb, maxiters=2000)

[...]

PR #608, opened 2022-09-22
Use different step sizes for different orders of derivatives

### LINKED ISSUE #607: The implementation of finite differences throws away its advantage over AD
I have heard quite a few times that finite differences are better than automatic differentiation in computing higher-order derivatives. However, this is not shown in this package. It can be seen in two examples in the documentation:
https://neuralpde.sciml.ai/stable/examples/3rd/
https://neuralpde.sciml.ai/stable/examples/ks/

that finite differences are quite bad. 

#445 was closed without actually fixing the issue. (Why are there some attached images from nowhere in the docs?). The culprit is the use of the same finite difference step size for all orders of derivatives, and the rounding error dominates when the order gets higher.

YichengDWu:
Closes #607 and #445

### DIFF
--- a/src/symbolic_utilities.jl
+++ b/src/symbolic_utilities.jl
@@ -80,8 +80,8 @@ function transform_expression(pinnrep::PINNRepresentation, ex; is_integral = fal
     return ex
 end
 
-function get_ε(dim, der_num, eltypeθ)
-    epsilon = cbrt(eps(eltypeθ))
+function get_ε(dim, der_num, eltypeθ, order)
+    epsilon = eltypeθ(^(eps(eltypeθ), 1/(2+order)))
     ε = zeros(eltypeθ, dim)
     ε[der_num] = epsilon
     ε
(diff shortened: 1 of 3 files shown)


YichengDWu:
The Kuramoto–Sivashinsky equation is finally working.
plot_10

src/symbolic_utilities.jl
```diff
@@ -80,8 +80,8 @@ function transform_expression(pinnrep::PINNRepresentation, ex; is_integral = fal
     return ex
 end
 
-function get_ε(dim, der_num, eltypeθ)
-    epsilon = cbrt(eps(eltypeθ))
+function get_ε(dim, der_num, eltypeθ, order)
+    epsilon = ^(eps(eltypeθ), one(eltypeθ) / (2 + order))
```
ChrisRackauckas, line 84: This seems off by one? It should be `sqrt(eps)` at first order.

src/symbolic_utilities.jl
```diff
@@ -80,8 +80,8 @@ function transform_expression(pinnrep::PINNRepresentation, ex; is_integral = fal
     return ex
 end
 
-function get_ε(dim, der_num, eltypeθ)
-    epsilon = cbrt(eps(eltypeθ))
+function get_ε(dim, der_num, eltypeθ, order)
```
ChrisRackauckas, line 83:
```suggestion
function get_ε(dim, der_num, ::Type{eltypeθ}, order) where eltypeθ
```

ChrisRackauckas: In fact, we should lock some of these down to just `Int` too

src/symbolic_utilities.jl
YichengDWu: It was cbrt(eps). When order = 1 they are consistent.

ChrisRackauckas: oh yes, double checked the formulae. It's +2 for central differences like in https://github.com/JuliaDiff/FiniteDiff.jl/blob/master/src/epsilons.jl#L30

PR #627, opened 2023-01-01
Allowing a function to be called multiple times with different inputs

nicholaskl97:
Currently something like u(x) - u(0) ~ sin(x) gets parsed as u(x) - u(x) ~ sin(x) when generating the loss function, because all instances of u(anything) are assumed to have the same input. I've made changes to _transform_expression and _dot_ that generate the correct loss function, and I made a minor edit to one 1D test case changing u(x) ~ x*cos(...) to u(x) - u(0) ~ x*cos(...), to test these changes and mimic someone trying to train a network that they new passed through (0,0) and so would represent by g(x) = u(x) - u(0).

### DIFF
--- a/.github/workflows/SpellCheck.yml
+++ b/.github/workflows/SpellCheck.yml
@@ -10,4 +10,4 @@ jobs:
       - name: Checkout Actions Repository
         uses: actions/checkout@v4
       - name: Check spelling
-        uses: crate-ci/typos@v1.32.0 
+        uses: crate-ci/typos@v1.33.1 


nicholaskl97:
Prior to my edits the loss function would implement u(x,y) - u(0,0) ~ 0 as something like
`coord1 = vcat(x,y)`
`loss = u(coord1) - u(coord1)`
Now it will do 
`coord1 = vcat(x,y)`
`loss = u(x,y) - u(0,0)`
I have not found and removed the code that inserts the line `coord1 = vcat(x,y)`, even though I don't believe it does anything anymore.

Not sure when it would come up, but the changes should also allow for things like u(cos(x)) ~ func(x), etc., which I don't think would have been parsed right before these changes.

ChrisRackauckas:
Test failures look real.

nicholaskl97:
Yeah, I think it was having trouble with multiple input dimensions, which makes sense since I was mostly working on 1D so far for simplicity. I'll give it a look soon, but will only have so much time for it in the next week or two since I'm visiting the in-laws at the moment. I do believe we've gotten closer, though.

nicholaskl97:
Everything passed on buildkite before the last commit, which only changed some docstrings from using three quotes to one (since I thought that might be why it was counting against code coverage?) and changed the number of sample points and iterations of a test in a different file from the one that is now failing. The test that failed on buildkite passes on my computer. So, I'm at a bit of a loss as to what stopped working and why.

src/symbolic_utilities.jl
```diff
@@ -460,21 +460,21 @@ function get_argument(eqs, _indvars::Array, _depvars::Array)
     get_argument(eqs, dict_indvars, dict_depvars)
 end
 function get_argument(eqs, dict_indvars, dict_depvars)
-    """Equations, as expressions"""
+    "Equations, as expressions"
     exprs = toexpr.(eqs)
```
ChrisRackauckas, line 464: You can't put docstrings inside of a function?

src/discretize.jl
```diff
@@ -61,7 +61,9 @@ function build_symbolic_loss_function(pinnrep::PINNRepresentation, eqs;
         this_eq_pair = pair(eqs, depvars, dict_depvars, dict_depvar_input)
         this_eq_indvars = unique(vcat(values(this_eq_pair)...))
     else
```
ChrisRackauckas, line 63: Update the docstring above. What does the code look like now?

src/discretize.jl
```diff
@@ -325,19 +320,33 @@ function get_bounds(domains, eqs, bcs, eltypeθ, dict_indvars, dict_depvars, str
...
-        bds = eltypeθ.(bds)
-        bds[1, :], bds[2, :]
+    pde_vars = get_variables(eqs, dict_indvars, dict_depvars)
+    pde_bounds = map(pde_vars) do pde_var
+        if !isempty(pde_var)
+            bds = mapreduce(s -> get(dict_span, s, fill(s, 2)), hcat, pde_var)
+            bds = eltypeθ.(bds)
+            bds[1, :], bds[2, :]
+        else
+            [eltypeθ(0.0)], [eltypeθ(0.0)]
```
ChrisRackauckas, line 330: what case is this handling?

src/symbolic_utilities.jl
```diff
@@ -128,14 +143,15 @@ function _transform_expression(pinnrep::PINNRepresentation, ex; is_integral = fa
             if e in keys(dict_depvars)
                 depvar = _args[1]
                 num_depvar = dict_depvars[depvar]
-                indvars = _args[2:end]
+                indvars = map((indvar_) -> transform_expression(pinnrep, indvar_),
+                              _args[2:end])
                 var_ = is_integral ? :(u) : :($(Expr(:$, :u)))
                 ex.args = if !multioutput
-                    [var_, Symbol(:cord, num_depvar), :($θ), :phi]
+                    [var_, :((_vcat)($(indvars...))), :($θ), :phi]
```
ChrisRackauckas, line 150: add a comment for why `_vcat` is used.

src/symbolic_utilities.jl
```diff
@@ -435,27 +460,48 @@ function get_argument(eqs, _indvars::Array, _depvars::Array)
     get_argument(eqs, dict_indvars, dict_depvars)
 end
 function get_argument(eqs, dict_indvars, dict_depvars)
+    "Equations, as expressions"
```
ChrisRackauckas, line 463: why is this string here?

src/symbolic_utilities.jl
```diff
@@ -435,27 +460,48 @@ function get_argument(eqs, _indvars::Array, _depvars::Array)
...
-        map(x -> first(x), f_vars)
     end
+    #    vars = [depvar for expr in vars for depvar in expr ]
     args_ = map(vars) do _vars
-        ind_args_ = map(var -> var.args[2:end], _vars)
+        "Arguments of all instances of dependent variable, by instance, by dependent variable"
+        ind_args_ = map.(var -> var.args[2:end], _vars)
         syms = Set{Symbol}()
-        filter(vcat(ind_args_...)) do ind_arg
+        "All arguments in any instance of a dependent variable"
```
ChrisRackauckas, line 477: are these supposed to be comments?

src/symbolic_utilities.jl
```diff
@@ -435,27 +460,48 @@ function get_argument(eqs, _indvars::Array, _depvars::Array)
...
+        all_ind_args = vcat((ind_args_...)...)
```
ChrisRackauckas, line 478: `reduce(vcat,...)`

src/symbolic_utilities.jl
```diff
@@ -435,27 +460,48 @@ function get_argument(eqs, _indvars::Array, _depvars::Array)
...
+        # Add any independent variables from expression dependent variable calls
+        for ind_arg in all_ind_args
+            if ind_arg isa Expr
+                for ind_var in collect(keys(dict_indvars))
+                    if !isempty(NeuralPDE.find_thing_in_expr(ind_arg, ind_var))
+                        push!(all_ind_args, ind_var)
+                    end
+                end
+            end
+        end
```
ChrisRackauckas, line 489: Ahh this looks like the key.

ChrisRackauckas:
That Lux failure could be stochastic? Let's see.

src/symbolic_utilities.jl
nicholaskl97: Yes, I think I misread the Julia documentation on docstrings. I'm inferring you only use them for explaining what functions are/do and not what a variable is/how it's organized. I've changed them to comments, which should help with code coverage, since these lines with just strings were all showing up as not running.

nicholaskl97:
@ChrisRackauckas, I've merged recent changes to the library into this branch and addressed all your comments in the most recent commit.

src/discretize.jl
```diff
@@ -10,35 +10,81 @@ Take expressions in the form:
...
+        #= ... =#
+        begin
+            (θ1, θ2) = (θ[1:205], θ[206:410])
+            (phi1, phi2) = (phi[1], phi[2])
+            let (x, y) = (cord[[1], :], cord[[2], :])
+                begin
+                    cord2 = vcat(x, y)
+                    cord1 = vcat(x, y)
+                end
+                (+).((*).(4, derivative(phi2, u, _vcat(x, y), [[0.0, ε]], 1, θ2)), derivative(phi1, u, _vcat(x, y), [[ε, 0.0]], 1, θ1)) .- 0
```
ChrisRackauckas, line 24: why are those made and then not used?

nicholaskl97: They're not used in this case any more. I think they may be used in the integral case, but that might not be true either. I can look through the different cases to see if they are ever used and remove them if not.

ChrisRackauckas: Yeah it looks like  deprecated code now so it would be good to just remove it

nicholaskl97: So, currently, `cord1 = vcat(...)` is being used for integral equations only and in my efforts to see if it's possible to remove it, I've found something I broke that wasn't being tested for the integral equations, so I'm can work more on fixing that next week. In particular, I'll look for a fix that removes any need for those lines.

nicholaskl97: Another thing I realized as I was working on this was that `Ix(u(sin(x))` will now be interpreted as $\int u(\sin x) dx$, which seems right. However, `Dx(u(sin(x))` is (under my current changes) being interpreted as $u'(\sin x)$, not $\frac{d}{dx}\left[ u(\sin x) \right] = u'(\sin x) \cos x$. They're interpreted this way because that's how the numeric integral and numeric derivative functions were already written. However, it feels a little inconsistent with the way the integral was interpreted; it's instead consistent with an interpretation of `Ix(u(sin(x))` as $U(\sin x)$, where $U$ is an antiderivative of $u$.

It feels to me like `Ix(u(sin(x))` is $\int u(\sin x) dx$ and `Dx(u(sin(x))` is $\frac{d}{dx}\left[ u(\sin x) \right]$, but then I don't know how you would actually specify $U(\sin x)$ or $u'(\sin x)$, or if you should even be allowed to. (I'm fine not letting people use $U(\sin x)$, but it feels like they should be able to use $u'(\sin x)$.)

Thoughts?

ChrisRackauckas: > However, Dx(u(sin(x)) is (under my current changes) being interpreted as

That's not correct and would not play nicely. It should give the same result as what happens when basic symbolic interactions are done:

```julia
julia> using Symbolics

julia> @variables u x
2-element Vector{Num}:
 u
 x

julia> @variables u(..) x
2-element Vector{Any}:
  u⋆
 x

julia> u(sin(x))
u(sin(x))

julia> D = Differential(x)
(::Differential) (generic function with 2 methods)

julia> D(u(sin(x)))
Differential(x)(u(sin(x)))

julia> expand_derivatives(D(u(sin(x))))
cos(x)*Differential(sin(x))(u(sin(x)))
```

ChrisRackauckas:
Are you talking with @xtalax on this to ensure it is into the rewrite?

nicholaskl97:
> Are you talking with @xtalax on this to ensure it is into the rewrite?

No, I didn't know about this rewrite. I'll message him on Slack about it.

Issue #46, opened 2019-12-22
Push nnode further


ChrisRackauckas:
We should get some examples showing bigger neural networks training things like Lotka-Volterra. Might need GPUs.

ChrisRackauckas:
starter code:

```julia
using OrdinaryDiffEq, NeuralNetDiffEq, Plots

function f(du,u,p,t)
  du[1] = p[1]*u[1] - p[2]*u[1]*u[2]
  du[2] = -p[3]*u[2] + p[4]*u[1]*u[2]
end
function f(u,p,t)
  [p[1]*u[1] - p[2]*u[1]*u[2],-p[3]*u[2] + p[4]*u[1]*u[2]]
end

p = Float32[1.5,1.0,3.0,1.0]
u0 = Float32[1.0,1.0]
prob = ODEProblem(f,u0,(0f0,3f0),p)
prob_oop = ODEProblem{false}(f,u0,(0f0,3f0),p)

true_sol = solve(prob,Tsit5())

chain = Flux.Chain(Dense(1,32, σ), Dense(32,32,tanh), Dense(32,32,tanh), Dense(32,32,tanh), Dense(32,32,tanh), Dense(32,256, σ), Dense(256,1028, tanh), Dense(1028,1028, tanh), Dense(1028,length(u0)))
opt = Flux.Descent(0.00000001)
sol  = solve(prob_oop,NeuralNetDiffEq.NNODE(chain,opt),maxiters = 100, verbose = true, dt=1/5f0)
plot(true_sol)
plot!(sol)
```

ChrisRackauckas:
From GCI:

```julia
using OrdinaryDiffEq, NeuralNetDiffEq, Plots, Flux


function f(du,u,p,t)
  du[1] = p[1]*u[1] - p[2]*u[1]*u[2]
  du[2] = -p[3]*u[2] + p[4]*u[1]*u[2]
end
function f(u,p,t)
  [p[1]*u[1] - p[2]*u[1]*u[2],-p[3]*u[2] + p[4]*u[1]*u[2]]
end

p = Float32[1.5,1.0,3.0,1.0]
u0 = Float32[1.0,1.0]
prob = ODEProblem(f,u0,(0f0,3f0),p)
prob_oop = ODEProblem{false}(f,u0,(0f0,3f0),p)

true_sol = solve(prob,Tsit5())

opt = ADAM(1e-03) #1e-04
# opt = NADAM()
# opt = Nesterov()
[...]

PR #635, opened 2023-02-05
Explored some new training strategies for large neural networks.

sdesai1287:
Fixes https://github.com/SciML/NeuralPDE.jl/issues/46


src/nnode_training_edits.jl
```diff
@@ -0,0 +1,27 @@
...
+
+func = sin
+N = 25
+chain = Lux.Chain(Lux.Dense(1, N, func), Lux.Dense(N, N, func), Lux.Dense(N, N, func),
+                    Lux.Dense(N, N, func), Lux.Dense(N, N, func),
+                    Lux.Dense(N, N, func), Lux.Dense(N, length(u0)))
+
+opt = BFGS(initial_stepnorm = 0.01)
+θ = Lux.setup(Random.default_rng(), chain)[1] |> Lux.ComponentArray .|> Float64
+alg = NeuralPDE.NNODE(chain, opt, θ, autodiff = false, strategy = GridTraining(0.01))
```
ChrisRackauckas, line 20: looks like it's still using `GridTraining`? Did you come up with a new strategy to implement here?

sdesai1287:
Sorry, seems like I accidentally pushed a file that I was experimenting with using GridTraining

src/training_strategies.jl
```diff
@@ -296,3 +296,39 @@ function get_loss_function(loss_function, lb, ub, eltypeθ, strategy::Quadrature
...
+
+    bc_loss_functions = [get_loss_function(_loss, _set, eltypeθ, strategy)
+                         for (_loss, _set) in zip(datafree_bc_loss_function, bcs_train_sets)]
+
+    pde_loss_functions, bc_loss_functions
+end
+
+function get_loss_function(loss_function, train_set, eltypeθ, strategy::FrontWeightedGridTraining;
+                           τ = nothing)
+    loss = (θ) -> mean(abs2, loss_function(train_set, θ))
```
ChrisRackauckas, line 333: I don't see how this is working in the PDE case? "Front" would only be defined for the ODEs?

ChrisRackauckas:
This is missing tests. Now that Lotka-Volterra is working, add that as a test in the NNODE stuff and make use of the new strategy and show it converges to the correct solution.

sdesai1287:
Thanks for your feedback, will start working on this stuff

src/training_strategies.jl
```diff
@@ -296,3 +296,39 @@ function get_loss_function(loss_function, lb, ub, eltypeθ, strategy::Quadrature
...
+                          for (_loss, _set) in zip(datafree_pde_loss_function,
+                                                   pde_train_sets)]
+
+    bc_loss_functions = [get_loss_function(_loss, _set, eltypeθ, strategy)
+                         for (_loss, _set) in zip(datafree_bc_loss_function, bcs_train_sets)]
+
+    pde_loss_functions, bc_loss_functions
+end
+
+function get_loss_function(loss_function, train_set, eltypeθ, strategy::FrontWeightedGridTraining;
```
ChrisRackauckas, line 331: Is this code ran in the test?

sdesai1287: I don't think so

sathvikbhagavan:
Instead of "FrontWeighted", shouldn't we have a strategy which can take multiple dts mapping to sub intervals for GridTraining? This way we can have more points in any sub interval across the interval depending on the problem.

Thoughts @sdesai1287 @ChrisRackauckas

sdesai1287:
Hi @sathvikbhagavan that definitely sounds like a great idea, and something I was thinking about as well. Different problems may be better suited for sampling more points at the end or the beginning, and giving the user the option to choose the weighting would be ideal. I first want to get this working with its corresponding test, which is why I pushed this code. I am not sure why the test file does not detect my training strategy, as I made sure it is exported in NeuralPDE.jl

sathvikbhagavan:
@sdesai1287 are you able to run it locally? 
The CI failure seems unrelated to this PR. @ChrisRackauckas, do you know what's happening in the CI build https://buildkite.com/julialang/neuralpde-dot-jl/builds/1456?

src/training_strategies.jl
ChrisRackauckas: Then remove it. The things that are not used should just be deleted.

sdesai1287:
@sathvikbhagavan I cannot run it locally either, when I run NNODE_tests.jl it says "FrontWeightedGridTraining not defined"

ChrisRackauckas:
You forgot to export it.

sdesai1287:
I added it to the list of exports in NeuralPDE.jl but it does not seem to pick that up. Do I export it in some other way?

ChrisRackauckas:
It seems to be working fine?

sathvikbhagavan:
yes, I am also able to run it locally and it works fine

sdesai1287:
Ok maybe it is an environment issue on my end, but it is just saying "UndefVarError: FrontWeightedGridTraining not defined" whenever I run NNode_tests.jl

ChrisRackauckas:
Did you dev it correctly? `]st` does that show the dev version?

sdesai1287:
When I run that command it shows this: ⌃ [315f7962] NeuralPDE v5.3.0

ChrisRackauckas:
Yeah so you never dev'd it. `]dev NeuralPDE`.

sdesai1287:
After running the dev command I get this as the error when running the test file: ERROR: importing NeuralPDE into Main conflicts with an existing identifier

ChrisRackauckas:
You need to restart your REPL after doing that. Then you should see the folder with your code in `]st` next to the package.

sdesai1287:
I see this when I restart: [315f7962] NeuralPDE v5.4.0 `~/.julia/dev/NeuralPDE`. Unfortunately I get the same error: UndefVarError: FrontWeightedGridTraining not defined

ChrisRackauckas:
And that's the folder where you put your updated code? And you restarted your Julia REPL?

sdesai1287:
Yes, I put my cloned version of NeuralPDE in that dev folder. Now when I run ] st I don't see NeuralPDE at all. When I try to dev it, I get this error: ERROR: package `NeuralPDE [315f7962]` has same name or UUID as the active project

sdesai1287:
I tried uninstalling the module and cloning it again, but I have a new error now: LoadError: UndefVarError: get_phi not defined

Any help with this would be appreciated. While trying to resolve this, I have been working on setting up the method to take in any weighting and any number of samples rather than being fixed to FrontWeightedGridTraining. It works pretty well, so I just committed it

src/nnode_training_edits.jl
```diff
@@ -0,0 +1,32 @@
+
+using NeuralPDE, OrdinaryDiffEq, DiffEqFlux, OptimizationPolyalgorithms, Lux, Random, Flux, OptimizationOptimJL, Test, Statistics
```
ChrisRackauckas, line 2: what's this? Is this supposed to be a test script?

src/training_strategies.jl
```diff
@@ -296,3 +296,80 @@ function get_loss_function(loss_function, lb, ub, eltypeθ, strategy::Quadrature
...
+    bcs_train_sets = adapt.(typeof(flat_init_params), bcs_train_sets)
+    pde_loss_functions = [get_loss_function(_loss, _set, eltypeθ, strategy)
+                          for (_loss, _set) in zip(datafree_pde_loss_function,
+                                                   pde_train_sets)]
+
+    bc_loss_functions = [get_loss_function(_loss, _set, eltypeθ, strategy)
+                         for (_loss, _set) in zip(datafree_bc_loss_function, bcs_train_sets)]
+
+    pde_loss_functions, bc_loss_functions
+end
```
ChrisRackauckas, line 334: this isn't called?

src/training_strategies.jl
ChrisRackauckas: Bump

test/NNODE_tests.jl
```diff
@@ -204,3 +204,29 @@ sol = solve(prob, NeuralPDE.NNODE(luxchain, opt; batch = true), verbose = true,
...
+opt = ADAM(0.01)
+dx = 0.05
+weights = [0.9, 0.09, 0.009]
+samples = 10000
+alg = NeuralPDE.NNODE(chain, opt, autodiff = false, strategy = NeuralPDE.WeightedGridTraining(dx, weights, samples))
+sol = solve(prob_oop, alg, verbose=true, maxiters = 100000, saveat = 0.01)
+
+println(abs(mean(true_sol .- sol)))
+println(abs(mean(sol) - mean(true_sol)))
+@test abs(mean(sol) - mean(true_sol)) < 0.5
```
ChrisRackauckas, line 232: Make a Lotkva-Volterra test case?

src/training_strategies.jl
sdesai1287: Yep, I just created it in case someone else wants to use it with the new strategy. I can remove it if you'd like

src/nnode_training_edits.jl
sdesai1287: Yes, I will remove it in the next commit

src/training_strategies.jl
ChrisRackauckas: add a test that uses it with PDEs then?

test/NNODE_tests.jl
ChrisRackauckas: I see, a bit oddly written but it works. mean difference of 0.5: isn't that quite high? The pictures were a bit better than that?

src/training_strategies.jl
```diff
@@ -296,3 +296,80 @@ function get_loss_function(loss_function, lb, ub, eltypeθ, strategy::Quadrature
     loss = (θ) -> 1 / area * f_(lb, ub, loss_function, θ)
     return loss
 end
+
+struct WeightedGridTraining <: AbstractTrainingStrategy
```
ChrisRackauckas, line 300: Missing documentation

ChrisRackauckas:
It seems like the simpler thing to do here would be to just add a `weights` field to GridTraining and require that it matches the number of points with an interpretation to PDEs as well, defaulting to `1` for everything if `nothing`?

src/ode_solve.jl
```diff
@@ -286,6 +286,21 @@ function generate_loss(strategy::StochasticTraining, phi, f, autodiff::Bool, tsp
     optf = OptimizationFunction(loss, Optimization.AutoZygote())
 end
 
+function generate_loss(strategy::WeightedGridTraining, phi, f, autodiff::Bool, tspan, p, batch)
+    ts = tspan[1]:(strategy.dx):tspan[2]
```
sathvikbhagavan, line 290: Even with WeightedGridTraining, the loss computation is still the same as GridTraining as timepoints are the same. 
So, time points for this strategy should be computed and passed here instead of in the  __solve function

test/NNODE_tests.jl
sdesai1287: Yep, I only used 0.5 because that is what was used in the other tests. The value is always much lower than that, I can make it test threshold lower if you want me to

src/ode_solve.jl
sdesai1287: Yep that sounds like a great idea, I will work on that now

src/training_strategies.jl
sdesai1287: Will work on this. I can also add the weights and samples field directly into GridTraining if that works

sdesai1287:
Just pushed a commit which changes the loss computation to calculate the weighted time points. I also realized the method isnt really a type of grid training, so I removed dx and changed the name. Adding some documentation to the method now before I try to do some other stuff. Chris, you mentioned possibly trying to add this as a part of GridTraining. I am not sure that applies anymore, but if that is something you would still like me to do, please let me know.

src/nnode_training_edits.jl
sdesai1287: Removed

Project.toml
```diff
@@ -9,6 +9,7 @@ ArrayInterfaceCore = "30b0a656-2188-435a-8636-2ec0e6a096e2"
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
 ComponentArrays = "b0b7db55-cfe3-40fc-9ded-d10e2dbeff66"
 DiffEqBase = "2b5f629d-d688-5b77-993f-72d75c75574e"
+DiffEqFlux = "aae7a2af-3d4f-5e19-a356-7da93b79d9d0"
```
ChrisRackauckas, line 12: why is this necessary?

src/neural_adapter.jl
```diff
@@ -67,6 +67,20 @@ function get_loss_function_(loss, init_params, pde_system, strategy::QuasiRandom
...
+#     end
+#     domains = pde_system.domain
+#     depvars, indvars, dict_indvars, dict_depvars = get_vars(pde_system.indvars,
+#                                                             pde_system.depvars)
+#     eltypeθ = eltype(init_params)
+#     dx = strategy.dx
+#     train_set = generate_training_sets(domains, dx, eqs, eltypeθ)
+#     get_loss_function(loss, train_set, eltypeθ, strategy)
+# end
+
```
ChrisRackauckas, line 83: please just delete.

src/ode_solve.jl
```diff
@@ -286,6 +286,41 @@ function generate_loss(strategy::StochasticTraining, phi, f, autodiff::Bool, tsp
...
+    
+    function loss(θ, _)
+        if batch
+            sum(abs2, inner_loss(phi, f, autodiff, ts, θ, p))
+        else
+            sum(abs2, [inner_loss(phi, f, autodiff, t, θ, p) for t in ts])
+        end
+    end
+    optf = OptimizationFunction(loss, Optimization.AutoZygote())
+end
```
ChrisRackauckas, line 322: Didn't we discuss just adding `weights` to `GridTraining`?

src/training_strategies.jl
```diff
@@ -296,3 +296,33 @@ function get_loss_function(loss_function, lb, ub, eltypeθ, strategy::Quadrature
...
+Then, we randomly generate the corresponding amount of points for each interval in the timespan
+
+## Positional Arguments
+
+* `weights`: A vector of weights that should sum to 1, representing the proportion of samples at each interval.
+* `samples`: the total number of samples that we want, across the entire time span
+
+"""
+
+struct WeightedSampleTraining{T} <: AbstractTrainingStrategy
```
ChrisRackauckas, line 316:
```suggestion
"""
struct WeightedSampleTraining{T} <: AbstractTrainingStrategy
```

Project.toml
```diff
@@ -23,16 +24,21 @@ ModelingToolkit = "961ee093-0014-501f-94e3-6117800e7a78"
...
 RecursiveArrayTools = "731186ca-8d62-57ce-b412-fbd966d074cd"
 Reexport = "189a3867-3050-52da-a836-e630ba90ab69"
+Revise = "295af30f-e4ad-537b-8983-00126c2a3abe"
 RuntimeGeneratedFunctions = "7e49a35a-f44a-4d26-94aa-eba1b4ca6b47"
 SciMLBase = "0bca4576-84f4-4d90-8ffe-ffa030f20462"
 Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 StochasticDiffEq = "789caeaf-c7a9-5a7d-9973-96adeb23e2a0"
 SymbolicUtils = "d1185830-fcd6-423d-90d6-eec64667417b"
 Symbolics = "0c5d862f-8b57-4792-8d23-62f2024744c7"
+Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
```
ChrisRackauckas, line 41: This isn't needed.

Project.toml
```diff
@@ -23,16 +24,21 @@ ModelingToolkit = "961ee093-0014-501f-94e3-6117800e7a78"
 Optim = "429524aa-4258-5aef-a3af-852621145aeb"
 Optimisers = "3bd65402-5787-11e9-1adc-39752487f4e2"
 Optimization = "7f7a1694-90dd-40f0-9382-eb1efda571ba"
+OptimizationOptimJL = "36348300-93cb-4f02-beb5-3c3902f8871e"
+OptimizationPolyalgorithms = "500b13db-7e66-49ce-bda4-eed966be6282"
+OrdinaryDiffEq = "1dea7af3-3e70-54e6-95c3-0bf5283fa5ed"
```
ChrisRackauckas, line 29: Why were these added?

sdesai1287: I did not add them, they must have been automatically added when I was testing something somewhere

Project.toml
sdesai1287: Yep, I did not add it, it must have been automatically added

src/neural_adapter.jl
sdesai1287: will do

src/ode_solve.jl
```diff
@@ -286,6 +286,41 @@ function generate_loss(strategy::StochasticTraining, phi, f, autodiff::Bool, tsp
...
+
+    weights = strategy.weights ./ sum(strategy.weights)
+
+    N = length(weights)
+    samples = strategy.samples
+
+    difference = (maxT - minT) / N
+
+    # data = rand(1, trunc(Int, samples * weights[1])) .* difference .+ minT
+    data = Float64[]
```
ChrisRackauckas, line 305: Just pre-size this instead of allocating?

sathvikbhagavan:
@sdesai1287, can you post screen shots of the plots of the predictions of the lv test case with this method to confirm it is learning accurately?

src/ode_solve.jl
```diff
@@ -378,6 +413,7 @@ function DiffEqBase.__solve(prob::DiffEqBase.AbstractODEProblem,
         alg.batch
     end
 
+    println("generating loss")
```
ChrisRackauckas, line 416:
```suggestion
```

src/ode_solve.jl
```diff
@@ -286,6 +286,41 @@ function generate_loss(strategy::StochasticTraining, phi, f, autodiff::Bool, tsp
     optf = OptimizationFunction(loss, Optimization.AutoZygote())
 end
 
+function generate_loss(strategy::WeightedSampleTraining, phi, f, autodiff::Bool, tspan, p, batch)
+    # ts = tspan[1]:(strategy.dx):tspan[2]
+
+    # sum(abs2,inner_loss(t,θ) for t in ts) but Zygote generators are broken
```
ChrisRackauckas, line 292:
```suggestion
```

These don't make sense.

src/ode_solve.jl
sdesai1287: fixed

src/ode_solve.jl
sdesai1287: fixed

sdesai1287:
> @sdesai1287, can you post screen shots of the plots of the predictions of the lv test case with this method to confirm it is learning accurately?

here is a screenshot of the graph
Screen Shot 2023-03-17 at 10 08 29 AM

src/training_strategies.jl
sdesai1287: fixed

Project.toml
sdesai1287: fixed

Project.toml
```diff
@@ -23,16 +24,21 @@ ModelingToolkit = "961ee093-0014-501f-94e3-6117800e7a78"
...
 QuasiMonteCarlo = "8a4e6c94-4038-4cdc-81c3-7e6ffdb2a71b"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 RecursiveArrayTools = "731186ca-8d62-57ce-b412-fbd966d074cd"
 Reexport = "189a3867-3050-52da-a836-e630ba90ab69"
+Revise = "295af30f-e4ad-537b-8983-00126c2a3abe"
```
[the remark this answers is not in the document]
sdesai1287: fixed

src/training_strategies.jl
```diff
@@ -296,3 +296,33 @@ function get_loss_function(loss_function, lb, ub, eltypeθ, strategy::Quadrature
...
+
+A training strategy that generates points for training based on the given inputs. 
+We split the timespan into equal segments based on the number of weights, 
+then sample points in each segment based on that segments corresponding weight,
+such that the total number of sampled points is equivalent to the given samples
+
+## Positional Arguments
+
+* `weights`: A vector of weights that should sum to 1, representing the proportion of samples at each interval.
+* `samples`: the total number of samples that we want, across the entire time span
```
ChrisRackauckas, line 314:
```suggestion
* `samples`: the total number of samples that we want, across the entire time span

## Limitations

This training strategy can only be used with ODEs (`NNODE`).
```

test/NNODE_tests.jl
```diff
@@ -204,3 +204,26 @@ sol = solve(prob, NeuralPDE.NNODE(luxchain, opt; batch = true), verbose = true,
             maxiters = 400,
             abstol = 1.0f-8, dt = 1 / 5.0f0)
 @test sol.errors[:l2] < 0.5
+
+using NeuralPDE, OrdinaryDiffEq, OptimizationPolyalgorithms, Lux, Random, Flux, OptimizationOptimJL, Test, Statistics, Plots, Optimisers
```
ChrisRackauckas, line 208: Usings at the top, are most of them in here already?

ChrisRackauckas: You don't need most of those?

ChrisRackauckas:
Rebase and fix the tests?

sdesai1287:
Ok the test file has been fixed

ChrisRackauckas (approved):

ChrisRackauckas (approved):

YichengDWu:
Nice PR, I wonder if it is possible to simplify it with [StatsBase.wsample!](https://juliastats.org/StatsBase.jl/stable/sampling/#StatsBase.wsample!) or just tune the parameters in Beta distribution.

PR #646, opened 2023-02-23
ComponentArray ERROR in Lux

Xinyu-Li-1997:
ComponentArray in Lux has been removed, so I used ComponentArrays.jl to connect gpu. "ps = ps |> Lux.ComponentArray |> gpu .|> Float64" has been rewriten as "ps = ps |> ComponentArray |> gpu .|> Float64" It can work on my PC


ChrisRackauckas:
Yes thanks for catching this.

PR #665, opened 2023-03-29
removed an extra end keyword

moonpiee:


lib/NeuralPDELogging/src/NeuralPDELogging.jl
```diff
@@ -21,4 +21,4 @@ function NeuralPDE.logscalar(logger::TBLogger, scalar::R, name::AbstractString,
     nothing
 end
 
-end
```
ChrisRackauckas, line 24: This is the end for the module? The tests will fail with this change. Was this tested locally?

Issue #640, opened 2023-02-12
Recreate the examples from the BINN paper with NNODE


ChrisRackauckas:
https://www.ncbi.nlm.nih.gov/pmc/articles/PMC7732115/#:~:text=Biologically%2Dinformed%20neural%20networks%20(BINNs,systems%20from%20sparse%20experimental%20data.

ChrisRackauckas:
https://arxiv.org/abs/2210.07880 is another case with examples.

ChrisRackauckas:
Completed by @AstitvaAggarwal

Issue #572, opened 2022-07-07
functional inverse problem


YichengDWu:
I have an inverse problem where the parameter function is a function of the solution, what is the correct way to implement it?
```julia
using ModelingToolkit, NeuralPDE, Lux, Random, NNlib
using Optimization, OptimizationOptimJL, Zygote
import ModelingToolkit: Interval

@parameters x
@variables u(..), b(..)

f(x) = (1+(sin(π*x)^2))/(1+2*(sin(π*x)^2)) * π^2 * sin(π*x) + sin(π*x)
Dxx = Differential(x)^2
eq = [-b(u(x)) * Dxx(u(x)) + u(x) ~ f(x)]

bcs = [u(0) ~ 0, u(1) ~ 0]
domain = [x ∈ Interval(0.0, 1.0)]

@named pdesys = PDESystem(eq, bcs, domain, [x], [u(x), b(u(x))])

# the neural networks
chain_u =  Chain(Dense(1, 20, tanh), Dense(20,20, tanh), Dense(20,1))
chain_b =  Chain(Dense(1, 20, tanh), Dense(20,20, tanh), Dense(20,1))

x_data = reshape(LinRange(0,1,101)[2:end-1],1,:) |> collect
u_data = sin.(π .* x_data)

function additional_loss(phi, θ, p)
    return sum(abs2, phi[1](x_data, θ.depvar.u).-u_data)/length(x_data)
end

dx = 0.01
discretization = NeuralPDE.PhysicsInformedNN([chain_u, chain_b],
                                             NeuralPDE.GridTraining(dx),
                                             additional_loss=additional_loss)  

prob = NeuralPDE.discretize(pdesys, discretization)                
callback = function (p,l)
    println("Current loss is: $l")
    return false
end
res = Optimization.solve(prob, BFGS(), callback = callback, maxiters=500)
```

MilkshakeForReal:
Changing `b(u(x))` to `b(x)` would work

MilkshakeForReal:
But yeah, a very good feature to have.

PR #666, opened 2023-03-30
added additional loss against data for NNODE

AstitvaAggarwal:
added option for additional loss against data for NNODE(used PhysicsInformedNN as a reference),work in progress as im getting some errors would greatly appreciate any help. Thanks
. (trying to solve issue #640, but NNODE dosent allow additional losses against surrogate Model predictions nor does it allow parameter estimation so this pr could help solve such ODE problems)


xtalax:
This looks good, can you write some tests to ensure its working? We will need to do documentation too.

AstitvaAggarwal:
in the paper the loss term has additional loss terms :Lgls and Lconstr(under introduction eq 3 to 8), Lgls is defined against cell density predictions via a Surrogate Model MLP which takes in training data as measured data(x,t), Lconstr performs in parameter Estimation for the Growth, Time Delay and Diffusion terms(parameter functions are functions of the solution itself -> leads to issue #572).

xtalax:
Ah I see it, yep this looks like it implements what's missing, good job (y)

ChrisRackauckas:
The test failure looks real.

src/ode_solve.jl
```diff
@@ -283,10 +291,44 @@ function generate_loss(strategy::StochasticTraining, phi, f, autodiff::Bool, tsp
...
+    weights = strategy.weights ./ sum(strategy.weights)
+
+    N = length(weights)
+    samples = strategy.samples
+
+    difference = (maxT - minT) / N
+
+    data = Float64[]
+    for (index, item) in enumerate(weights)
+        temp_data = rand(1, trunc(Int, samples * item)) .* difference .+ minT .+
```
ChrisRackauckas, line 312: Rebase issue? Rebase to the new master.

ChrisRackauckas:
Rebase this onto the new master so that it's not dependent on any other PR (it's now merged).

src/ode_solve.jl
```diff
@@ -407,7 +418,34 @@ function DiffEqBase.__solve(prob::DiffEqBase.AbstractODEProblem,
...
+    elseif strategy isa StochasticTraining
+        Optimization.AutoZygote()
+    elseif strategy isa WeightedIntervalTraining
+        Optimization.AutoZygote()
+    else
+        # by default GridTraining choice of Optimization
+        # if adding new training algorithms we can extend this,
+        # if-elseif-else block for choices of optimization algos
+        Optimization.AutoZygote()
+    end
```
ChrisRackauckas, line 445:
```suggestion
    opt_algo = if strategy isa QuadratureTraining
        Optimization.AutoForwardDiff()
    else
        Optimization.AutoZygote()
    end
```

src/ode_solve.jl
```diff
@@ -23,7 +23,14 @@ of the physics-informed neural network which is used as a solver for a standard
...
 ## Keyword Arguments
-
+* `additional_loss`: A function additional_loss(phi, θ) where phi are the neural network trial solutions,
+                     θ are the weights of the neural network(s).
+example: 
+    ts=[t for t in 1:100]
+    (u_, t_) = (analytical_func(ts), ts)
+    function additional_loss(phi, θ)
+        return sum(sum(abs2, [phi(t, θ) for t in t_] .- u_)) / length(u_)
+    end
```
ChrisRackauckas, line 33: This isn't in the right spot. Make an example section.

src/ode_solve.jl
```diff
@@ -312,7 +322,8 @@ function generate_loss(strategy::WeightedIntervalTraining, phi, f, autodiff::Boo
             sum(abs2, [inner_loss(phi, f, autodiff, t, θ, p) for t in ts])
         end
     end
-    optf = OptimizationFunction(loss, Optimization.AutoZygote())
+
+    return loss
```
ChrisRackauckas, line 326:
```suggestion
    return loss
```

src/ode_solve.jl
```diff
@@ -250,38 +260,38 @@ function generate_loss(strategy::QuadratureTraining, phi, f, autodiff::Bool, tsp
...
-
         if batch
             sum(abs2, inner_loss(phi, f, autodiff, ts, θ, p))
         else
             sum(abs2, [inner_loss(phi, f, autodiff, t, θ, p) for t in ts])
         end
     end
-    optf = OptimizationFunction(loss, Optimization.AutoZygote())
+
+    return loss
```
ChrisRackauckas, line 294:
```suggestion
    return loss
```

src/ode_solve.jl
AstitvaAggarwal: under this argument's description? or along with the lower example?

ChrisRackauckas (approved):

Issue #500, opened 2022-03-21
Inverse Dirichlet Adaptive Loss


zoemcc:
Implementing the Inverse Dirichlet adaptive loss method proposed in this paper "Inverse Dirichlet weighting enables reliable training
of physics informed neural networks" [paper](https://iopscience.iop.org/article/10.1088/2632-2153/ac3712/pdf) by Suryanarayana Maddu, Dominik Sturm, Christian L Müller, and Ivo F Sbalzarini.

The algorithm is section (3.2) in the paper.  The algorithm should be implemented as a concrete subtype of `AbstractAdaptiveLoss` so that it fits within our pre-existing code gen infrastructure in the `discretize_inner_functions` function.  Note that the algorithm in section (3.1) is implemented in our repo as `GradientScaleAdaptiveLoss`, and and that implementation is a good reference for the implementation of the Inverse Dirichlet algorithm since they utilize many of the same quantities.

(i.e.)
```
struct InverseDirichletAdaptiveLoss <: AbstractAdaptiveLoss
...
end
```

This doesn't seem that difficult to implement as most of the quantities are already calculated in the `GradientScaleAdaptiveLoss`, the method just has to use the calculated values differently.  This would be good to compare as it seems similar to some of the ADAM-like variance-analyzing optimizers.

hpieper14:
I'll start looking into this!

PR #504, opened 2022-03-25
Inverse Dirichlet Adaptive Loss

hpieper14:
Adds inverse dirichlet adaptive loss function to address [#500](https://github.com/SciML/NeuralPDE.jl/issues/500)

PR #673, opened 2023-04-01
Added i) Inverse Dirichlet Adaptive Loss and ii) Neural Tangent Kernel Loss Implementations

sphinx-tech:
[Neural Kernel Tangent Adaptive Loss](https://github.com/SciML/NeuralPDE.jl/issues/501) 
- [Paper (Section 3) and algorithm 1](https://arxiv.org/pdf/2007.14527.pdf) 

[Inverse Dirichlet Adaptive Loss](https://github.com/SciML/NeuralPDE.jl/issues/500) 
- [Paper (Section 3.2)](https://iopscience.iop.org/article/10.1088/2632-2153/ac3712/pdf)


sphinx-tech:
Instead of using the Jacobians for `K` (as implemented) by the authors; I have used dot products because we were only concerned with the main diagonal entries of the `K` matrix for our implementation - in Algorithm 1 (Pg. 10)

xtalax:
Put Plots and Revise in to your `v1.8` environment, and add `using Plots, Revise` to your juliarc

ChrisRackauckas:
Note that for this there is starter code in https://github.com/SciML/NeuralPDE.jl/pull/504 and https://github.com/SciML/NeuralPDE.jl/pull/506. It doesn't make sense to do both in the same PR though.

sphinx-tech:
@xtalax @ChrisRackauckas I have made some changes according to your suggestions.

- Two independent commits have been made that deal with the different loss functions
- The Inverse Dirichlet is performing well, but the loss in Neural Tangent kernel is high

Pls let me know of any improvements that can be made in the implementations.
Also any direction on how to improve the testing part would be very helpful !

ChrisRackauckas:
Rebase onto master.

ChrisRackauckas:
The next step would be to setup tutorials and benchmarks with these.

ChrisRackauckas:
Looks like tests failed.

PR #684, opened 2023-05-13
Fix Mixed derivatives

YichengDWu:


YichengDWu:
Ah, this is just to verify a guess if the current implementation is correct

YichengDWu:
I believe the source code has reversed the order of εs. It should go from the innermost derivative to the outermost. This mistake might have been obscured in the past due to Dxy=Dyx.

YichengDWu:
Now I can confirm:
```julia
using NeuralPDE, Lux, ModelingToolkit, Optimization, OptimizationOptimJL

@parameters t, x
@variables u(..)
Dt = Differential(t)
Dx = Differential(x)
Dxx = Differential(x)^2

eq = Dt(Dxx(u(t, x)))  ~ 0
bcs = [u(0, x) ~ 0]

domains = [t ∈ Interval(0.0, 1.0),
    x ∈ Interval(-1.0, 1.0)]

# Discretization
dx = 0.05

# Neural network
chain = Lux.Chain(Dense(2, 16, Lux.σ), Dense(16, 16, Lux.σ), Dense(16, 1))
strategy = NeuralPDE.GridTraining(dx)

indvars = [t, x]
depvars = [u(t, x)]
@named pde_system = PDESystem(eq, bcs, domains, indvars, depvars)

discretization = PhysicsInformedNN(chain, strategy)
sym_prob = symbolic_discretize(pde_system, discretization)
```
The symbolic loss function for `eq` is
```julia
 begin
        let (t, x) = (cord[[1], :], cord[[2], :])
            begin
                cord1 = vcat(t, x)
            end
            derivative(phi, u, cord1, [[0.000740095979741405, 0.0], [0.0, 0.000740095979741405], [0.0, 0.000740095979741405]], 3, var"##θ#226") .- 0
        end
    end
```
Then https://github.com/SciML/NeuralPDE.jl/blob/master/src/pinn_types.jl#L393 computes the derivative in direction `[0.0, 0.000740095979741405]` as the outermost one, i.e.,`Dx`, but the correct one should be `Dt`.

ChrisRackauckas:
Have you talked with @xtalax about his rewrite of the parser? At this point I'm pretty sure we'll just throw out the one from here and start with the new one once it's ready enough. It would be good to get a few more eyeballs on that.

YichengDWu:
Not yet, I am currently quite busy preparing for the upcoming graduation thesis defense. However, I will open an issue about rewriting NeuralPDE.

ChrisRackauckas:
did you have a test case for this?

YichengDWu:
I can't think of any equations that has `Dxx(Dy(u))` for now. I'll come back to this once I find one.

xtalax:
Can you point me to the test that verifies this works?

YichengDWu:
No need to worry too much about this bug, nothing should be affected by it, after all, we almost always have Dxy=Dyx.

As Chris said, let's now focus on rolling out the new parsing backend.

YichengDWu:
I mentioned in #687 that we should generated a nested derivative expression to handle mixed derivatives, so this should no longer be relevant.

PR #692, opened 2023-06-06
[WIP] Bayesian PINN solver for ODE

### LINKED ISSUE #682: Bayesian Inference for NNODE
#205  [Bayesian Neural PDEs](https://github.com/SciML/NeuralPDE.jl/issues/205)
### **For ODE solving**

1. Uses Turing.jl Models and for reference the turing_inference method in DiffEqBayes.jl. (my Turing.jl implementation is too slow)
2. Return a MCMCChains.jl chain as a solution for further analysis and plotting.
3. Uses AdvancedHMC.jl for faster sampling than Turing.jl, its Modular design.

AstitvaAggarwal:
fixes #682


src/turing_MCMC.jl
```diff
@@ -0,0 +1,49 @@
...
+    alpha = 0.09
+    sig = sqrt(1.0 / alpha)
+    @model function bayes_pinn(dataset)
+        theta = Vector{T}{undef, length(priors)}
+        for i in eachindex(priors)
+            theta[i] ~ NamedDist(priors[i], sym[i])
+        end
+        nnparameters ~ MvNormal(zeros(nparameters), sig .* ones(nparameters))
+
+        preds = pinn(prob, parameters, dataset, recon)
```
Vaibhavdixit02, line 28:
```suggestion
        preds = pinn(prob, nnparameters, dataset, recon)
```
Did you mean this here?

src/turing_MCMC.jl
```diff
@@ -0,0 +1,49 @@
...
+        for i in eachindex(ts)
+            datapoints ~ MvNormal(pred)
```
Vaibhavdixit02, line 30:
```suggestion
            dataset ~ MvNormal(pred)
```

src/turing_MCMC.jl
```diff
@@ -0,0 +1,49 @@
...
+            Turing.@addlogprob! physloglikelihood(pred, μ)
```
Vaibhavdixit02, line 31: It would be good for future refernce to add comments with the equation the likelihoods correspond to for example `p(theta| pred)` here iirc, will also help you think more clearly about the individual components here

AstitvaAggarwal:
only changed Project.toml, Manifest.toml, NeuralPDE.jl and added 2 new files(BPINN_Tests.jl, advancedHMC_MCMC.jl), the rest is the result of JuliaFormatter.format().

Vaibhavdixit02:
Why are there formatting changes in this PR, I think it might have been by mistake? Try to see if you can remove those since it makes hard to review relevant files

AstitvaAggarwal:
yeah in commit-978e115 I formatted the whole NeuralPDE directory, ive fixed it now. Sorry for the inconvenience.

src/advancedHMC_MCMC.jl
```diff
@@ -0,0 +1,215 @@
...
+                                     θ) where {C <: Optimisers.Restructure, S}
+    f.prob.u0 + (t - f.prob.tspan[1]) * first(f.chain(θ)(adapt(parameterless_type(θ), [t])))
+end
+
+using ForwardDiff: Chunk
+# ODE DU/DX
+function NNodederi(phi::LogTargetDensity, t::AbstractVector, θ, autodiff::Bool)
+    if autodiff
+        # diag(ForwardDiff.jacobian(t -> phi(t, θ), t))
+        [ForwardDiff.derivative(ti -> phi(ti, θ), ti) for ti in t]
```
Vaibhavdixit02, line 95: Use broadcast here

```
julia> using ForwardDiff

julia> f = x -> x^2
#7 (generic function with 1 method)

julia> @time [ForwardDiff.derivative(f, ti) for ti in 1:5]
  0.137852 seconds (206.46 k allocations: 14.140 MiB, 10.25% gc time, 99.81% compilation time)
5-element Vector{Int64}:
  2
  4
  6
  8
 10

julia> @time ForwardDiff.derivative.(f, 1:5)
  0.066792 seconds (104.22 k allocations: 7.056 MiB, 99.58% compilation time)
5-element Vector{Int64}:
  2
  4
  6
  8
 10

julia> @time [ForwardDiff.derivative(f, ti) for ti in 1:5]
  0.041586 seconds (26.12 k allocations: 1.746 MiB, 99.43% compilation time)
5-element Vector{Int64}:
  2
  4
  6
  8
 10

julia> @time ForwardDiff.derivative.(f, 1:5)
  0.000028 seconds (8 allocations: 328 bytes)
5-element Vector{Int64}:
  2
  4
  6
  8
 10
```

AstitvaAggarwal:
7mins for finite diff derivative scheme and ~20mins for autodiff, 700 samples

AstitvaAggarwal:
7mins for finite diff derivative scheme and ~20mins for autodiff, 700 samples. beyond a certain number of samples the Hamiltonian energy starts stagnating, maybe [Adaptive weighting_of BPINNs]https://www.researchgate.net/publication/368829305_Adaptive_weighting_of_Bayesian_physics_informed_neural_networks_for_multitask_and_multiscale_forward_and_inverse_problems)

src/advancedHMC_MCMC.jl
```diff
@@ -0,0 +1,318 @@
...
+    meaninv = [invparam[1] for invparam in invparams]
+    # nn weights
+    nnwparams = allparams[1]
+    varw = nnwparams[2]^2
+    prisw = nnwparams[1] .* ones(length(θ) - Tar.extraparams)
+
+    if Tar.extraparams > 0
+        # ode parameters
+        invparams = allparams[2:length(allparams)]
+        varinv = [invparam[2]^2 for invparam in invparams]
```
Vaibhavdixit02, line 171: Don't need to create this vector I think there's a dispatch for `MvNormal(::Vector, ::Real)`

AstitvaAggarwal: each ODE param might have a different std deviations so ill have to pass a vector, but yeah ive changed this part in other parts of the code.

test/BPINN_Tests.jl
```diff
@@ -0,0 +1,221 @@
...
+plot(sol1.t, sol1.u)
+
+x̂ = collect(Float64, Array(u) + 0.005 * randn(size(u)))
+dataset = [x̂, time]
+plot!(time, x̂)
+
+# chainfh = Flux.Chain(Dense(1, 8, sigmoid_fast), Dense(8, 2))
+chainfh = Flux.Chain(Dense(1, 5, sigmoid), Dense(5, 1))
+
+fh_mcmc_chain1, fhsamples1, fhstats1 = ahmc_bayesian_pinn_ode(prob, chainfh, dataset,
```
Vaibhavdixit02, line 176: Since you are redefining `prob` `dataset` etc you don't expect this one to work for LV? You are passing the `param` for LV here hence the confusion.

src/advancedHMC_MCMC.jl
```diff
@@ -0,0 +1,317 @@
...
+    # ode parameters
+    invparams = allparams[2:length(allparams)]
+    stdinv = [invparam[2] for invparam in invparams]
+    # nn weights
+    nnwparams = allparams[1]
+    stdw = nnwparams[2]
+    prisw = nnwparams[1] .* ones(length(θ) - Tar.extraparams)
+
+    if Tar.extraparams > 0
+        return (logpdf(MvNormal(zeros(Tar.extraparams), stdinv),
```
Vaibhavdixit02, line 175: `MvNormal` expects variance not stddev

AstitvaAggarwal: the Distributions.MvNormal() ends up creating the covariance by squaring the stddev.
```
julia> MvNormal([1,2],[1,3])
MvNormal{Int64, PDMats.PDiagMat{Int64, Vector{Int64}}, Vector{Int64}}(
dim: 2
μ: [1, 2]
Σ: [1 0; 0 9]
)
```

Vaibhavdixit02 (commented):
Since we are dropping using Tueing and sticking with AHMC you can remove the code for that and this can be made ready to be merged?

test/BPINN_Tests.jl
AstitvaAggarwal: yeah, corrected that. i still need to clean up the BPINN test file alot

AstitvaAggarwal:
Ill start working upon the PDE solver now, its implementation will be building upon the existing symbolic parser scheme. the BPINN ode solver is done.

xtalax:
Remember to add your tests to `test/runtests.jl`, you may also need to add it's group to `.github/workflows/CI.yml` if you created a new group for it

.github/workflows/CI.yml
```diff
@@ -13,14 +13,17 @@ jobs:
       fail-fast: false
       matrix:
         group:
+          #fixes 682
+          -"Bpinn ODE solver"
```
xtalax, line 17: this should be ODEBPINN with no quotes

AstitvaAggarwal: yeah my bad fixed it(ended up pushing it without changing)

.github/workflows/CI.yml
```diff
@@ -13,14 +13,17 @@ jobs:
       fail-fast: false
       matrix:
         group:
+          #fixes 682
+          -ODEBPINN
```
xtalax, line 17: `- ODEBPINN`

needs a space. One check runs for each group, not for each testset

src/advancedHMC_MCMC.jl
```diff
@@ -0,0 +1,423 @@
...
+    end
+end
+
+# physics loglikelihood over problem timespan
+function physloglikelihood(Tar::LogTargetDensity, θ)
+    f = Tar.prob.f
+    p = Tar.prob.p
+    dt = Tar.physdt
+    if isempty(Tar.dataset[end])
+        t = collect(Float64, Tar.prob.tspan[1]:dt:Tar.prob.tspan[2])
```
xtalax, line 196: replace all `Float64` with `eltype(\theta)` in case we're running on GPU

src/advancedHMC_MCMC.jl
```diff
@@ -0,0 +1,441 @@
...
+    if Adaptor != AdvancedHMC.NoAdaptation()
+        Adaptor(mma, ssa)
+    else
+        AdvancedHMC.NoAdaptation()
+    end
+end
+
+# dataset would be (x̂,t)
+# priors: pdf for W,b + pdf for ODE params
+function ahmc_bayesian_pinn_ode(prob::DiffEqBase.DEProblem, chain;
```
Vaibhavdixit02, line 321: @xtalax does this suit the API in NeuralPDE or should we change this?

Vaibhavdixit02 (approved):
LGTM, just curious about the API here being suitable which can also be addressed from another PR as per @xtalax preference

.github/workflows/CI.yml
```diff
@@ -13,14 +13,17 @@ jobs:
       fail-fast: false
       matrix:
         group:
+          #fixes 682
+          - ODEBPINN
+
```
ChrisRackauckas, line 18:
```suggestion
          - ODEBPINN
```

src/advancedHMC_MCMC.jl
```diff
@@ -0,0 +1,459 @@
...
+targetacceptancerate -> Target percentage(in decimal) of iterations in which the proposals were accepted(0.8 by default)
+Adaptor -> https://turinglang.org/AdvancedHMC.jl/stable/
+Integrator -> https://turinglang.org/AdvancedHMC.jl/stable/
+Metric -> https://turinglang.org/AdvancedHMC.jl/stable/
+
+## References  
+
+"""
+
+mutable struct LogTargetDensity{C, S, I}
```
ChrisRackauckas, line 79:
```suggestion
"""
mutable struct LogTargetDensity{C, S, I}
```

src/advancedHMC_MCMC.jl
```diff
@@ -0,0 +1,459 @@
+"""
```
ChrisRackauckas, line 1: this docstring needs to move to be right on top of the function, otherwise it won't work

src/advancedHMC_MCMC.jl
```diff
@@ -0,0 +1,459 @@
...
+
+"""
+
+mutable struct LogTargetDensity{C, S, I}
+    dim::Int
+    prob::DiffEqBase.DEProblem
+    chain::C
+    st::S
+    dataset::Vector{Vector{Float64}}
+    priors::Vector{Distribution}
```
ChrisRackauckas, line 85: This is going to be unstable: does it not effect performance?

src/advancedHMC_MCMC.jl
```diff
@@ -0,0 +1,459 @@
...
+Integrator -> https://turinglang.org/AdvancedHMC.jl/stable/
+Metric -> https://turinglang.org/AdvancedHMC.jl/stable/
+
+## References  
+
+"""
+
+mutable struct LogTargetDensity{C, S, I}
+    dim::Int
+    prob::DiffEqBase.DEProblem
```
ChrisRackauckas, line 81: Currently the code looks specialized to ODEs, is there a reason this isn't?

src/advancedHMC_MCMC.jl
```diff
@@ -0,0 +1,459 @@
...
+                     ode_params,
+                     t[i])
+                   for i in 1:length(out[1, :])]
+    else
+        physsol = [f(out[:, i],
+                     ode_params,
+                     t[i])
+                   for i in 1:length(out[1, :])]
+    end
+    physsol = hcat(physsol...)
```
ChrisRackauckas, line 238:
```suggestion
    physsol = reduce(hcat,physsol)
```

ChrisRackauckas:
Implementation-wise, I think it generally looks pretty good now. I left a few comments but those are pretty minor.

As for API, I do think that might need to change a little. I think this should be setup as just a `solve(prob::ODEProblem, alg::BNNODE(...))` with the details for the fitting process in the BNNODE. Now this would be a bit of a weird ODE solver since the solution cannot be easily represented as an ODESolution, but what we can do is add a `.original` field to the ODESolution (which exists in most other solution types already anyways so it's a reasonable extension) which then holds the `fh_mcmc_chain, fhsamples, fhstats` (which we can put in a BPINNStatistics struct or something), and then the "standard" solution computation can be done in advance to construct the `sol.u`. That would make all of this stuff (https://github.com/SciML/NeuralPDE.jl/pull/692/files#diff-f426df4f589f3ec64dfa582ee30371490608d20f836b399cbfebcb443e832c98R130-R147) just `isapprox(sol.u , ...)` which makes sense because most users will first want to know "what's the ODE solution"? Something that would be nice to the user would be to represent this as a MonteCarloMeasurements or Measurements type so that plotting and such automatically works with error bounds. As done in NNODE, the `t` would just be chosen to match the `saveat`. 

Then `ahmc_bayesian_pinn_ode` can be documented as a lower level API for those who really want to dive in, but `solve(prob::ODEProblem, alg::BNNODE(...))` would then work very naturally "just like any other ODE solver", with the extra information always available in `sol.original` but with the uncertainty information represented in the `u`. That API would be much nicer than the current post-solution analysis required in the tests.

That all said, creating that higher level API is simply calls to `ahmc_bayesian_pinn_ode` under the hood and then the post analysis that's exactly done in the tests right now, so I don't think doing this high level API will require any new code. It's just restructuring it to be a bit simpler.

src/advancedHMC_MCMC.jl
AstitvaAggarwal: Initially I was thinking of making this a PDE solver as well. But Ill change this as imo it would be better to implement that separately. Thanks

AstitvaAggarwal:
Thank you for the reviews, I'll create a seperate Issue, PR for the High level APIs for both the ODE solver and PDE solver together. Also ill move onto the PDE solver for now, unless you all would recommend finishing with the APIs first.

ChrisRackauckas:
Finishing the higher level ODE interface would be top priority. This functionality effectively won't exist until that piece is done.

src/advancedHMC_MCMC.jl
AstitvaAggarwal: ive been doing @time and @btime on the function on the merged struct version and the version you put this review (but uses prob::ODEProblem) the difference dosent change much. In fact on many runs the unstable version performs better.

PR #695, opened 2023-06-27
NNODE for DAE Problems

sdesai1287:
Initial changes to ode_solve, added NNDAE to get this solver working for DAE problems


sdesai1287:
Hi @ChrisRackauckas I wanted to check with you to see if the theory that I have added in is correct. I am writing out some tests to verify that it works and will add them in the test file shortly

sdesai1287:
I solved a bunch of bugs with this, but when I run this test file I get this long error that I do not really know how to solve. I believe the error starts at line 300 [line 300](https://github.com/sdesai1287/NeuralPDE.jl/blob/c56922c7f291dc07aa5b9b48cc29a5115510fbcb/src/ode_solve.jl#L300C2-L300C2)
Screen Shot 2023-06-27 at 1 28 41 AM

xtalax:
I think the problem is the argument signature that you have called `f` with. You have provided a path for the case where `f` is an out of place function like `f(u, p, t)`, but you must also provide a path for the case where it is in place like `f!(du, u, p, t)`, which commonly happens as it is a pattern for better speed in iterative solves.

sdesai1287:
Hi @xtalax I have tried a several things but nothing seems to work. However, I think NNODE itself doesnt work for in place problems, so modifying it for the DAE case for in place methods is probably not going to work either. I am trying to find a DAE problem to test that is out of place, but both the DAEProblems from the DAE Problem Library (from DiffEqProblemLibrary.jl), and their corresponding DAEFunctions are in place. If you can direct me to an out of place DAE problem to test (or some other solution for this issue), that would be greatly appreciated

ChrisRackauckas:
> I think the problem is the argument signature that you have called f with. You have provided a path for the case where f is an out of place function like f(u, p, t), but you must also provide a path for the case where it is in place like f!(du, u, p, t), which commonly happens as it is a pattern for better speed in iterative solves.

Kind of. DAEProblems are represented via `f(out,du,u,p,t)` and `f(du,u,p,t)`. See https://docs.sciml.ai/DiffEqDocs/stable/tutorials/dae_example/#Implicitly-Defined-Differential-Algebraic-Equations-(DAEs).

src/ode_solve.jl
```diff
@@ -252,6 +271,35 @@ function inner_loss(phi::ODEPhi{C, T, U}, f, autodiff::Bool, t::AbstractVector,
...
+                    p) where {C, T, U}
+    sum(abs2, ode_dfdx(phi, t, θ, autodiff) .- f(phi(t, θ), p, t))
+end
+
+function inner_loss_DAE(phi::ODEPhi{C, T, U}, f, autodiff::Bool, t::AbstractVector, θ,
+                    p) where {C, T, U}
+    out = Array(phi(t, θ))
+    arrt = Array(t)
+    dxdtguess = Array(ode_dfdx(phi, t, θ, autodiff))
+    sum(abs2, f(dxdtguess[:, i], p, arrt[i]) for i in 1:size(out, 2)) / length(t)
```
ChrisRackauckas, line 300: again missing the u

ChrisRackauckas:
The error is pointing to the fact that you're using the wrong function signature. The out of place DAE is 4 arguments, not 3.  It needs to know the current state, not just the derivative.

sdesai1287:
Hi @ChrisRackauckas I tried adding u like you said but it still gives me an error for u being undefined at line 302. Going to keep working on this

src/dae_problem_test.jl
```diff
@@ -0,0 +1,50 @@
...
+using NeuralPDE, Lux, Test, Statistics, Plots
+
+f = function (r, yp, y, p, tres)
+    r[1] = -0.04 * y[1] + 1.0e4 * y[2] * y[3]
+    r[2] = -r[1] - 3.0e7 * y[2] * y[2] - yp[2]
+    r[1] -= yp[1]
+    r[3] = y[1] + y[2] + y[3] - 1.0
+end
+u0 = [1.0, 0, 0]
+du0 = [-0.04, 0.04, 0.0]
```
ChrisRackauckas, line 11:
```suggestion
f = function (yp, y, p, tres)
    [-0.04 * y[1] + 1.0e4 * y[2] * y[3] - yp[1],
     -r[1] - 3.0e7 * y[2] * y[2] - yp[2],
      y[1] + y[2] + y[3] - 1.0]
end
u0 = [1.0, 0, 0]
du0 = [-0.04, 0.04, 0.0]
```

src/ode_solve.jl
```diff
@@ -252,6 +271,37 @@ function inner_loss(phi::ODEPhi{C, T, U}, f, autodiff::Bool, t::AbstractVector,
...
 
+"""
+L2 inner loss for DAEProblems
+"""
+
+function inner_loss_DAE end
+
+function inner_loss_DAE(phi::ODEPhi{C, T, U}, f, autodiff::Bool, t::Number, θ,
+                    p, u) where {C, T, U <: Number}
+    sum(abs2, ode_dfdx(phi, t, θ, autodiff) - f(phi(t, θ), u, p, t))
```
ChrisRackauckas, line 282: u is phi. This is completely off. Please try writing this on pen and paper first.

sdesai1287: will do, got some stuff mixed up

src/dae_problem_test.jl
sdesai1287: I think this is fixed

sdesai1287:
I made some fixes to this but I still have the following error: 
Screen Shot 2023-07-26 at 7 27 23 PM

I think my new problem definition is a valid out of place DAE problem. My primary confusion lies with when to use `phi` vs `phi(t, θ)` vs `Array(phi(t, θ))` when generating the loss function. The error occurs [here](https://github.com/sdesai1287/NeuralPDE.jl/blob/685a688ca8ea4e3e3920065130a36664d8967e6b/src/ode_solve.jl#L30)

sdesai1287:
> what's your MWE?

Does that mean minimum working example? If so, I dont really have one. I tried using a simpler chain but it solved nothing, and I used a problem directly from the DAEProblemLibrary (and from you) so I doubt that is the issue either. Not too sure how to resolve this

src/ode_solve.jl
```diff
@@ -252,6 +271,37 @@ function inner_loss(phi::ODEPhi{C, T, U}, f, autodiff::Bool, t::AbstractVector,
...
 
+"""
+L2 inner loss for DAEProblems
+"""
+
+function inner_loss_DAE end
+
+function inner_loss_DAE(phi::ODEPhi{C, T, U}, f, autodiff::Bool, t::Number, θ,
+                    p) where {C, T, U <: Number}
+    sum(abs2, ode_dfdx(phi, t, θ, autodiff) - f(ode_dfdx(phi, t, θ, autodiff), phi, p, t))
```
ChrisRackauckas, line 282: this doesn't make sense.

ChrisRackauckas: the output of a dae function is the residual, so there's nothing to subtract

src/ode_solve.jl
```diff
@@ -252,6 +271,37 @@ function inner_loss(phi::ODEPhi{C, T, U}, f, autodiff::Bool, t::AbstractVector,
...
+end
+
+function inner_loss_DAE(phi::ODEPhi{C, T, U}, f, autodiff::Bool, t::AbstractVector, θ,
+                    p) where {C, T, U <: Number}
+    out = phi(t, θ)
+    dxdtguess = Array(ode_dfdx(phi, t, θ, autodiff))
+    sum(abs2, f(dxdtguess[i], phi, p, t[i]) for i in 1:size(out, 2)) / length(t)
```
ChrisRackauckas, line 289: you can't pass `phi` to `f`, it's a function.

ChrisRackauckas, line 289:
```suggestion
    sum(abs2, f(dxdtguess[i], out, p, t[i]) for i in 1:size(out, 2)) / length(t)
```

ChrisRackauckas: but you need out at each `t`

src/ode_solve.jl
```diff
@@ -252,6 +271,37 @@ function inner_loss(phi::ODEPhi{C, T, U}, f, autodiff::Bool, t::AbstractVector,
...
+                    p) where {C, T, U}
+    sum(abs2, ode_dfdx(phi, t, θ, autodiff) .- f(ode_dfdx(phi, t, θ, autodiff), phi, p, t))
+end
+
+function inner_loss_DAE(phi::ODEPhi{C, T, U}, f, autodiff::Bool, t::AbstractVector, θ,
+                    p) where {C, T, U}
+    out = Array(phi(t, θ))
+    arrt = Array(t)
+    dxdtguess = Array(ode_dfdx(phi, t, θ, autodiff))
+    sum(abs2, f(dxdtguess[:, i], phi, p, arrt[i]) for i in 1:size(out, 2)) / length(t)
```
ChrisRackauckas, line 302: These two have the same errors as above.

ChrisRackauckas:
What example are you working with to test? It should point exactly to the 4 issues I pointed out above.

src/ode_solve.jl
sdesai1287: do I need `out` or `out[t]`?

src/ode_solve.jl
sdesai1287: Do I just make that line `sum(abs2, ode_dfdx(phi, t, θ, autodiff))` then?

sdesai1287:
> What example are you working with to test? It should point exactly to the 4 issues I pointed out above.

Here is my test code, which should also be in the commit
```
using Optimisers, OptimizationOptimisers, Sundials
using NeuralPDE, Lux, Test, Statistics, Plots

f = function (yp, y, p, tres)
    [-0.04 * y[1] + 1.0e4 * y[2] * y[3] - yp[1],
     -(-0.04 * y[1] + 1.0e4 * y[2] * y[3]) - 3.0e7 * y[2] * y[2] - yp[2],
      y[1] + y[2] + y[3] - 1.0]
end
u0 = [1.0, 0, 0]
du0 = [-0.04, 0.04, 0.0]

prob_oop = DAEProblem{false}(f, du0, u0, (0.0, 100000.0))
true_sol = solve(prob_oop, IDA(), saveat = 0.01)

func = Lux.σ
N = 12
chain = Lux.Chain(Lux.Dense(1, N, func), Lux.Dense(N, N, func), Lux.Dense(N, N, func),
                    Lux.Dense(N, N, func), Lux.Dense(N, length(u0)))

opt = Optimisers.Adam(0.01)
dx = 0.05
alg = NeuralPDE.NNDAE(chain, opt, autodiff = false, strategy = NeuralPDE.GridTraining(dx))
sol = solve(prob_oop, alg, verbose=true, maxiters = 100000, saveat = 0.01)
```
This should be the Robertson biochemical reactions in DAE form. I've checked this over several times, so I am pretty sure this is not the problem, but there could certainly be something I missed

src/ode_solve.jl
ChrisRackauckas: No, you need to evaluate f... ` sum(abs2,f(ode_dfdx(phi, t, θ, autodiff), phi(t, θ), p, t)))`

sdesai1287:
@ChrisRackauckas @xtalax I implemented the new edits that you suggested and I am pretty sure that the loss function generation is working properly now. However, the last step [here](https://github.com/sdesai1287/NeuralPDE.jl/blob/fe9274f8c55d711d389cfacc8ec933c61fffded0/src/ode_solve.jl#L721C4-L721C4) of calling Julia's `solve` method within my `DAEProblem` solver runs infinitely without printing anything as part of the callback function. I think I am now very close on this, and any tips with debugging this would be greatly appreciated

ChrisRackauckas:
Done in https://github.com/SciML/NeuralPDE.jl/pull/790

Issue #644, opened 2023-02-15
Have a training strategy for PreDefined Points


sathvikbhagavan:
For training PINNs, can we have a strategy where the user can sample points themselves and feed it into training? This way, they can handle if the system requires very specific discretization

cc: @ChrisRackauckas

ChrisRackauckas:
For NNODE, tstops is would be a way to do this. But for general training yes this could be a good idea for say precomputing shocks

sathvikbhagavan:
>For NNODE, tstops is would be a way to do this

Can you explain how tstops would work? IIUC, those are extra timepoints the solver steps into.


I was thinking more in terms of a strategy having an array of these generated points which we use in `generate_loss` function to compute loss like: https://github.com/SciML/NeuralPDE.jl/blob/master/src/ode_solve.jl#L272

sathvikbhagavan:
@ChrisRackauckas can I close this as https://github.com/SciML/NeuralPDE.jl/pull/723 added this feature?

PR #722, opened 2023-08-17
Given points training

sdesai1287:
#644 Initial commit for GivenPointsTraining. It works on the example I used but may be useful to test more cases. I can also remove the DAE problem stuff if necessary


ChrisRackauckas:
please do this on a fresh branch

PR #723, opened 2023-08-17
Incorporating given points into training

sdesai1287:
#644 This is working for the given example


ChrisRackauckas:
As mentioned in https://github.com/SciML/NeuralPDE.jl/issues/644, tstops would be the right way to do this. Basically, `tstops` is a common arg https://docs.sciml.ai/DiffEqDocs/stable/basics/common_solver_opts/ that forces a solver to take into account those points. Those would be your "given points" in this sense for an NNODE. So what we'd want to do is take any strategy and then augment the points from that strategy with the `tstops` vector at every training stage.

This would be a lot more useful since only using the given points is likely never a good idea. The points should be changing and randomized so that just won't be possible with given points. But what we can do is do for example a QuadratureStrategy but, knowing the point of a discontinuity, add a few training points around this point that we know is going to be difficult.

sdesai1287:
Ok got it, we do not want a separate training strategy, we want to add this functionality to each training strategy. I've written up some stuff so far, but there are some issues currently. For `WeightedIntervalTraining`, it is giving me a `stackoverflowerror` and for StochasticTraining it is telling me I cannot mutate an array (the points array to add the given points) because that is a limitation of Zygote has. I am currently reading up about Zygote to see if I can solve this issue

ChrisRackauckas:
You don't need to mutate or add to the loss function code of any strategy. Just append the values by adding the loss.

sdesai1287:
I think this is what you are referring to? I tested it with 3 training strategies and it turned a failing solver into a successful solver in all 3 cases

src/NeuralPDE.jl
```diff
@@ -56,7 +56,7 @@ export NNODE, TerminalPDEProblem, NNPDEHan, NNPDENS, NNRODE,
        KolmogorovParamDomain, NNParamKolmogorov,
        PhysicsInformedNN, discretize,
        GridTraining, StochasticTraining, QuadratureTraining, QuasiRandomTraining,
-       WeightedIntervalTraining,
+       WeightedIntervalTraining, GivenPointsTraining,
```
ChrisRackauckas, line 59: not needed

src/ode_solve.jl
```diff
@@ -330,10 +331,24 @@ function generate_loss(strategy::WeightedIntervalTraining, phi, f, autodiff::Boo
...
 end
 
+function generate_additional_points_loss(tstops, phi, f, autodiff::Bool, tspan, p, batch)
+
+    # sum(abs2,inner_loss(t,θ) for t in ts) but Zygote generators are broken
+    function loss(θ, _)
+        if batch
+            sum(abs2, inner_loss(phi, f, autodiff, tstops, θ, p))
+        else
+            sum(abs2, [inner_loss(phi, f, autodiff, t, θ, p) for t in tstops])
```
ChrisRackauckas, line 341: tstops is not defined in this function

src/ode_solve.jl
```diff
@@ -330,10 +331,24 @@ function generate_loss(strategy::WeightedIntervalTraining, phi, f, autodiff::Boo
     return loss
 end
 
+function generate_additional_points_loss(tstops, phi, f, autodiff::Bool, tspan, p, batch)
+
+    # sum(abs2,inner_loss(t,θ) for t in ts) but Zygote generators are broken
+    function loss(θ, _)
+        if batch
+            sum(abs2, inner_loss(phi, f, autodiff, tstops, θ, p))
```
ChrisRackauckas, line 339: why is tstops not used here?

src/ode_solve.jl
```diff
@@ -330,10 +331,24 @@ function generate_loss(strategy::WeightedIntervalTraining, phi, f, autodiff::Boo
     return loss
 end
 
+function generate_additional_points_loss(tstops, phi, f, autodiff::Bool, tspan, p, batch)
```
ChrisRackauckas, line 334:
```suggestion
function generate_additional_points_loss(phi, f, autodiff::Bool, tstops, p, batch)
```

Follow the normal ordering you see elsewhere. No reason to just mix things up.

src/ode_solve.jl
```diff
@@ -429,9 +445,13 @@ function DiffEqBase.__solve(prob::DiffEqBase.AbstractODEProblem,
     function total_loss(θ, _)
         L2_loss = inner_f(θ, phi)
         if !(additional_loss isa Nothing)
-            return additional_loss(phi, θ) + L2_loss
+            L2_loss = L2_loss + additional_loss(phi, θ) 
+        end
+        if !(tstops isa Nothing)
+            addedPointsLossFunc = generate_additional_points_loss(tstops, phi, f, autodiff, tspan, p, batch)
```
ChrisRackauckas, line 451:
```suggestion
            addedPointsLossFunc = generate_additional_points_loss(phi, f, autodiff, tstops, p, batch) 
```

src/ode_solve.jl
ChrisRackauckas, line 334: Also this needs a better name. `additional_loss` is something different, and this is about `tstops`, not `additional_loss`, so call it `evaluate_tstops_loss`

test/given_points_training_test.jl
```diff
@@ -0,0 +1,70 @@
+
+using OrdinaryDiffEq, OptimizationPolyalgorithms, Lux, OptimizationOptimJL, Test, Statistics, Plots, Optimisers
```
ChrisRackauckas, line 2:
```suggestion
using OrdinaryDiffEq, OptimizationPolyalgorithms, Lux, OptimizationOptimJL, Test, Statistics, Plots, Optimisers
```

test/given_points_training_test.jl
```diff
@@ -0,0 +1,70 @@
...
+
```
ChrisRackauckas, line 3: This file is not in `runtests.jl` so it's not going to run.

test/given_points_training_test.jl
```diff
@@ -0,0 +1,70 @@
...
+function fu(u, p, t)
```
ChrisRackauckas, line 4: This file also should be renamed `NNODE_tstops_test.jl`

ChrisRackauckas:
This looks close, just needs the clean ups mentioned.

test/given_points_training_test.jl
sdesai1287: Fixed

src/ode_solve.jl
sdesai1287: Fixed

src/NeuralPDE.jl
sdesai1287: Fixed

test/given_points_training_test.jl
sdesai1287: Fixed

src/ode_solve.jl
sdesai1287: Fixed

src/ode_solve.jl
```diff
@@ -29,7 +29,8 @@ of the physics-informed neural network which is used as a solver for a standard
 
 ## Example
 
-```julia
+```juliap = [1.5, 1.0, 3.0, 1.0]
```
ChrisRackauckas, line 32: This looks like an error

ChrisRackauckas:
Why make a separate test group? Does this one really take half an hour?

sdesai1287:
> Why make a separate test group? Does this one really take half an hour?

It does not take half an hour, I just fixed that now and just made it run as part of 'All'

src/ode_solve.jl
sdesai1287: Not sure where that came from but its fixed now, sorry about that

ChrisRackauckas:
All is not a test group. Please put it with the other nnode tests

sdesai1287:
> All is not a test group. Please put it with the other nnode tests

Fixed now, sorry for the confusion

ChrisRackauckas:
Test fails

sdesai1287:
Weird, it was working on my end when I ran it, but I made some edits and accounted for the relative size of the tstops entry when calculating the total loss. This works much more consistently now. I am not sure if this is exactly what we were looking for with this idea but hopefully it does what we want

ChrisRackauckas:
No, read the CI... it was missing test imports.

ChrisRackauckas:
No, it's saying `OptimizationPolyalgorithms` isn't a test dependency.

sdesai1287:
> No, it's saying `OptimizationPolyalgorithms` isn't a test dependency.

Ok I removed it from the file, hopefully everything is good now

ChrisRackauckas:
https://buildkite.com/julialang/neuralpde-dot-jl/builds/1754#018a164a-4217-4906-bb48-31b96d18ed21/6-13280

You never loaded it.

sdesai1287:
> https://buildkite.com/julialang/neuralpde-dot-jl/builds/1754#018a164a-4217-4906-bb48-31b96d18ed21/6-13280
> 
> You never loaded it.

oops, loaded now

sdesai1287:
Screen Shot 2023-08-21 at 10 46 38 PM
Screen Shot 2023-08-21 at 10 48 33 PM
not really sure where these are coming from in the buildkite

Issue #719, opened 2023-08-17
API for BNNODE.


AstitvaAggarwal:
extends #682 ahmc_bayesian_pinn_ode() function call.

AstitvaAggarwal:
Also added support for all current training strategies at the lower level. But I've noticed that stochastic, weighted interval training do not fair well, so new strategies can be added.

PR #724, opened 2023-08-18
[WIP] API for BNNODE

AstitvaAggarwal:
for issue #719


xtalax:
Looking good, I wanna see your tests passing and documentation with an example then lgtm

AstitvaAggarwal:
@Vaibhavdixit02 @xtalax on my local machine all tests pass(v1.9.3) but for the v1 and v1.6 marginal errors crop up, what should I do? Using earlier versions affects performance. Also the v1.6 logging CI tests gives the statistics compat error as the AHMC version I'm using(NeuralPDE allows v0.5) works only with the latest Julia versions(refer https://julialang.slack.com/archives/CN04R7WKE/p1693333431525009?thread_ts=1693333431.525009&cid=CN04R7WKE)

Vaibhavdixit02:
Maybe we can drop 1.6 for this test group (it wouldn't be straightforward though). I don't recall why the compat was needed, is it possible to get rid of that?

Vaibhavdixit02:
Looking at test failures the values seem to be in the right ballpark so my guess is this is just a numerical difference between your local runs and the CI machine. You are probably not on the same OS (linux)? Relax the tolerances of the tests

AstitvaAggarwal:
> Maybe we can drop 1.6 for this test group (it wouldn't be straightforward though). I don't recall why the compat was needed, is it possible to get rid of that?

 was done [here](https://github.com/SciML/NeuralPDE.jl/pull/718) by @ChrisRackauckas

AstitvaAggarwal:
whoa! after [this](https://github.com/SciML/NeuralPDE.jl/actions/runs/6019722719/job/16329939664) update in Build to the latest Julia version the tests are almost exactly how i wanted them to be!

AstitvaAggarwal:
@Vaibhavdixit02 ive added the extra Likelihood term i was talking about, if you want to run it with the regular tests just uncomment line 53 in the file "advancedHMC_MCMC.jl", ive tried it works!

src/advancedHMC_MCMC.jl
```diff
@@ -190,6 +192,66 @@ function physloglikelihood(Tar::LogTargetDensity, θ)
...
+        else
+            physsol = [f(û[:, i],
+                ode_params,
+                t[i])
+                       for i in 1:length(û[1, :])]
+        end
+        #form of NN output matrix output dim x n
+        deri_physsol = reduce(hcat, physsol)
+
+        #   OG deriv(basically gradient matching in case of an ODEFunction)
```
Vaibhavdixit02, line 226: I would have expected the loss to be against the solution of the DE instead of the derivative matching here

AstitvaAggarwal: that is already happening, in L2loss

src/advancedHMC_MCMC.jl
```diff
@@ -181,18 +184,79 @@ function physloglikelihood(Tar::LogTargetDensity, θ)
     for i in 1:length(Tar.prob.u0)
         # can add phystd[i] for u[i]
         physlogprob += logpdf(MvNormal(nnsol[i, :],
-                                       LinearAlgebra.Diagonal(map(abs2,
-                                                                  Tar.phystd[i] .*
-                                                                  ones(length(physsol[i, :]))))),
-                              physsol[i, :])
+                LinearAlgebra.Diagonal(map(abs2,
```
Vaibhavdixit02, line 187: It would be good to split this out into multiple lines, it's a bit unreadable right now

src/advancedHMC_MCMC.jl
Vaibhavdixit02: The conclusion was that it won't be efficient to simulate PDEs and do this so we'll stick with gradient matching

AstitvaAggarwal: We cannot solve the pde at current params, But we can substitute the data points(observed solutions) into the pde at current params So now the only moving thing params is now bound. We could solve the pde at individual param values during sampling But that is being done already as nn vs dataset output is basically.

src/advancedHMC_MCMC.jl
AstitvaAggarwal: done

test/BPINN_Tests.jl
```diff
@@ -189,10 +191,10 @@ meanscurve2 = prob.u0 .+ (t .- prob.tspan[1]) .* luxmean
 @test abs(p - mean([fhsamples1[i][23] for i in 2000:2500])) < abs(0.2 * p)
 
 #---------------------- solve() call 
-@test mean(abs.(x̂1 .- sol2flux.ensemblesol)) < 5e-1
-@test mean(abs.(physsol1_1 .- sol2flux.ensemblesol)) < 5e-1
-@test mean(abs.(x̂1 .- sol2lux.ensemblesol)) < 6e-2
-@test mean(abs.(physsol1_1 .- sol2lux.ensemblesol)) < 6e-2
+@test mean(abs.(x̂1 .- sol2flux.ensemblesol[1])) < 5e-1
+@test mean(abs.(physsol1_1 .- sol2flux.ensemblesol[1])) < 5e-1
+@test mean(abs.(x̂1 .- sol2lux.ensemblesol[1])) < 5e-1
```
Vaibhavdixit02, line 196: Does the error of the solution become worse here or is it stochastic?

AstitvaAggarwal: Stochastic nature

xtalax:
@Vaibhavdixit02 lgtm?

Vaibhavdixit02:
A couple of minor things, but they can be addressed in a different PR since this one's quite big already.

AstitvaAggarwal:
Thank you 🎉

PR #727, opened 2023-08-28
Rename `samples` field of `WeightedIntervalTraining` to `points` and clean `tstops_loss` logic

sathvikbhagavan:
Renaming would be consistent with other strategies `StochasticTraining`, `QuasiRandomTraining` etc. as it contains the number of points.

Cleaning up `tstops_loss` to avoid redundant function evaluation.

### DIFF
--- a/src/training_strategies.jl
+++ b/src/training_strategies.jl
@@ -320,7 +320,7 @@ This training strategy can only be used with ODEs (`NNODE`).
 """
 struct WeightedIntervalTraining{T} <: AbstractTrainingStrategy
     weights::Vector{T}
-    points::Int64
+    points::Int
 end
 
 function WeightedIntervalTraining(weights, points)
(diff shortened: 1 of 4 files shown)


src/training_strategies.jl
```diff
@@ -312,19 +312,19 @@ such that the total number of sampled points is equivalent to the given samples
...
+* `points`: the total number of samples that we want, across the entire time span
 
 ## Limitations
 
 This training strategy can only be used with ODEs (`NNODE`).
 """
 struct WeightedIntervalTraining{T} <: AbstractTrainingStrategy
     weights::Vector{T}
-    samples::Int
+    points::Int64
```
ChrisRackauckas, line 323:
```suggestion
    points::Int
```

Never use `Int64` unless it's required.

PR #728, opened 2023-09-02
Change `iteration` type from size 1 `Vector{Int}` to `Ref{Int}`

### LINKED ISSUE #552: Iteration should be a `Ref{Int}`, not a size 1 `Vector{Int}`
This is just safer, @zoemcc

pri1311:
Closes/Fixes #552


avik-pal:
handled in #882

PR #734, opened 2023-09-12
energies: adding energy loss functions

drsk0:
This commit adds energy loss functions. These are loss functions analogous to those obtained from PDE equations or boundary equations, but the energy integrand can be given explicitly in symbolic form.

The energies are maintained in a separate field from the PDE equations so that they can be weighted differently in adaptive loss functions.

I am happy to extend the docs once the general direction of this PR is approved.

This PR depends on https://github.com/SciML/ModelingToolkit.jl/pull/2256.


ChrisRackauckas:
> The energies are maintained in a separate field from the PDE equations so that they can be weighted differently in adaptive loss functions.

But the adaptive loss has a vector that it's adapting, so is this really any different than including the energy functions as part of the BCs?

drsk0:
You could also have the energies as part of the BC's or PDE equations. I thought there might be a benefit to keeping them separate to be able to treat them differently if needed.

ChrisRackauckas:
Can you show an example which demonstrates a benefit?

drsk0:
A trivial example that comes to mind is that I'd like to have them logged separately from the BC's. I consider the energy in a PDE problem a fairly different concept than the boundary conditions.

The other solution would be to allow for energies in BCs and equations. I could imagine a bit more subtle bugs when you accidentally create an energy instead of an equation in this case.

PR #741, opened 2023-09-27
changes from final reviews, improved Docs

### LINKED ISSUE #737: fix NUTS kwargs handling and Default ode params vector
changes from final reviews on #724

AstitvaAggarwal:
fixes #737

### DIFF
--- a/docs/src/examples/Lotka_Volterra_BPINNs.md
+++ b/docs/src/examples/Lotka_Volterra_BPINNs.md
@@ -60,7 +60,7 @@ To make the example more realistic we add random normally distributed noise to t
 
 
 ```julia
-# Dataset creation for parameter estimation(30 percent noise)
+# Dataset creation for parameter estimation (30% noise)
 time = solution.t
 u = hcat(solution.u...)
 x = u[1, :] + (0.3 .*u[1, :]).*randn(length(u[1, :]))
--- a/docs/src/examples/Lotka_Volterra_BPINNs.md
+++ b/docs/src/examples/Lotka_Volterra_BPINNs.md
@@ -23,9 +23,9 @@ We then solve the equations and estimate the parameters of the model with priors
 
 And also solve the equations for the constructed ODEProblem's provided ideal `p` values using a Lux.jl Neural Network, chain_lux.
 
-```julia 
-using NeuralPDE, Flux, Lux, Plots, StatsPlots, OrdinaryDiffEq
-
+```julia
+using NeuralPDE, Flux, Lux, Plots, StatsPlots, OrdinaryDiffEq, Distributions 
+ 
 function lotka_volterra(u, p, t)
     # Model parameters.
     α, β, γ, δ = p
@@ -62,7 +62,7 @@ To make the example more realistic we add random normally distributed noise to t
 ```julia
 # Dataset creation for parameter estimation(30 percent noise)
 time = solution.t
-u = hcat(solution.u...) 
+u = hcat(solution.u...)
 x = u[1, :] + (0.3 .*u[1, :]).*randn(length(u[1, :]))
 y = u[2, :] + (0.3 .*u[2, :]).*randn(length(u[2, :]))
 dataset = [x, y, time]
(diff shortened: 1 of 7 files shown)


AstitvaAggarwal:
@xtalax @Vaibhavdixit02 this can be merged once all checks pass.

src/BPINN_ode.jl
```diff
@@ -99,48 +99,38 @@ struct BNNODE{C, K, ST <: Union{Nothing, AbstractTrainingStrategy}, IT, A, M,
...
-            Adaptor, targetacceptancerate,
-            Metric, jitter_rate, tempering_rate,
-            max_depth, Δ_max, n_leapfrog,
-            δ, λ, progress, verbose)
-    end
+end
+function BNNODE(chain, Kernel = HMC; strategy = nothing,
+    draw_samples = 2000, priorsNNw = (0.0, 2.0), param = nothing,
+    l2std = [0.05], phystd = [0.05], dataset = [nothing], physdt = 1 / 20.0,
+    n_leapfrog = 20, NUTSkwargs = (δ = 0.65, max_depth = 10, Δ_max = 1000),
```
Vaibhavdixit02, line 121: Either one of NUTSkwards or HMCDAkwargs will be used at a time right? Make this a single arg `mcmc_kwargs` or something like that and then depending on the typeof the mcmc method you have you can use them

AstitvaAggarwal: fixed, even the lower level call is more streamlined now

src/BPINN_ode.jl
```diff
@@ -99,48 +99,33 @@ struct BNNODE{C, K, ST <: Union{Nothing, AbstractTrainingStrategy}, IT, A, M,
...
-            physdt, nchains, autodiff, Integrator,
-            Adaptor, targetacceptancerate,
-            Metric, jitter_rate, tempering_rate,
-            max_depth, Δ_max, n_leapfrog,
-            δ, λ, progress, verbose)
-    end
+end
+function BNNODE(chain, Kernel = HMC; strategy = nothing, draw_samples = 2000,
+    priorsNNw = (0.0, 2.0), param = nothing, l2std = [0.05], phystd = [0.05],
+    dataset = [nothing], physdt = 1 / 20.0, MCMCargs = (n_leapfrog = 30), nchains = 1,
```
Vaibhavdixit02, line 118: This is not a namedtuple, also why not put everything (Integrator, Adaptor, Metric, target acceptance etc) into this? It should be documented in the docstring but the arguments can be reduced here. Also make the name MCMCkwargs instead of args since we expect a namedtuple and they are all kwargs to the downstream functions

AstitvaAggarwal: in the case of NUTS and HMCDA this becomes a NamedTuple, ive let this become an Int in the case of HMC as the format of input(MCMCargs) is similar to case of NUTS and HMCDA for the user. For making this a NamedTuple I would have had to make it MCMCargs = (n_leapfrog=30,)  instead of MCMCargs = (n_leapfrog=30) ,that extra comma can be unintuitive for case of HMC.

Vaibhavdixit02: >(n_leapfrog=30,) instead of MCMCargs = (n_leapfrog=30) ,that extra comma can be unintuitive for case of HMC.

I think most of our users are usually familiar with namedtuple so that should be fine, it makes sense to keep the type the same across the sampling algs.

>Also should i put integrator/Adaptor related args in seperate Nametuples(as they are seperate entities from the sampling alg)?

Yeah, that should be fine as well. If the fields you need to pass are not going to vary too much between the methods make it a struct.

xtalax:
passing, lgtm

docs/src/examples/Lotka_Volterra_BPINNs.md
```diff
@@ -58,67 +60,49 @@ To make the example more realistic we add random normally distributed noise to t
 
 
 ```julia
-# Dataset creation for parameter estimation
+# Dataset creation for parameter estimation(30 percent noise)
```
Vaibhavdixit02, line 63:
```suggestion
# Dataset creation for parameter estimation (30% noise)
```

Vaibhavdixit02:
Lgtm :+1:

PR #744, opened 2023-10-01
Fix compat of IntegralsCubature to be 0.2.2

sathvikbhagavan:
This is because 0.2.3 is causing the docs to break. Example build: https://github.com/SciML/NeuralPDE.jl/actions/runs/6366153358/job/17283653595?pr=742


ChrisRackauckas:
Integrated into https://github.com/SciML/NeuralPDE.jl/pull/743

Issue #712, opened 2023-08-07
Bayesian Inference for PDE's


AstitvaAggarwal:
1. Built over current Symbolic parsing system for PhysicsInformedNN (symbolic_discretize, PINNRepresentation)
2. similar to #682 in implementation needs

Issue #205, opened 2020-12-17
Bayesian Neural PDEs


ChrisRackauckas:
See https://arxiv.org/pdf/2006.04228.pdf as a guiding document. We can essentially have the NNBayesianPDE which is the same as the other but spits out a Bayesian estimation problem a la https://turing.ml/dev/tutorials/3-bayesnn/ with ADVI and use that to train with uncertainties. This would be a great GSoC project.

asdasdqwdqwfsdf:
@ChrisRackauckas 

What is the best reference/**theoretical** article for "NeuralPDE.jl", are we still using collocation points in "NeuralPDE.jl". namely this article [https://www.sciencedirect.com/science/article/pii/S0021999118307125]:

> Raissi, Maziar, Paris Perdikaris, and George E. Karniadakis. "Physics-informed neural networks: A deep learning framework for solving forward and inverse problems involving nonlinear partial differential equations." Journal of Computational Physics 378 (2019): 686-707.

ChrisRackauckas:
That is the right article. That said, this question is completely off topic for this thread so please do not ask it in a thread that's supposed to be about Bayesian versions.

asdasdqwdqwfsdf:
In Raissi et al. [1] article, **the collocation points** are used in their PINNs applications.

But **no information about collocation points** is found in "NeuralPDE.jl" and its applications.

[1]. https://www.sciencedirect.com/science/article/pii/S0021999118307125

ChrisRackauckas:
This is still completely off topic for this thread. Please do not abuse threads.

>But no information about collocation points is found in "NeuralPDE.jl" and its applications.

It is in the manual as the training strategies. There is an entire section written about this: https://neuralpde.sciml.ai/dev/solvers/pinn/#Training-strategy

ChrisRackauckas:
Another relevant paper: https://arxiv.org/abs/2003.06097

PR #745, opened 2023-10-01
BPINN PDE solver

AstitvaAggarwal:
for #712 and #205


src/discretize.jl
```diff
@@ -586,87 +586,150 @@ function SciMLBase.symbolic_discretize(pde_system::PDESystem,
...
+                        if (phi isa Vector && phi[1].f isa Optimisers.Restructure) ||
+                           (!(phi isa Vector) && phi.f isa Optimisers.Restructure)
+                            # Isa Flux Chain
+                            θ[1:(end - length(default_p))],
+                            θ[(end - length(default_p) + 1):end]
+                        else
+                            θ.depvar, θ.p
+                        end
                     else
-                        θ.depvar, θ.p
```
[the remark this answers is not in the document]
AstitvaAggarwal: i think ive fixed it now, thanks

ChrisRackauckas:
The docs successfully built on master, so I put this on top of a successful master.

ChrisRackauckas:
That looks like a legitimate test failure. You may want to split BPINN PDE to a new test group.

src/BPINN_ode.jl
```diff
@@ -148,21 +148,36 @@ end
 BPINN Solution contains the original solution from AdvancedHMC.jl sampling(BPINNstats contains fields related to that)
 > ensemblesol is the Probabilistic Estimate(MonteCarloMeasurements.jl Particles type) of Ensemble solution from All Neural Network's(made using all sampled parameters) output's.
 > estimated_nn_params - Probabilistic Estimate of NN params from sampled weights,biases
-> estimated_ode_params - Probabilistic Estimate of ODE params from sampled unknown ode paramters
+> estimated_de_params - Probabilistic Estimate of DE params from sampled unknown de paramters
 """
 struct BPINNsolution{O <: BPINNstats, E,
-    NP <: Vector{<:MonteCarloMeasurements.Particles{<:Float64}},
+    NP <:
+    Union{
```
Vaibhavdixit02, line 155: Move these into a const at the top level file.

Vaibhavdixit02: This is too complicated actually, lets make it a parameteric type instead

src/BPINN_ode.jl
```diff
@@ -148,21 +148,36 @@ end
...
-        Vector{<:MonteCarloMeasurements.Particles{<:Float64}}}}
+        Vector{<:MonteCarloMeasurements.Particles{<:Float64}}}, P}
     original::O
     ensemblesol::E
     estimated_nn_params::NP
-    estimated_ode_params::OP
-
-    function BPINNsolution(original, ensemblesol, estimated_nn_params, estimated_ode_params)
+    estimated_de_params::OP
+    points::P
```
Vaibhavdixit02, line 171: What are the points here?

src/PDE_BPINN.jl
```diff
@@ -0,0 +1,497 @@
...
+    dim::Int64
+    strategy::ST
+    dataset::D
+    priors::P
+    allstd::Vector{Vector{Float64}}
+    names::Tuple
+    extraparams::Int
+    init_params::I
+    full_loglikelihood::F
+    Phi::PH
```
Vaibhavdixit02, line 18: use unicode for such variable names

src/PDE_BPINN.jl
```diff
@@ -0,0 +1,497 @@
...
+function setparameters(Tar::PDELogTargetDensity, θ)
+    names = Tar.names
+    ps_new = θ[1:(end - Tar.extraparams)]
+    ps = Tar.init_params
+
+    if (ps[names[1]] isa ComponentArrays.ComponentVector)
+        # multioutput case for Lux chains, for each depvar ps would contain Lux ComponentVectors
+        # which we use for mapping current ahmc sampled vector of parameters onto NNs
+        i = 0
+        Luxparams = []
```
Vaibhavdixit02, line 86: type it?

src/BPINN_ode.jl
AstitvaAggarwal: The Domain points got using the domains and saveat vector for grid spacings for corresponding Var, at which the ensemble solution is evaluated

Vaibhavdixit02: Let's make the name more descriptiove, timepoints?

src/BPINN_ode.jl
```diff
@@ -148,21 +148,24 @@ end
 BPINN Solution contains the original solution from AdvancedHMC.jl sampling(BPINNstats contains fields related to that)
 > ensemblesol is the Probabilistic Estimate(MonteCarloMeasurements.jl Particles type) of Ensemble solution from All Neural Network's(made using all sampled parameters) output's.
 > estimated_nn_params - Probabilistic Estimate of NN params from sampled weights,biases
-> estimated_ode_params - Probabilistic Estimate of ODE params from sampled unknown ode paramters
+> estimated_de_params - Probabilistic Estimate of DE params from sampled unknown de paramters
```
Vaibhavdixit02, line 151:
```suggestion
> estimated_de_params - Probabilistic Estimate of DE params from sampled unknown DE parameters
```

src/advancedHMC_MCMC.jl
```diff
@@ -65,17 +65,23 @@ mutable struct LogTargetDensity{C, S, ST <: AbstractTrainingStrategy, I,
 end
 
 """
-cool function to convert parameter's vector to ComponentArray of parameters (for Lux Chain: vector of samples -> Lux ComponentArrays)
+Cool function needed for converting vector of sampled parameters into ComponentVector in case of Lux chain output, derivatives
```
Vaibhavdixit02, line 68:
```suggestion
Function needed for converting vector of sampled parameters into ComponentVector in case of Lux chain output, derivatives
```

src/advancedHMC_MCMC.jl
```diff
@@ -552,6 +558,10 @@ function ahmc_bayesian_pinn_ode(prob::DiffEqBase.ODEProblem, chain;
         end
     end
 
+    println("Current Physics Log-likelihood : ", physloglikelihood(ℓπ, initial_θ))
```
Vaibhavdixit02, line 561: These should be `@info` instead of the `println`s

src/discretize.jl
```diff
@@ -401,7 +401,7 @@ to the PDE.
 For more information, see `discretize` and `PINNRepresentation`.
 """
 function SciMLBase.symbolic_discretize(pde_system::PDESystem,
-                                       discretization::PhysicsInformedNN)
+    discretization::PhysicsInformedNN; bayesian::Bool = false,dataset_given=[nothing])
```
Vaibhavdixit02, line 404: Why `[nothing]` and not `nothing`?

Vaibhavdixit02, line 404: The bayesian flag doesn't feel right, it should be a new discretization type and have a separate dispatch for this function

Vaibhavdixit02:
The comments above are significant but since this PR is already quite big if the tests pass I'd be in favor of merging this and fixing the above comments from a new PR

Vaibhavdixit02:
A general developemnt tip, I would try to avoid such placeholder commits as much as possible, you can just add the filename you changed or something like that so there is at least some information in it because in the future you or someone else might come across this and it can be irritating to look at the changes in files to get an idea if the commit is relevant to you

test/BPINN_PDE_tests.jl
```diff
@@ -0,0 +1,201 @@
...
+# Space and time domains
+domains = [x ∈ Interval(0.0, 1.0)]
+
+# Neural network
+chain = [
+    Lux.Chain(Lux.Dense(1, 10, Lux.tanh), Lux.Dense(10, 10, Lux.tanh),
+        Lux.Dense(10, 1)), Lux.Chain(Lux.Dense(1, 10, Lux.tanh), Lux.Dense(10, 10, Lux.tanh),
+        Lux.Dense(10, 1)), Lux.Chain(Lux.Dense(1, 10, Lux.tanh), Lux.Dense(10, 10, Lux.tanh),
+        Lux.Dense(10, 1)), Lux.Chain(Lux.Dense(1, 4, Lux.tanh), Lux.Dense(4, 1)),
+    Lux.Chain(Lux.Dense(1, 4, Lux.tanh), Lux.Dense(4, 1))]
```
ChrisRackauckas, line 124: That's a pretty deep network for a test. Is that big actually required?

test/BPINN_PDE_tests.jl
```diff
@@ -0,0 +1,201 @@
...
+    Dxu(x) ~ Dx(u(x)) + ep * O1(x),
+    Dxxu(x) ~ Dx(Dxu(x)) + ep * O2(x)]
+
+# Space and time domains
+domains = [x ∈ Interval(0.0, 1.0)]
+
+# Neural network
+chain = [
+    Lux.Chain(Lux.Dense(1, 10, Lux.tanh), Lux.Dense(10, 10, Lux.tanh),
+        Lux.Dense(10, 1)), Lux.Chain(Lux.Dense(1, 10, Lux.tanh), Lux.Dense(10, 10, Lux.tanh),
```
ChrisRackauckas, line 121: 10 -> 1? That doesn't make too much sense

src/discretize.jl
```diff
@@ -586,87 +585,172 @@ function SciMLBase.symbolic_discretize(pde_system::PDESystem,
                                                            pde_loss_functions,
                                                            bc_loss_functions)
 
-    function full_loss_function(θ, p)
+    if bayesian
```
ChrisRackauckas, line 588: Nothing in the function is reused? Make it a dispatch

src/discretize.jl
```diff
@@ -586,87 +585,172 @@ function SciMLBase.symbolic_discretize(pde_system::PDESystem,
...
+            weighted_bc_losses = adaloss.bc_loss_weights .* bc_losses
+
+            sum_weighted_pde_losses = sum(weighted_pde_losses)
+            sum_weighted_bc_losses = sum(weighted_bc_losses)
+            weighted_loss_before_additional = sum_weighted_pde_losses + sum_weighted_bc_losses
+
+            full_weighted_loss = if additional_loss isa Nothing
+                weighted_loss_before_additional
+            else
+                function _additional_loss(phi, θ)
```
ChrisRackauckas, line 698: Is this unchanged?

src/advancedHMC_MCMC.jl
```diff
@@ -598,6 +608,12 @@ function ahmc_bayesian_pinn_ode(prob::DiffEqBase.ODEProblem, chain;
         samples, stats = sample(hamiltonian, Kernel, initial_θ, draw_samples,
             adaptor; progress = progress, verbose = verbose)
 
+        @info("Sampling Complete.")
+        @info("Current Physics Log-likelihood : ", physloglikelihood(ℓπ, samples[end]))
+        @info("Current Prior Log-likelihood : ", priorweights(ℓπ, samples[end]))
+        @info("Current MSE against dataset Log-likelihood : ",
+            L2LossData(ℓπ, samples[end]))
```
ChrisRackauckas, line 615: Is this intended?

src/advancedHMC_MCMC.jl
```diff
@@ -65,17 +65,23 @@ mutable struct LogTargetDensity{C, S, ST <: AbstractTrainingStrategy, I,
...
-function vector_to_parameters(ps_new::AbstractVector, ps::NamedTuple)
-    @assert length(ps_new) == Lux.parameterlength(ps)
-    i = 1
-    function get_ps(x)
-        z = reshape(view(ps_new, i:(i + length(x) - 1)), size(x))
-        i += length(x)
-        return z
+function vector_to_parameters(ps_new::AbstractVector,
+        ps::Union{NamedTuple, ComponentArrays.ComponentVector, AbstractVector})
+    if (ps isa ComponentArrays.ComponentVector) || (ps isa NamedTuple)
```
ChrisRackauckas, line 73: Use dispatch...

src/PDE_BPINN.jl
```diff
@@ -0,0 +1,502 @@
...
+function setparameters(Tar::PDELogTargetDensity, θ)
+    names = Tar.names
+    ps_new = θ[1:(end - Tar.extraparams)]
+    ps = Tar.init_params
+
+    if (ps[names[1]] isa ComponentArrays.ComponentVector)
+        # multioutput case for Lux chains, for each depvar ps would contain Lux ComponentVectors
+        # which we use for mapping current ahmc sampled vector of parameters onto NNs
+        i = 0
+        Luxparams = Vector{ComponentArrays.ComponentVector}()
```
ChrisRackauckas, line 86: This is non-concrete and will slow things down.

src/PDE_BPINN.jl
```diff
@@ -0,0 +1,502 @@
...
+               Tar.allstd) + priorlogpdf(Tar, θ) + L2LossData(Tar, θ)
+    # + L2loss2(Tar, θ)
+end
+
+# function L2loss2(Tar::PDELogTargetDensity, θ)
+#     return Tar.full_loglikelihood(setparameters(Tar, θ),
+#         Tar.allstd)
+# end
+
+function setparameters(Tar::PDELogTargetDensity, θ)
```
ChrisRackauckas, line 77: Doesn't lux have a utility for the transformation of vector into componentvector for this kind of thing?

src/PDE_BPINN.jl
```diff
@@ -0,0 +1,502 @@
...
+    else
+        return logpdf(nnwparams, θ)
+    end
+end
+
+function kernelchoice(Kernel, MCMCkwargs)
+    if Kernel == HMCDA
+        δ, λ = MCMCkwargs[:δ], MCMCkwargs[:λ]
+        Kernel(δ, λ)
+    elseif Kernel == NUTS
```
ChrisRackauckas, line 202: That won't match if it's a type. Needs to use `isa`

src/PDE_BPINN.jl
```diff
@@ -0,0 +1,502 @@
...
+    end
+
+    # NN solutions for loglikelihood which is used for L2lossdata
+    Φ = pinnrep.phi
+
+    # for new L2 loss
+    # discretization.additional_loss = 
+
+    if nchains > Threads.nthreads()
+        throw(error("number of chains is greater than available threads"))
```
ChrisRackauckas, line 359: Why wouldn't that be allowed?

src/PDE_BPINN.jl
```diff
@@ -0,0 +1,502 @@
...
+    elseif nchains < 1
+        throw(error("number of chains must be greater than 1"))
```
ChrisRackauckas, line 361: or equal to

ChrisRackauckas:
Generally looking good, though there's some performance issues to clean up and some issues with style not using dispatch when it would clean things up

src/PDE_BPINN.jl
AstitvaAggarwal: sampling for each chain is done on a single thread, https://turinglang.org/AdvancedHMC.jl/stable/#Parallel-sampling

src/PDE_BPINN.jl
AstitvaAggarwal: i couldnt find it in the docs(atleaast something specific to this case), so i worked it out

src/discretize.jl
AstitvaAggarwal: yes

src/advancedHMC_MCMC.jl
AstitvaAggarwal: yes

src/PDE_BPINN.jl
AstitvaAggarwal: Kernel here is the constructor name, the constructors are exported via AdvancedHMC.jl

ChrisRackauckas: yes but why on the datatype? That's going to ruin inference, and doesn't allow kernel-specific parameters

src/advancedHMC_MCMC.jl
ChrisRackauckas: Seems that should be returned information rather than just printed

src/discretize.jl
ChrisRackauckas: Okay then use dispatch so it's not diff'd

src/PDE_BPINN.jl
ChrisRackauckas: but you could still have main chains than threads. Nothing stops that, it just batches

src/advancedHMC_MCMC.jl
AstitvaAggarwal: i could do that, but the info would be returned only after the sampling. The reason ive chosen to display it is so that at the beginning of the sampling the values of different loglikelihoods are visible and if proportion wrt to each other is off (weak priors will need us to have larger L2 loglikelihoods) we can simply exit sampling and adjust stds,etc.

AstitvaAggarwal: the final values of loglikelihoods shows which objective has been optimized more.

src/PDE_BPINN.jl
ChrisRackauckas: @avik-pal

src/discretize.jl
AstitvaAggarwal: within full_loss_function reweighting and the additional loss is reused, i think i would need to write a similar function again anyways again in dispatch

src/discretize.jl
```diff
@@ -586,8 +585,24 @@ function SciMLBase.symbolic_discretize(pde_system::PDESystem,
                                                            pde_loss_functions,
                                                            bc_loss_functions)
 
-    function full_loss_function(θ, p)
+    if bayesian
+        # required as Physics loss also needed on dataset domain points
+        pde_loss_functions1, bc_loss_functions1 = if !(dataset_given isa Nothing)
+            if !(strategy isa  GridTraining)
```
ChrisRackauckas, line 591: why would only gridtraining be allowed? Wouldn't you just add these points?

src/discretize.jl
ChrisRackauckas: I was thinking that too, but the top level interface has to be different since it's not building an optimization problem. So this is mostly internal, so I think it's okay for simplification.

src/PDE_BPINN.jl
ChrisRackauckas: Let's make sure an issue get's opened about this after merging.

src/discretize.jl
```diff
@@ -401,7 +401,7 @@ to the PDE.
 For more information, see `discretize` and `PINNRepresentation`.
 """
 function SciMLBase.symbolic_discretize(pde_system::PDESystem,
-                                       discretization::PhysicsInformedNN)
+    discretization::PhysicsInformedNN; bayesian::Bool = false,dataset_given=nothing)
```
ChrisRackauckas, line 404: This is missing from the docstring.

ChrisRackauckas: oh actually, this is an interface break.

ChrisRackauckas:
```
function SciMLBase.symbolic_discretize(pde_system::PDESystem,
    discretization::PhysicsInformedNN; bayesian::Bool = false,dataset_given=nothing)
```

that's an interface break, and not one that makes sense. What would `bayesian` or `dataset_given` mean for a finite difference method? Those cannot be added to the high level interface as that's not well thought out in a general sense.

It's probably best to make a new discretization type and then have an `AbstractPINNDiscretizer` that is used in order to make it easy to reuse dispatches. `bayesian` wouldn't be needed then. I'm not entirely sure about `dataset_given` though, that should be known by the PDESystem or discretization?

ChrisRackauckas:
Almost there.

So I think make the discretizer object so the interface is kept, and then open issues for the following:

1. BPINNs on PDEs needs a tutorial and other docs
2. The vector to parameters stuff is a bit slow for type stability issues, but we can merge without handling that since I don't think the functionality should be here anyways, so we should have an issue about reusing some Lux functionality for this.

src/discretize.jl
AstitvaAggarwal: yeah, the points would get included into the training sets, I end up using the GridTraining strategy for loss on those points anyways so was thinking i should allow GridTraining as then throughout domain you have the same strategy applied

src/discretize.jl
```diff
@@ -401,7 +401,7 @@ to the PDE.
 For more information, see `discretize` and `PINNRepresentation`.
 """
 function SciMLBase.symbolic_discretize(pde_system::PDESystem,
-                                       discretization::PhysicsInformedNN)
+    discretization)
```
ChrisRackauckas, line 404: This is type piracy. It needs to specify the top level dispatch.

ChrisRackauckas, line 404:
```suggestion
    discretization::AbstractPINN)
```

ChrisRackauckas (approved):

src/discretize.jl
AstitvaAggarwal: missed out on this, thanks

AstitvaAggarwal:
awesome, thanks @ChrisRackauckas @Vaibhavdixit02 and @xtalax.

src/PDE_BPINN.jl
avik-pal: Typically you store the axes of the ComponentVector via `ax = getaxes(ps_ca)` and then construct the CA via `ComponentArray(theta, ax)`

PR #729, opened 2023-09-03
Docs modifications to use QuadratureTraining instead of GridTraining

sdesai1287:
solved [#634](url)

PR #751, opened 2023-10-06
Investigating doc failure

sathvikbhagavan:
Investigating docs failure of https://github.com/SciML/NeuralPDE.jl/pull/729

### DIFF
--- a/docs/src/tutorials/constraints.md
+++ b/docs/src/tutorials/constraints.md
@@ -35,8 +35,6 @@ Dxx = Differential(x)^2
 _σ = 0.5
 x_0 = -2.2
 x_end = 2.2
-# Discretization
-dx = 0.01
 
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
I already set the optim bound

sathvikbhagavan:
Ok.. what was the reason it was failing in https://buildkite.com/julialang/neuralpde-dot-jl/builds/1875?

ChrisRackauckas:
Optim broke with https://github.com/JuliaNLSolvers/Optim.jl/pull/1034. We're going to need to make an Optim alternative since it has too many problems, and this is accelerating that process.

pkofod:
I can only repeat what I’ve said many times before. These “too many problems” are not consistent with the number of issues opened by sciml developers.

PR #782, opened 2024-01-14
CI trigger

ArnoStrouwen:


ArnoStrouwen:
The test failures on 1.10 are far more severe than on 1.9 @ChrisRackauckas.

ChrisRackauckas:
This is known. Continuing in https://github.com/SciML/NeuralPDE.jl/pull/783

PR #791, opened 2024-01-24
build: try CI with ChainRules@1.58

sathvikbhagavan:
CI on master is broken. I suspect it is because of ChainRules@1.58.2 - https://github.com/JuliaRegistries/General/pull/99357

Testing CI on ChainRules@1.58.1


sathvikbhagavan:
@ChrisRackauckas, the problem was coming from https://github.com/SciML/NeuralPDE.jl/blob/master/src/pinn_types.jl#L520 where it can be `zero(::Vector{Float64})` which was erroring out like in https://github.com/SciML/NeuralPDE.jl/actions/runs/7625540660/job/20770044115#step:6:823.

I just added the rule here which was removed in https://github.com/JuliaDiff/ChainRules.jl/pull/770 which fixes CI, although I am not completely sure if this is the right thing to do.

MasonProtter:
Ugh, this thing is a real whack-a-mole. I'm having trouble understanding though why `zero(::Vector)` should be non-differentiable and return `NoTangent()` instead of returning `ZeroTangent()`, that just seems wrong to me. 

I'm curious @sathvikbhagavan, can you test what happens here if you try to do
```julia
using ChainRulesCore: @scalar_rule
@scalar_rule zero(x::Any) ZeroTangent()
``` 
instead of using `@non_differentiable`?

MasonProtter:
Okay, this is now fixed by https://github.com/JuliaDiff/ChainRules.jl/pull/773

sathvikbhagavan:
Thanks @MasonProtter!

PR #792, opened 2024-01-24
[Do not Merge] Try custom `frule`/`rrule` for `zero`

MasonProtter:
This is a test to see if this is how `zero` should be handled upstream in ChainRules itself.

If this works, I'd like to try putting this version of the rule in ChainRules.jl as an alternative to https://github.com/SciML/NeuralPDE.jl/pull/791 pirating it in.


MasonProtter:
@ChrisRackauckas can you approve the workflows?

sathvikbhagavan:
@MasonProtter this seems to work!

Issue #575, opened 2022-07-15
Integrate NeuralOperator and NeuralPDE


KirillZubov:
Couple of strategy/scenarios
1) train-to-train model  
- train a sub-class of PDEs with NeuralOperator
- transfer learning pre-trained prediction from NeuralOperator problem to PINNs
- PINNs training with  pre-trained model for the instance of the PDE

2) physics informed neural operator 
integrate both into one training scheme 

3) use  each individual prediction, solution, and data together to adaptive update the predict for some the core /kernel/ Green's function of class PDEs(major equations for some problem ) as a pre-trained state.  

https://zongyi-li.github.io/neural-operator/Neural-Operator-CS159-0503-2022.pdf
https://www2.compute.dtu.dk/~apek/SCIML2022/Surrogate_methods.pdf

Physics-Informed Neural Operators
https://arxiv.org/abs/2111.03794
Graph Kernel Network
https://arxiv.org/abs/2003.03485
Generative Adversarial Neural Operators
https://arxiv.org/abs/2205.03017

TODO 
- provide support to the main Neural Operators https://github.com/SciML/NeuralOperators.jl/tree/main/src :
symbolic part, inner types, strategies, training, transfer learning, GPU

- I would like to experiment with this database and do something interesting as demonstration example
https://developer.ibm.com/data/spot-challenge-wildfires/

KirillZubov:
https://github.com/SciML/NeuralOperators.jl/issues/83

KirillZubov:
the first script, training neural operator for a family of Burger equations.
training is  mapping between functional space of some initial conditions  {u(t0 x)} and solution of equation u(t,x)
here use only data 

```julia
using MAT
using Plots
using DataDeps, MAT, MLUtils
using NeuralOperators, Flux
using CUDA, FluxTraining, BSON
import Flux: params
using BSON: @save, @load


#   Burgers' equation dataset from
#         [fourier_neural_operator](https://github.com/zongyi-li/fourier_neural_operator)
#         """,
#         "http://www.med.cgu.edu.tw/NeuralOperators/Burgers_R10.zip",

#### data
burgermat = matopen(joinpath("/home/kirillzubov/data/burgers_pino.mat"))
b = read(burgermat)

input = b["input"]
output =  b["output"]

tspan = b["tspan"]
x = b["x"]
a = b["a"]

size(input)
size(output)

output_ = Array{Float32,4}(undef, 1, 101, 128, 1200)
[...]

PR #806, opened 2024-02-12
[WIP] physics informed neural operator ode 

KirillZubov:
https://github.com/SciML/NeuralPDE.jl/issues/575

Checklist 
- [ ] pino ode 
- [ ] tests
- [ ] doc

### DIFF
--- a/src/NeuralPDE.jl
+++ b/src/NeuralPDE.jl
@@ -45,6 +45,7 @@ include("adaptive_losses.jl")
 include("ode_solve.jl")
 # include("rode_solve.jl")
 include("dae_solve.jl")
+include("pino_ode_solve.jl")
 include("transform_inf_integral.jl")
 include("discretize.jl")
 include("neural_adapter.jl")
@@ -52,7 +53,7 @@ include("advancedHMC_MCMC.jl")
 include("BPINN_ode.jl")
 include("PDE_BPINN.jl")
 
-export NNODE, NNDAE,
+export NNODE, NNDAE, PINOODE
     PhysicsInformedNN, discretize,
     GridTraining, StochasticTraining, QuadratureTraining, QuasiRandomTraining,
     WeightedIntervalTraining,
--- a/src/pino_ode_solve.jl
+++ b/src/pino_ode_solve.jl
@@ -1,0 +1,180 @@
+struct PINOODE{C, O, T, P, B, K} <: DiffEqBase.AbstractODEAlgorithm
+    chain::C
+    opt::O
+    training_mapping::T
+    init_params::P
+    batch::B
+    kwargs::K
+end
+
+function PINOODE(chain,
+        opt,
+        training_mapping,
+        init_params = nothing;
+        batch = nothing,
+        kwargs...)
+    !(chain isa Lux.AbstractExplicitLayer) && (chain = Lux.transform(chain))
+    PINOODE(chain, opt, training_mapping, init_params, batch, kwargs)
+end
+
+"""
+    PINOPhi(chain::Lux.AbstractExplicitLayer, t, u0, st)
+"""
+mutable struct PINOPhi{C, T, U, S}
+    chain::C
+    t0::T
+    u0::U
+    st::S
+    function PINOPhi(chain::Lux.AbstractExplicitLayer, t::Number, u0, st)
+        new{typeof(chain), typeof(t), typeof(u0), typeof(st)}(chain, t, u0, st)
+    end
+end
+
+function generate_pino_phi_θ(chain::Lux.AbstractExplicitLayer, t, u0, init_params)
+    θ, st = Lux.setup(Random.default_rng(), chain)
+    if init_params === nothing
+        init_params = ComponentArrays.ComponentArray(θ)
+    else
+        init_params = ComponentArrays.ComponentArray(init_params)
+    end
+    PINOPhi(chain, t, u0, st), init_params
+end
+
+function (f::PINOPhi{C, T, U})(t::AbstractMatrix,
+        θ) where {C <: Lux.AbstractExplicitLayer, T, U}
+    # Batch via data as row vectors
+    # y, st = f.chain(adapt(parameterless_type(ComponentArrays.getdata(θ.depvar)), t'),
+        # θ.depvar,
+        # f.st)
+    y, st = f.chain(adapt(parameterless_type(ComponentArrays.getdata(θ)), t),
+        θ,
+        f.st)
+    ChainRulesCore.@ignore_derivatives f.st = st
+    # f.u0 .+ (t' .- f.t0) .* y
+    y
+end
+
+function inner_data_loss(phi, ts, θ, a, ground_u, i)
+    u_size = size(ground_u[i])
+    a_arr = fill(a[i], u_size)
+    input_data = reduce(vcat, [ts, a_arr])
+    y = phi(input_data, θ)
+    y - ground_u[i]
+end
+function data_loss(phi,tspan, θ, training_mapping)
+    a, ground_u = training_mapping
+    t0, t_end = tspan
+    size_1 = size(ground_u[1])
+    size_2 = size(ground_u)[1]
+    ts = Float32.(reshape(collect(range(t0, stop = t_end, length = size_1[2])), size_1))
+    loss = [inner_data_loss(phi, ts, θ, a, ground_u,i)
+            for i in 1:size_2]
+    # reduce(vcat, loss)
+    sum(abs2, reduce(vcat, loss)) / (size_1[2] * size_2)
+end
+
+function inner_loss(phi, tspan, θ, training_mapping)
+    loss = data_loss(phi, tspan, θ, training_mapping ) #+ physics_loss()
+    loss
+end
+
+function generate_loss(phi, tspan, training_mapping) #training_mapping::Tuple
+    function loss(θ, _)
+        # inner_loss(phi, f, tspan, θ, p,training_mapping )
+        sum(abs2, inner_loss(phi, tspan, θ, training_mapping))
+    end
+    return loss
+end
+
+function DiffEqBase.__solve(prob::DiffEqBase.AbstractODEProblem,
+        alg::PINOODE,
+        args...;
+        dt = nothing,
+        abstol = 1.0f-6,
+        reltol = 1.0f-3,
+        verbose = false,
+        saveat = nothing,
+        maxiters = nothing)
+    u0 = prob.u0
+    tspan = prob.tspan
+    f = prob.f
+    p = prob.p
+    t0 = tspan[1]
+    # param_estim = alg.param_estim
+
+    #hidden layer
+    chain = alg.chain
+    opt = alg.opt
+
+    #train points generation
+    init_params = alg.init_params
+
+    # mapping between functional space of some vararible 'a' of equation (for example initial
+    # codition {u(t0 x)} or parameter) and solution of equation u(t)
+    training_mapping = alg.training_mapping
+
+    !(chain isa Lux.AbstractExplicitLayer) &&
+        error("Only Lux.AbstractExplicitLayer neural networks are supported")
+
+    phi, init_params = generate_pino_phi_θ(chain, t0, u0, init_params)
+
+    # init_params = ComponentArrays.ComponentArray(init_params)
+    # init_params = ComponentArrays.ComponentArray(;depvar = ComponentArrays.ComponentArray(init_params))
+    @show phi(rand(2,1), init_params)
+    @show phi(rand(2, 10), init_params)
+
+    isinplace(prob) &&
+        throw(error("The PINOODE solver only supports out-of-place ODE definitions, i.e. du=f(u,p,t)."))
+
+    try
+        # TODO
+        phi(rand(2, 1), init_params)
+    catch err
+        if isa(err, DimensionMismatch)
+            throw(DimensionMismatch("Dimensions of the initial u0 and chain should match"))
+        else
+            throw(err)
+        end
+    end
+
+    # batch = if alg.batch === nothing
+    #     true
+    # else
+    #     alg.batch
+    # end
+
+    inner_f = generate_loss(phi, tspan, training_mapping)
+
+    # Creates OptimizationFunction Object from total_loss
+    total_loss(θ, _) = inner_f(θ, phi)
+
+
+    # Optimization Algo for Training Strategies
+    opt_algo = Optimization.AutoZygote()
+
+    # Creates OptimizationFunction Object from total_loss
+    optf = OptimizationFunction(total_loss, opt_algo)
+
+    # iteration = 0
+    # callback = function (p, l)
+    #     iteration += 1
+    #     verbose && println("Current loss is: $l, Iteration: $iteration")
+    #     l < abstol
+    # end
+
+    callback = function (p, l)
+        println("Current loss is: $l")
+        return false
+    end
+
+    # init_params = ComponentArrays.ComponentArray(θ)
+
+    # Zygote.gradient(θ -> total_loss(θ, 1), init_params)
+
+    optprob = OptimizationProblem(optf, init_params)
+    res = solve(optprob, opt; callback = callback, maxiters = maxiters) # alg.kwargs...)
+
+    # PINOsolution(fullsolution)
+    # res
+    (total_loss,optprob,opt,callback, maxiters)
+end
--- a/test/PINO_ode_tests.jl
+++ b/test/PINO_ode_tests.jl
@@ -1,0 +1,40 @@
+using Test
+using OrdinaryDiffEq,OptimizationOptimisers
+using Flux, Lux
+using Statistics, Random
+using NeuralOperators
+# using NeuralPDE
+
+@testset "Example 1" begin
+    linear_analytic = (u0, p, t) -> u0 + sin(p * t) / (p)
+    linear = (u, p, t) -> cos(p * t)
+    tspan = (0.0, 2.0)
+    u0 = 0.0
+    a = Float32.([i  for i in 0.1:0.05:pi])
+
+    u_output_ = []
+    for i in 1:length(a)
+        prob = ODEProblem(ODEFunction(linear, analytic = linear_analytic), u0, tspan, a[i])
+        sol1 = solve(prob, Tsit5(); saveat = 0.02)
+        push!(u_output_, Float32.(sol1.u'))
+    end
+    #TODO u0 -> [u0,a]?
+    prob = ODEProblem(linear, u0, tspan)
+    chain = Lux.Chain(Lux.Dense(2, 5, Lux.σ), Lux.Dense(5, 1))
+    # opt = OptimizationOptimisers.Adam(0.1)
+    opt = OptimizationOptimisers.Adam(0.03)
+    training_mapping = (a, u_output_)
+    alg = NeuralPDE.PINOODE(chain, opt, training_mapping)
+
+    sol = solve(prob,
+        alg, verbose = true,
+        maxiters = 2000, abstol = 1.0f-7)
+
+    total_loss
+    total_loss, optprob, opt, callback, maxiters = sol
+    phi(rand(2, 10), init_params)
+
+    total_loss(init_params, 1)
+
+     using Plots
+end


sathvikbhagavan:
I think we should update NeuralOperator.jl dependency versions rather than changing things here.

KirillZubov:
> I think we should update NeuralOperator.jl dependency versions rather than changing things here.

ok, I will try.

docs/src/tutorials/pino_ode.md
```diff
@@ -0,0 +1,65 @@
...
+
+```@example pino
+prob = ODEProblem(linear, u0, tspan, 0)
+flat_no = FourierNeuralOperator(ch = (2, 16, 16, 16, 16, 16, 32, 1), modes = (16,),
+    σ = gelu)
+opt = OptimizationOptimisers.Adam(0.03)
+alg = PINOODE(flat_no, opt, train_set; is_data_loss = true, is_physics_loss = true)
+pino_solution = solve(prob, alg, verbose = false, maxiters = 1000)
+predict = pino_solution.predict
+ground = u_output_
```
ChrisRackauckas, line 56: Shouldn't it give you a function of the parameters or initial conditions or controls to then re-evaluate?

docs/src/tutorials/pino_ode.md
```diff
@@ -0,0 +1,65 @@
...
+u_output_ = zeros(Float32, 1, instances_size, batch_size)
+prob_set = []
+for (i, a_i) in enumerate(as)
+    prob = ODEProblem(ODEFunction(linear, analytic = linear_analytic), u0, tspan, a_i)
+    sol1 = solve(prob, Tsit5(); saveat = 0.0204)
+    reshape_sol = Float32.(reshape(sol1(range_).u', 1, instances_size, 1))
+    push!(prob_set, prob)
+    u_output_[:, :, i] = reshape_sol
+end
+train_set = TRAINSET(prob_set, u_output_);
```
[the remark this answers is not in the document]
KirillZubov: Yes, it is quite possible to learn nn for a family of parametric equations use only physics loss. In the same way, as in a regular pinns. There is only one more dimension as a parameters.

docs/src/tutorials/pino_ode.md
KirillZubov: I forgot about fine-tuning instances. I'll add the feature.

docs/src/tutorials/pino_ode.md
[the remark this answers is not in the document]
KirillZubov: Although Fourier neural operator it look like it's not really suitable for this role(learn only on physics operator) and It was first of all design for learning from data.

KirillZubov:
@ChrisRackauckas @sathvikbhagavan  Could you please review the PR, I guess it ready for merge.

docs/src/tutorials/pino_ode.md
```diff
@@ -0,0 +1,105 @@
...
+    Lux.Dense(16, 16, Lux.σ),
+    Lux.Dense(16, 16, Lux.σ),
+    Lux.Dense(16, 32, Lux.σ),
+    Lux.Dense(32, 32, Lux.σ),
+    Lux.Dense(32, 1))
+# flat_no = FourierNeuralOperator(ch = (2, 16, 16, 16, 16, 16, 32, 1), modes = (16,),
+#     σ = gelu)
+
+opt = OptimizationOptimisers.Adam(0.01)
+pino_phase = OperatorLearning(train_set, is_data_loss = true, is_physics_loss = true)
```
ChrisRackauckas, line 62: Why is a training dataset required here? That's just the normal neural operator. Can you show the pure PINO?

docs/src/tutorials/pino_ode.md
```diff
@@ -0,0 +1,105 @@
...
+## Solve ODE using learned operator family of parametric ODE for fine tuning.
+```@example pino
+dt = (t_end - t0) / instances_size
+pino_phase = EquationSolving(dt, pino_solution)
+chain = Lux.Chain(Lux.Dense(2, 16, Lux.σ),
+    Lux.Dense(16, 16, Lux.σ),
+    Lux.Dense(16, 32, Lux.σ),
+    Lux.Dense(32, 1))
+alg = PINOODE(chain, opt, pino_phase)
+fine_tune_solution = solve( prob, alg, verbose = false, maxiters = 2000)
```
ChrisRackauckas, line 91: I don't understand why this one is needed. Why not just use the `pino_solution `?

src/pino_ode_solve.jl
```diff
@@ -0,0 +1,366 @@
+"""
+TRAINSET(input_data, output_data; isu0 = false)
+## Positional Arguments
+* input_data: variables set 'a' of equation (for example initial condition {u(t0 x)} or parameter p
+* output_data: set of solutions u(t){a} corresponding parameter 'a'.
+"""
+struct TRAINSET{}
+    input_data::Vector{ODEProblem}
+    output_data::Array
+end
```
ChrisRackauckas, line 10: Too generic of a name to export

ChrisRackauckas:
It doesn't look like this can do the pure PINO with just the physics loss?

Also, it looks like this needs a big rebase.

docs/src/tutorials/pino_ode.md
KirillZubov: usually what I've seen PINO mean train with both and data and physics loss.

docs/src/tutorials/pino_ode.md
KirillZubov: 'pino_solution' is prediction for all family of parametric ODE. 'fine_tune_solution' is prediction for only one instance of ODE from all family. So we do additional training but for predict only one exemplar and increase accuracy using  already trained operator from 'pino_solution'.
'pino_solution' have trained for all but less accurate,  'fine_tune_solution'  for one but with better accurate then 'pino_solution'

KirillZubov:
> It doesn't look like this can do the pure PINO with just the physics loss?
> 
> Also, it looks like this needs a big rebase.

it can with just the physics loss but obvious not so good how it with data. I will add test. 

ok, what do you think it need to rebase?

docs/src/tutorials/pino_ode.md
ChrisRackauckas: But it shouldn't be required.

ChrisRackauckas:
This should have had a bit more of an API discussion before starting. The API is really the key here. I think PINO just falls out of doing that API correctly. So let's dive into https://arxiv.org/abs/2103.10974 . 

The core element of PINO is the way that the network takes functional inputs. While that's the theory, in practice it usually gets simplified to over some vector space of inputs like in https://arxiv.org/abs/2103.10974. So the point of the PINO is that it should learn over basically the space of `u0` and `p`.

Thus there's a few things to disconnect here. You could have a non neural operator also take in u0 and p. But it would treat it slightly differently. But the sample space and the neural network are not necessarily the same thing here.

So that leads us down the path to an API. First of all, the `PINO` should be like `PhysicsInformedNN` except it should make use of information from the bounds metadata https://docs.sciml.ai/ModelingToolkit/stable/basics/Variable_metadata/#Bounds of the parameters. For anything with bounds, it should seek to train a neural network that satisfies all parameter values within the bounds. To keep things simple, it should have a keyword argument `bounds` which takes an array of tuples for the bounds of the parameters, and pre-populate it to match the bounds from the metadata. Anything without bounds would be treated as a constant. 

Initial conditions can simply be set to be functions of parameters, so only parameters need to be supported and a tutorial can handle the rest.

So `PINO` would take the same information as `PhysicsInformedNN` except that it would also require parameter bounds. Then its physical loss would sample over the parameter space as well. That would need possibly its own strategy, but random and quasi-random would do for starters.

For solution data, `PhysicsInformedNN` and `PINO` should just have a nice way of supporting that, that's just a completely separate feature.

What would be required though is a slightly different implementation of the NN. We should then require `NN(indepvars,p)` where the first part is the independent variables `[t,x,y,...]`. Thus this is a bit of a difference from the NN form from before. But this is required for example for things like the DeepONets which treat the two in separate neural networks and then merge. The output is a solution for those parameters.

So given this is the PINO... I don't understand how this PR implements an API like this at all. 

```julia
    pino_phase = EquationSolving(dt, pino_solution)
    chain = Lux.Chain(Lux.Dense(2, 16, Lux.σ),
        Lux.Dense(16, 16, Lux.σ),
        Lux.Dense(16, 32, Lux.σ),
        Lux.Dense(32, 1))
    alg = PINOODE(chain, opt, pino_phase)
    fine_tune_solution = solve(
        prob, alg, verbose = false, maxiters = 2000)
```

this doesn't have the right arguments, the space over which things are trained, a neural operator compliant NN? I don't understand how any of this is the PINO.

So let's take a step back. Before doing it for the PDEs, let's get the ODE form. `PINOODE` needs a chain of the form `NN([t],p)` and, because ODEProblems don't have it, some representation of the bounds over which to sample `p`. It needs to then learn by sampling both `t` and`p`. Now the weird thing is representing the solution here, since it's not quite an ODESolution. What I would recommend is giving it the ODE solution at the `p` specified by the `prob`, but then document that `sol.original` gives the neural network weights for the `NN(t,p)` object, and show how it can be used to sample at new points. 

@KirillZubov @sathvikbhagavan are we in agreement on the API here?

sathvikbhagavan:
>are we in agreement on the API here?

yes, makes sense.

KirillZubov:
> are we in agreement on the API here?

@ChrisRackauckas thank for you comment.
I oriented on this article https://arxiv.org/abs/2111.03794 while implementing 'pinoode'. Here it is from appeared 'fine_tune_solution' and other features which remained unclear to you.

I agree with your comment. The mapping between the functional space by parameters and the solution is not explicitly shown in the interface as arguments. Something like 'mapping = [u0, u(t)]'. Instead of this,  space of parameters are implicitly generated as datasets and put in 'TRAINSET', which is not a better API solution - agree.

Also, in a more general form, it can be not limited only to 'u0' and 'p' giving access to parameterize any argument or even a function as part of an equation.  

I’ll think about it and provide in this PR what an API for PINOODE might look like according to your comments.

After agreeing with you about API, I will begin to upgrade the code for PINOODE following the new requirements.

To discuss API for PINO PDE, I'll create an added issue and provide my version of the prototype for API there too but later.

KirillZubov:
@ChrisRackauckas  Considering your comments above, I tried to make a new API for PINOODE. Could you please check, that is what you described and waiting for?

```julia
#API
#only physics
equation = (u, p, t) -> cos(p * t)
tspan = (0.0f0, 2.0f0)
u0 = 0.0f0

prob = ODEProblem(equation, u0, tspan)
# init neural operator, NN([t],p)
neural_operator = Lux.Chain(Lux.Dense(2, 32, Lux.σ), Lux.Dense(32, 1))
opt = Adam(0.01)
#functions of parameters and bound
bounds = (p = [0, pi / 2])
alg = PINOODE(neural_operator, bounds, opt)
pino_solution = solve(prob, alg)


# with data
equation = (u, p, t) -> cos(p * t)
tspan = (0.0f0, 2.0f0)
u0 = 0.0f0
prob = ODEProblem(linear, u0, tspan)
# init neural operator, NN([t],p)
neural_operator = Lux.Chain(Lux.Dense(2, 32, Lux.σ), Lux.Dense(32, 1))
opt = Adam(0.01)
#functions of parameters and bound
bounds = (p = [0, pi / 2])

train_set = ... #vector of ground solution
alg = PINOODE(neural_operator, bounds, opt; train_set = train_set)
#or
ground_solution = (u0, p, t) -> u0 + sin(p * t) / (p)
alg = PINOODE(neural_operator, bounds, opt; ground_solution = ground_solution)
pino_solution = solve(prob, alg)
```

KirillZubov:
Implementation Physics-informed neural operator method for solve parametric Ordinary Differential Equations (ODE) use DeepOnet.

finmod:
@KirillZubov Can I suggest a PINO-based spatiotemporal MFG example and test  https://www.mdpi.com/2227-7390/12/6/803

docs/src/tutorials/pino_ode.md
```diff
@@ -0,0 +1,66 @@
...
+    Lux.Dense(1, 10, Lux.tanh_fast),
+    Lux.Dense(10, 10, Lux.tanh_fast),
+    Lux.Dense(10, 10))
+trunk = Lux.Chain(
+    Lux.Dense(1, 10, Lux.tanh_fast),
+    Lux.Dense(10, 10, Lux.tanh_fast),
+    Lux.Dense(10, 10, Lux.tanh_fast))
+deeponet = NeuralPDE.DeepONet(branch, trunk; linear = nothing)
+
+bounds = (p = [0.1f0, pi],)
```
ChrisRackauckas, line 32: Having more than one parameters would be more illustrative of it's use case. Right now branch and trunk both have size one input and that makes it potentially confusing to the user for how to modify this demo towards a case with more parameters

docs/src/tutorials/pino_ode.md
```diff
@@ -0,0 +1,66 @@
...
+db = (bounds.p[2] - bounds.p[1]) / 50
+dt = (tspan[2] - tspan[1]) / 40
+strategy = NeuralPDE.GridTraining([db, dt])
```
ChrisRackauckas, line 35: Don't use grid training

docs/src/tutorials/pino_ode.md
```diff
@@ -0,0 +1,66 @@
...
+```
+
+Now let's compare the prediction from the learned operator with the ground truth solution which is obtained by analytic solution the parametric ODE. Where 
+Compare prediction with ground truth.
+
+```@example pino
+using Plots
+# Compute the ground truth solution for each parameter
+ground_analytic = (u0, p, t) -> u0 + sin(p * t) / (p)
+p_ = bounds.p[1]:strategy.dx[1]:bounds.p[2]
```
ChrisRackauckas, line 49: Not using grid training will make this more compelling since it should predict at new parameters, not ones trained on

docs/src/tutorials/pino_ode.md
```diff
@@ -0,0 +1,66 @@
...
+p = reshape(p_, 1, size(p_)[1], 1)
+ground_solution = ground_analytic.(u0, p, sol.t.trunk)
+
+# Plot the predicted solution and the ground truth solution as a filled contour plot
+# sol.u[1, :, :], represents the predicted solution for each parameter value and time
+plot(predict[1, :, :], linetype = :contourf)
+plot!(ground_solution[1, :, :], linetype = :contourf)
```
ChrisRackauckas, line 56: You don't show how to generate the solution at new parameters which is the key to the pini interface

docs/src/tutorials/pino_ode.md
```diff
@@ -0,0 +1,66 @@
...
+deeponet = NeuralPDE.DeepONet(branch, trunk; linear = nothing)
+
+bounds = (p = [0.1f0, pi],)
+db = (bounds.p[2] - bounds.p[1]) / 50
+dt = (tspan[2] - tspan[1]) / 40
+strategy = NeuralPDE.GridTraining([db, dt])
+opt = OptimizationOptimisers.Adam(0.03)
+alg = NeuralPDE.PINOODE(deeponet, opt, bounds; strategy = strategy)
+sol = solve(prob, alg, verbose = false, maxiters = 2000)
+predict = sol.u
```
ChrisRackauckas, line 39: sol.original is needed to explain

src/NeuralPDE.jl
```diff
@@ -55,7 +58,8 @@ include("BPINN_ode.jl")
 include("PDE_BPINN.jl")
 include("dgm.jl")
 
-export NNODE, NNDAE,
+
+export NNODE, NNDAE, PINOODE, DeepONet, SomeStrategy #TODO remove SomeStrategy
```
ChrisRackauckas, line 62: Some strategy?

src/neural_operators.jl
```diff
@@ -0,0 +1,101 @@
...
+x = (branch = a, trunk = b)
+θ, st = Lux.setup(Random.default_rng(), deeponet)
+y, st = deeponet(x, θ, st)
+```
+
+## References
+* Lu Lu, Pengzhan Jin, George Em Karniadakis "DeepONet: Learning nonlinear operators for identifying differential equations based on the universal approximation theorem of operators"
+* Sifan Wang "Learning the solution operator of parametric partial differential equations with physics-informed DeepOnets"
+"""
+struct DeepONet{L <: Union{Nothing, Lux.AbstractExplicitLayer }} <: NeuralOperator
```
ChrisRackauckas, line 56: This should really be living in NeuralOperators.jl. cc @avik-pal

src/pino_ode_solve.jl
```diff
@@ -0,0 +1,217 @@
...
+    x_left = (branch = branch_left, trunk = trunk_left)
+    x_right = (branch = branch_right, trunk = trunk_right)
+    (phi(x_left, θ) .- phi(x_right, θ)) / sqrt(eps(eltype(t)))
+end
+
+function physics_loss(
+        phi::PINOPhi{C, T}, prob::ODEProblem, x, θ) where {C <: DeepONet, T}
+    p, t = x
+    f = prob.f
+    #TODO If du = f(u,p,t), where f = g(p,t)*u so it will wrong, f(0, p, t) = g(p,t)*0 = 0
```
ChrisRackauckas, line 92: That's a limit on the architecture?

src/pino_ode_solve.jl
```diff
@@ -0,0 +1,217 @@
...
+    tuple = (branch = f_0, trunk = t0)
+    out = phi(tuple, θ)
+    u = vec(out)
+    u0_ = fill(prob.u0, size(out))
+    u0 = vec(u0_)
+    norm = prod(size(u0_))
+    sum(abs2, u .- u0) / norm
+end
+
+function get_trainset(strategy::GridTraining, bounds, tspan)
```
ChrisRackauckas, line 118: Just reuse the non pino?

ChrisRackauckas:
I don't think it's quite there yet. Some comments

test/PINO_ode_tests.jl
```diff
@@ -0,0 +1,142 @@
...
+    alg = NeuralPDE.PINOODE(
+        deeponet, opt, bounds; strategy = strategy, additional_loss = additional_loss_)
+    sol = solve(prob, alg, verbose = false, maxiters = 2000)
+
+    p_ = bounds.p[1]:strategy.dx[1]:bounds.p[2]
+    p = reshape(p_, 1, size(p_)[1], 1)
+    ground_solution = ground_analytic.(u0, p, sol.t.trunk)
+
+    @test ground_solution≈sol.u rtol=0.005
+end
```
sathvikbhagavan, line 142: Can you also tests with vector outputs and multiple parameters?

src/neural_operators.jl
KirillZubov: I just implemented it for tests because NeuralOperators.jl was outdate and couldn't used it. Yes agree, it should relocate to NeuralOperators.jl

src/pino_ode_solve.jl
```diff
@@ -0,0 +1,217 @@
...
+        opt,
+        bounds;
+        init_params = nothing,
+        strategy = nothing,
+        additional_loss = nothing,
+        kwargs...)
+    !(chain isa Lux.AbstractExplicitLayer) && (chain = Lux.transform(chain))
+    PINOODE(chain, opt, bounds, init_params, strategy, additional_loss, kwargs)
+end
+mutable struct PINOPhi{C, S}
```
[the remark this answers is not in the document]
ChrisRackauckas: No, any Lux layer should be fine as immutable.

src/neural_operators.jl
sathvikbhagavan: Should it be NeuralOperators.jl or LuxNeuralOperators.jl? I see https://github.com/LuxDL/LuxNeuralOperators.jl/pull/5 implementing Deeponet

src/NeuralPDE.jl
KirillZubov: fix it

src/neural_operators.jl
KirillZubov: yes, Lux. I didn't know there was already one like this

KirillZubov: I can do this task: move DeepOnet from here to LuxNeuralOperators

docs/src/tutorials/pino_ode.md
KirillZubov: Ok, I will support also QuasiRandomTraining for PINO ODE and use it in Doc  example

docs/src/tutorials/pino_ode.md
KirillZubov: Ok, I will do. It will take some time to implement this feature for many parameters.

docs/src/tutorials/pino_ode.md
KirillZubov: yes,I agree

docs/src/tutorials/pino_ode.md
KirillZubov: as new parameters you means some another mesh that don't use for training but in the same boundary of parameters?

src/pino_ode_solve.jl
KirillZubov: I'm not sure, it takes time to figure it out, I'll come back to this later.

src/pino_ode_solve.jl
KirillZubov: here, function that is create a set of training data is for both the physics loss function and the loss function for the initial condition ||u(0)-u0|| (it was named operator_loss I renamed it to inital_condition_loss)

test/PINO_ode_tests.jl
KirillZubov: ok, I can

docs/src/tutorials/pino_ode.md
ChrisRackauckas: yes

src/pino_ode_solve.jl
```diff
@@ -0,0 +1,219 @@
...
+        bounds;
+        init_params = nothing,
+        strategy = nothing,
+        additional_loss = nothing,
+        kwargs...)
+    !(chain isa Lux.AbstractExplicitLayer) && (chain = Lux.transform(chain))
+    PINOODE(chain, opt, bounds, init_params, strategy, additional_loss, kwargs)
+end
+
+struct PINOPhi{C, S}
```
KirillZubov: @ChrisRackauckas oh actually it should be mutable, how it is done for `ODEPhi`  

https://github.com/SciML/NeuralPDE.jl/blob/ecc6a408283c069d129b797e72a27b2817be4748/src/ode_solve.jl#L103C1-L111C4

ChrisRackauckas: that should probably be made immutable too 😅

sathvikbhagavan: Ah, I can look into that

src/neural_operators.jl
ayushinav: Basic NO and DeepONet are now in LuxNeuralOperators.jl
@KirillZubov

src/pino_ode_solve.jl
KirillZubov: @ChrisRackauckas yes it was limit on the architecture. if we have equation du =f(u,p,t), which  ’du = cos(p * t) + u’ and operator for PINO is  ‘G(s(p,t), t)’, where ’s(p,t) =  cos(p * t)’.

it turns out that we  do not have the opportunity to turn only to cos(p * t) so I did it with f(0,p,t) =  cos(p * t) +0.  and it is the reason for this #TODO
 
I see two ways how to solve it problem. First, it is add an element to the interface where we explicitly denote a parametric function like ‘parametric_function = (function =cos(p * t), bound= (a,b))’. 
 
or we can change the form of the operator to ‘G(p, t)’,  ’ use ‘p’ inspite of ’s(p,t) as training data but it lead to  lower guaranteed convergence and a lower learning rate.

KirillZubov: After some reflection about it and couple of tests, I lead it will better use last case. To use just p and operator  G(p, t).

KirillZubov:
https://github.com/LuxDL/LuxNeuralOperators.jl/issues/7
https://github.com/LuxDL/LuxNeuralOperators.jl/issues/9

docs/src/tutorials/pino_ode.md
KirillZubov: StochasticTraining*

KirillZubov:
hey @sathvikbhagavan  @ChrisRackauckas  I fixed all that was noted from the last review, except for the vector output, this is a issue that is currently in progress https://github.com/LuxDL/LuxNeuralOperators.jl/issues/9. I will do it later in a separate PR.

src/pino_ode_solve.jl
sathvikbhagavan: https://github.com/SciML/NeuralPDE.jl/pull/864

docs/src/tutorials/pino_ode.md
```diff
@@ -0,0 +1,95 @@
...
+    Chain(Dense(1 => 10, Lux.tanh_fast), Dense(10 => 10, Lux.tanh_fast),
+        Dense(10 => 10, Lux.tanh_fast)))
+
+u = rand(3, 50)
+v = rand(1, 40, 1)
+θ, st = Lux.setup(Random.default_rng(), deeponet)
+c = deeponet((u, v), θ, st)[1]
+
+bounds = [(1.0f0, pi), (1.0f0, 2.0f0), (2.0f0, 3.0f0)]
+number_of_parameters = 50
```
sathvikbhagavan, line 35:
```suggestion
number_of_parameter_samples = 50
```

to avoid confusion from `number_of_parameter`

docs/src/tutorials/pino_ode.md
```diff
@@ -0,0 +1,95 @@
...
+strategy = StochasticTraining(40)
+opt = OptimizationOptimisers.Adam(0.03)
+alg = PINOODE(deeponet, opt, bounds, number_of_parameters; strategy = strategy)
+sol = solve(prob, alg, verbose = true, maxiters = 3000)
```
sathvikbhagavan, line 39:
```suggestion
sol = solve(prob, alg, verbose = false, maxiters = 3000)
```

better for users and CI

docs/src/tutorials/pino_ode.md
```diff
@@ -0,0 +1,95 @@
...
+ground_solution = (u0, p, t) -> u0 + p[1] / p[2] * sin(p[2] * t) + p[3]*t
+function ground_solution_f(p,t)
+    reduce(hcat,[[ground_solution(u0, p[:, i], t[j]) for j in axes(t, 2)] for i in axes(p, 2)])
+end
+
+(p,t) = get_trainset(bounds, tspan, 50, 0.025f0)
+ground_solution_ = ground_solution_f(p,t)
+predict = sol.interp((p,t))
+
+# Calculate the mean error and  the standard deviation of the errors
```
sathvikbhagavan, line 66:
```suggestion
# Calculate the mean error and the standard deviation of the errors
```

docs/src/tutorials/pino_ode.md
```diff
@@ -0,0 +1,95 @@
...
+```@example pino
+using Test
+using OptimizationOptimisers
+using Lux
+using Statistics, Random
+using LuxNeuralOperators
+using NeuralPDE
+
+equation = (u, p, t) -> p[1] * cos(p[2] * t) + p[3]
+tspan = (0.0f0, 1.0f0)
```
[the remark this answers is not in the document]
KirillZubov: why?

KirillZubov:
> A few comments mostly cleanups/clarifications. Can you make CI happy as well?

It is because I began implement vector output and now PR in WIP again. Yep, sure, I will make CI to not fail condition at the end.

test/PINO_ode_tests.jl
```diff
@@ -0,0 +1,265 @@
...
+    predict = sol.interp((p, t))
+    @test ground_solution_≈predict rtol=0.01
+end
+
+#vector output
+
+linear = (u, p, t) -> [cos(2pi * t), sin(2pi * t)]
+tspan = (0.0f0, 1.0f0)
+u0 = [0.0f0, -1.0f0 / 2pi]
+linear_analytic = (u0, p, t) -> [sin(2pi * t) / 2pi, -cos(2pi * t) / 2pi]
```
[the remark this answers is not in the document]
KirillZubov: now it is in work in progress for vector output again but probably I will `undo` it and froze this task for a while and will finish it in separate the PR.  I will come back to it later and for now jump to my main PINO PDE task.

KirillZubov:
> A few comments mostly cleanups/clarifications. Can you make CI happy as well?
@sathvikbhagavan Can we add not registered package on CI? 
`ERROR: LoadError: expected package `LuxNeuralOperators [c0ba2cc5]` to be registered`

sathvikbhagavan:
> ERROR: LoadError: expected package `LuxNeuralOperators [c0ba2cc5]` to be registered

@avik-pal, can you register LuxNeuralOperators? cc: @ChrisRackauckas

docs/src/tutorials/pino_ode.md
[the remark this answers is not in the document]
sathvikbhagavan: All PINNs/DeepONet etc based are best trained with high precision. IMO, our defaults should be Float64.

KirillZubov:
I did all from my side for now with this PR. `LuxNeuralOperators` is unregister yet. Multi output is work in progress  https://github.com/LuxDL/LuxNeuralOperators.jl/pull/11#issuecomment-2210419738 . So now I move to PDE PINO issue https://github.com/SciML/NeuralPDE.jl/pull/862 and come back to PINO ODE later.

sathvikbhagavan:
I will review this today.

.github/workflows/CI.yml
```diff
@@ -58,3 +59,5 @@ jobs:
           files: lcov.info
           token: ${{ secrets.CODECOV_TOKEN }}
           fail_ci_if_error: true
+    
+
```
sathvikbhagavan, line 63:
```suggestion
```

Project.toml
```diff
@@ -97,4 +98,4 @@ SafeTestsets = "1bc83da4-3b8d-516f-aca4-4fe02f6d838f"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [targets]
-test = ["Aqua", "Test", "CUDA", "SafeTestsets", "OptimizationOptimJL", "Pkg", "OrdinaryDiffEq", "LineSearches", "LuxCUDA", "Flux", "MethodOfLines"]
+test = ["Aqua", "Test", "CUDA", "SafeTestsets", "OptimizationOptimJL", "Pkg", "OrdinaryDiffEq", "LineSearches", "LuxCUDA", "Flux", "MethodOfLines"]
```
sathvikbhagavan, line 101:
```suggestion
test = ["Aqua", "Test", "CUDA", "SafeTestsets", "OptimizationOptimJL", "Pkg", "OrdinaryDiffEq", "LineSearches", "LuxCUDA", "Flux", "MethodOfLines"]

```

src/NeuralPDE.jl
```diff
@@ -55,7 +57,8 @@ include("BPINN_ode.jl")
 include("PDE_BPINN.jl")
 include("dgm.jl")
 
-export NNODE, NNDAE,
+
+export NNODE, NNDAE, PINOODE
```
sathvikbhagavan, line 61:
```suggestion
export NNODE, NNDAE, PINOODE
```

docs/src/tutorials/pino_ode.md
```diff
@@ -0,0 +1,95 @@
+# Physics Informed Neural Operator for ODEs
+
+This tutorial provides an example of how to use the Physics Informed Neural Operator (PINO) for solving a family of parametric ordinary differential equations (ODEs).
+
+## Operator Learning for a family of parametric ODEs
+
+In this section, we will define a parametric ODE and solve it using a PINO. The PINO will be trained to learn the mapping from the parameters of the ODE to its solution.
```
sathvikbhagavan, line 7:
```suggestion
In this section, we will define a parametric ODE and then learn it with a PINO using [`PINOODE`](@ref). The PINO will be trained to learn the mapping from the parameters of the ODE to its solution.
```

docs/src/tutorials/pino_ode.md
```diff
@@ -0,0 +1,95 @@
...
+bounds = [(1.0f0, pi), (1.0f0, 2.0f0), (2.0f0, 3.0f0)]
+number_of_parameter_samples = 50
+strategy = StochasticTraining(40)
+opt = OptimizationOptimisers.Adam(0.03)
+alg = PINOODE(deeponet, opt, bounds, number_of_parameters; strategy = strategy)
+sol = solve(prob, alg, verbose = false, maxiters = 3000)
+```
+
+Now let's compare the prediction from the learned operator with the ground truth solution which is obtained by analytic solution the parametric ODE. Where 
+Compare prediction with ground truth.
```
sathvikbhagavan, line 43:
```suggestion
Now let's compare the prediction from the learned operator with the ground truth solution which is obtained by analytic solution of the parametric ODE.
```

docs/src/tutorials/pino_ode.md
```diff
@@ -0,0 +1,95 @@
...
+u = rand(3, 50)
+v = rand(1, 40, 1)
+θ, st = Lux.setup(Random.default_rng(), deeponet)
+c = deeponet((u, v), θ, st)[1]
+
+bounds = [(1.0f0, pi), (1.0f0, 2.0f0), (2.0f0, 3.0f0)]
+number_of_parameter_samples = 50
+strategy = StochasticTraining(40)
+opt = OptimizationOptimisers.Adam(0.03)
+alg = PINOODE(deeponet, opt, bounds, number_of_parameters; strategy = strategy)
```
sathvikbhagavan, line 38: I think this needs to explained step by step as its not intuitive to a user. So breakdown step by step with some small explanations.

docs/src/tutorials/pino_ode.md
```diff
@@ -0,0 +1,95 @@
...
+sol = solve(prob, alg, verbose = false, maxiters = 3000)
+```
+
+Now let's compare the prediction from the learned operator with the ground truth solution which is obtained by analytic solution the parametric ODE. Where 
+Compare prediction with ground truth.
+
+```@example pino
+using Plots
```
sathvikbhagavan, line 46: Same here, add couple of lines about what is being visualized.

docs/src/tutorials/pino_ode.md
```diff
@@ -0,0 +1,95 @@
...
+```@example pino
+# 'i' is the index of the parameter 'p' in the dataset 
+i = 5
+# 'predict' is the predicted solution from the PINO model
+plot(predict[:, i], label = "Predicted")
+# 'ground' is the ground truth solution
+plot!(ground_solution_[:, i], label = "Ground truth")
+```
+
+
```
sathvikbhagavan, line 95:
```suggestion
```

src/pino_ode_solve.jl
```diff
@@ -0,0 +1,287 @@
...
+    dt = strategy.dx
+    if size(bounds,1) == 1
+        bound = bounds[1]
+        p_ = range(start = bound[1], length = number_of_parameters, stop = bound[2])
+        p = collect(reshape(p_, 1, size(p_,1)))
+    else
+        p_ = [range(start = b[1], length = number_of_parameters, stop = b[2])
+              for b in bounds]
+        p = vcat([collect(reshape(p_i, 1, size(p_i,1))) for p_i in p_]...)
+    end
```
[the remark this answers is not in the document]
sathvikbhagavan: Can you address this?

src/pino_ode_solve.jl
```diff
@@ -0,0 +1,275 @@
...
+    branch_left, branch_right = p, p
+    trunk_left, trunk_right = t .+ sqrt(eps(eltype(t))), t
+    x_left = (branch_left, trunk_left)
+    x_right = (branch_right, trunk_right)
+    (phi(x_left, θ) .- phi(x_right, θ)) ./ sqrt(eps(eltype(t)))
+end
+
+function dfdx(phi::PINOPhi{C, T}, x::Tuple,
+        θ) where {C <: CompactLuxLayer{:FourierNeuralOperator,}, T}
+    #TODO
```
sathvikbhagavan, line 96: Why is this TODO?

src/pino_ode_solve.jl
```diff
@@ -0,0 +1,275 @@
...
+        f_vec = vec(fs)
+    else
+        f_vec = reduce(
+            vcat, [[f(out[i], p[:, i], t[j]) for j in axes(t, 2)] for i in axes(p, 2)])
+    end
+    du = vec(dfdx(phi, x, θ))
+    norm = prod(size(du))
+    sum(abs2, du .- f_vec) / norm
+end
+#TODO Lux.AbstractExplicitLayer
```
sathvikbhagavan, line 119: What is the TODO here?

src/pino_ode_solve.jl
```diff
@@ -0,0 +1,275 @@
...
+    t0 = reshape([prob.tspan[1]], (1, 1, 1)) # t[:, [1], :]
+    x0 = (p, t0)
+    out = phi(x0, θ)
+    u = vec(out)
+    u0 = vec(fill(prob.u0, size(out)))
+    norm = prod(size(u0))
+    sum(abs2, u .- u0) / norm
+end
+
+#TODO for input FourierNeuralOperator
```
sathvikbhagavan, line 133: What is the TODO here?

sathvikbhagavan (commented):
I have suggested some cleanups/changes. There are some comments from last review not addressed. Can you address them? Can you also run through formatter as well? I think a couple more iterations, it should be close to merging.

KirillZubov:
> I have suggested some cleanups/changes. There are some comments from last review not addressed. Can you address them? Can you also run through formatter as well? I think a couple more iterations, it should be close to merging.

Which  comments from last review not addressed? LuxNeuralOperator still not registered and` DeepONet multi output` task is in work in progress so I can't address to them before they will have been not done. So I work on PDE PINO and support FourierNeuralOperator now.

src/pino_ode_solve.jl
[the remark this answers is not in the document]
KirillZubov: yes, it can be combined in one. I will do it.

test/PINO_ode_tests.jl
```diff
@@ -0,0 +1,265 @@
...
+using OptimizationOptimisers
+using Lux
+using Statistics, Random
+using LuxNeuralOperators
+using NeuralPDE
+
+# dG(u(t, p), t) = f(G,u(t, p))
+@testset "Example du = cos(p * t)" begin
+    equation = (u, p, t) -> cos(p * t)
+    tspan = (0.0f0, 1.0f0)
```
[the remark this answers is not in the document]
KirillZubov: Why? In others tests PINNS ODE and many others we use Float32. And more over we need initialize neural network with weights Float64 then it will make sense. Otherwise, we multiply float64 by float32 in the neural network and do not get increase accuracy, only an unnecessary conversion. This is not critical for one-dimensional ODE problems, but for multi-dimensional ones with PDE, training in float64 will take much longer and it can be too long for test time. So I don’t really understand why use Float64 everywhere, if only one test and then it’s not necessary

src/pino_ode_solve.jl
```diff
@@ -0,0 +1,287 @@
...
+    out = phi(x, θ)
+    if size(p,1) == 1
+        fs = f.(out, p, vec(t))
+        f_vec = vec(fs)
+    else
+        f_vec = reduce(
+            vcat, [[f(out[i], p[:, i], t[j]) for j in axes(t, 2)] for i in axes(p, 2)])
+    end
+    du = vec(dfdx(phi, x, θ))
+    norm = prod(size(du))
```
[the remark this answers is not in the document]
KirillZubov: normalisation for loss function

src/pino_ode_solve.jl
```diff
@@ -0,0 +1,287 @@
...
+    (phi(x_left, θ) .- phi(x_right, θ)) ./ sqrt(eps(eltype(t)))
+end
+
+function physics_loss(
+        phi::PINOPhi{C, T}, prob::ODEProblem, x::Tuple, θ) where {
+        C <: CompactLuxLayer{:DeepONet,}, T}
+    p, t = x
+    f = prob.f
+    out = phi(x, θ)
+    if size(p,1) == 1
```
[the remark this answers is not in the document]
KirillZubov: yes, it can be also combine in one but with loss runtime for first but it is ok here.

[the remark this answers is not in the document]
sathvikbhagavan: If it can be combined, it would be better as its not a special case 😅

test/PINO_ode_tests.jl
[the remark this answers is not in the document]
sathvikbhagavan: Alright, I would be fine with tests but not with tutorials as people usually copy paste.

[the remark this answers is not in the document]
KirillZubov: Oh, I didn't see some comments from last preview. I will fix it in soon.

KirillZubov:
@sathvikbhagavan Do you know how we can add  [NeuralOperators](https://github.com/LuxDL/NeuralOperators.jl) to CI while it is not registered?

KirillZubov:
I would like to add multi output when [DeepOnet with additional layer](https://github.com/LuxDL/NeuralOperators.jl/issues/25) will be fix and add support just with Chain because it can be usefulness for tests but in next PR. 
This PR, I guess, is ready for review and then merge if it all fine.

sathvikbhagavan:
Thanks! I will review this in a couple of days.

src/pino_ode_solve.jl
```diff
@@ -0,0 +1,250 @@
...
+* `chain`: A neural network architecture, defined as a `Lux.AbstractExplicitLayer` or `Flux.Chain`.
+                 `Flux.Chain` will be converted to `Lux` using `adapt(FromFluxAdaptor(false, false), chain)`
+* `opt`: The optimizer to train the neural network.
+* `bounds`: A dictionary containing the bounds for the parameters of the neural network
+in which will be train the prediction of parametric ODE.
+* `number_of_parameters`: The number of points of train set in parameters boundaries.
+
+## Keyword Arguments
+
+* `init_params`: The initial parameter of the neural network. By default, this is `nothing`,
```
sathvikbhagavan, line 23:
```suggestion
* `init_params`: The initial parameters of the neural network. By default, this is `nothing`,
```

src/pino_ode_solve.jl
```diff
@@ -0,0 +1,250 @@
...
+Algorithm for solving paramentric ordinary differential equations using a physics-informed
+neural operator, which is used as a solver for a parametrized `ODEProblem`.
+
+## Positional Arguments
+
+* `chain`: A neural network architecture, defined as a `Lux.AbstractExplicitLayer` or `Flux.Chain`.
+                 `Flux.Chain` will be converted to `Lux` using `adapt(FromFluxAdaptor(false, false), chain)`
+* `opt`: The optimizer to train the neural network.
+* `bounds`: A dictionary containing the bounds for the parameters of the neural network
+in which will be train the prediction of parametric ODE.
```
sathvikbhagavan, line 18:
```suggestion
* `bounds`: A dictionary containing the bounds for the parameters of the parametric ODE.
```

src/pino_ode_solve.jl
```diff
@@ -0,0 +1,250 @@
...
+* `number_of_parameters`: The number of points of train set in parameters boundaries.
+
+## Keyword Arguments
+
+* `init_params`: The initial parameter of the neural network. By default, this is `nothing`,
+                             which thus uses the random initialization provided by the neural network library.
+* `strategy`: The strategy for training the neural network.
+* `additional_loss`: additional function to the main one. For example, add training on data.
```
sathvikbhagavan, line 26:
```suggestion
* `additional_loss`: additional loss function added to the default one. For example, add training on data.
```

docs/src/tutorials/pino_ode.md
```diff
@@ -0,0 +1,95 @@
+# Physics Informed Neural Operator for ODEs Solvers
+
+This tutorial provides an example of how to use the Physics Informed Neural Operator (PINO) for solving a family of parametric ordinary differential equations (ODEs).
+
+## Operator Learning  for a family of parametric ODE.
+
+In this section, we will define a parametric ODE and solve it using a PINO. The PINO will be trained to learn the mapping from the parameters of the ODE to its solution.
+
+```@example pino
+using Test
```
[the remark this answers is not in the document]
sathvikbhagavan: I don't think it is resolved.

docs/src/tutorials/pino_ode.md
sathvikbhagavan: This too is not resolved

sathvikbhagavan (commented):
This is almost ready. Some minor comments. I feel the tutorial needs a bit of work - every step need to be explained to make the users clear on how to use it. Currently it feels like an example more than a tutorial. Maybe a new PR would be good as this one got too big.

sathvikbhagavan:
> Do you know how we can add [NeuralOperators](https://github.com/LuxDL/NeuralOperators.jl) to CI while it is not registered?

I am not sure if its possible - the reason being we use it in the package itself, so it will fail in precompilation itself. @avik-pal what's the plan with registering the package?

avik-pal:
Once https://github.com/LuxDL/NeuralOperators.jl/pull/24 is merged, I will rename the old repo and move this version to SciML (and tag a new release)

sathvikbhagavan:
@KirillZubov, This looks good to go, but I do not want to merge until CI is ran successfully. So, have to wait a bit on that. Until then, I see Chris has pinged you to look into https://github.com/SciML/NeuralPDE.jl/issues/883. Can you look into fixing that?

KirillZubov:
> @KirillZubov, This looks good to go, but I do not want to merge until CI is ran successfully. So, have to wait a bit on that. Until then, I see Chris has pinged you to look into #883. Can you look into fixing that?

Hey @sathvikbhagavan  yes, it is related to my task with output vector so yes I going to figure out to it on this week. 
I agree there will be needed sure that all work fine before merge it. I will complete this task, waiting for the remaining parts to be ready in order to merge this PR.

KirillZubov:
Hi @sathvikbhagavan, @ChrisRackauckas I updated this PR for Lux v1.0 and some another update in order to PinoODE tests would  passed for CI. The fact that other tests don't work due to the Lux update is beyond the scope of my summer project. This PR is ready for review.

ChrisRackauckas:
What's left here before review?

KirillZubov:
> What's left here before review?

@ChrisRackauckas all is done here, but it is necessary for the tests to pass, but this is related to Lux updates and not to the code in this PR

ChrisRackauckas:
These two tests don't pass and that doesn't seem lux related? https://github.com/SciML/NeuralPDE.jl/actions/runs/11405946017/job/31738592363?pr=806#step:6:1096 

It just needs to match https://github.com/SciML/NeuralPDE.jl/actions/runs/11395770870/job/31708526179 and make sure the new tests pass, but the new tests you added don't pass.

src/pino_ode_solve.jl
```diff
@@ -0,0 +1,327 @@
...
+             for i in axes(p, 2)])
+    else
+        f_vec = reduce(hcat,
+            [reduce(vcat, [f(out[j, i], p[:, i], t[j]) for j in axes(t, 2)])
+             for i in axes(p, 2)])
+    end
+    du = dfdx(phi, x, θ)
+    norm = prod(size(du))
+    sum(abs2, du .- f_vec) / norm
+end
```
ChrisRackauckas, line 117: why can't this use the ODE code?

src/pino_ode_solve.jl
```diff
@@ -0,0 +1,327 @@
...
+        C <: DeepONet, T}
+    p, t = x
+    t0 = reshape([prob.tspan[1]], (1, 1, 1))
+    x0 = (p, t0)
+    u = phi(x0, θ)
+    u0 = size(prob.u0, 1) == 1 ? fill(prob.u0, size(u)) :
+         reduce(vcat, [fill(u0, size(u)) for u0 in prob.u0])
+    norm = prod(size(u0))
+    sum(abs2, u .- u0) / norm
+end
```
ChrisRackauckas, line 153: why can't this use the ODE code?

ChrisRackauckas:
Seems a bit odd to take over `sol.interp` like that, what does `sol(t)` do in this case?

src/pino_ode_solve.jl
KirillZubov: Here it has parameters and coordinates on input, unlike the usual ODE with just cord, also ODE does not process DeepONet and NeuralOperator in general. It can't be reduced to instance ODE task. If only on the contrary, ODE is a special case of PinoODE as a parametric ODE with one parameter

src/pino_ode_solve.jl
KirillZubov: the same here

KirillZubov:
> Seems a bit odd to take over `sol.interp` like that, what does `sol(t)` do in this case?

it wan't identified `sol(t)`. I added `solt(t)` and replace `interp` with `solt(t)`

KirillZubov:
it is reason why CI fail https://github.com/LuxDL/LuxLib.jl/issues/179

test/PINO_ode_tests.jl
```diff
@@ -0,0 +1,304 @@
...
+    sol = solve(prob, alg, verbose = false, maxiters = 5000)
+    ground_analytic = (u0, p, t) -> u0 + sin(p * t) / (p)
+    dt = 0.025f0
+    p, t = get_trainset(chain, bounds, number_of_parameters, tspan, dt)
+    ground_solution = ground_analytic.(u0, p, t)
+    predict_sol = sol(reduce(vcat, (p, t)))
+    @test ground_solution≈predict_sol rtol=0.05
+    p, t = get_trainset(chain, bounds, 100, tspan, 0.01)
+    ground_solution = ground_analytic.(u0, p, t)
+    predict_sol = sol(reduce(vcat, (p, t)))
```
ChrisRackauckas, line 52: it doesn't match the documented interface though, so it's really odd. `sol.original(...)` would be allowed to do this, but what I was saying is that we shouldn't have an interface break here.

KirillZubov: How do you think it would be better ?

ChrisRackauckas: `sol(t)` should do the normal interoplation w.r.t. `p` in the ODEProblem. You can overload `sol.interp` to do extra things and document that as well, but `sol(t)` is described as having very specific behavior which should be kept the same with all other ODE solvers.

src/pino_ode_solve.jl
```diff
@@ -0,0 +1,380 @@
...
+    else
+        error("Only DeepONet and Chain neural networks are supported with PINO ODE")
+    end
+end
+
+SciMLBase.interp_summary(::PINOODEInterpolation) = "Trained neural network interpolation"
+SciMLBase.allowscomplex(::PINOODE) = true
+
+function (sol::SciMLBase.AbstractODESolution)(t::AbstractArray)
+    p, _ = sol.t
```
ChrisRackauckas, line 278: I don't understand how `.t` would have `p`?

KirillZubov: how can we place t and parameters in ODEsolution separately without override t is how all input data(t and p), or is it better identify PINOODEsolution here?

ChrisRackauckas: I would've guessed you'd do:

```julia
function (sol::SciMLBase.AbstractODESolution)(t::Union{Number,AbstractArray})
  sol.interp(sol.prob.p, t)
end
```

? If that works then this PR is complete.

KirillZubov: yes, it would works  but with Chain it need that `size(t) == size(p)`. I guess, It need add the warning "`t` should be same size as `p`" 

or 
```julia
function (sol::SciMLBase.AbstractODESolution)(t::Union{Number,AbstractArray})
   p = gen(sol.prob.p, size(t)) # generate p same size as t 
   sol.interp(p, t)
end
```

KirillZubov: I've implemented it with the warning

ChrisRackauckas: alright, I think we're good to go.

ChrisRackauckas (approved):

ChrisRackauckas:
🎉 🎉 🎉 🎉 🎉 🎉

KirillZubov:
🥳🥳🥳🥳🥳

PR #803, opened 2024-02-11
Bump optimization packages

ChrisRackauckas:

PR #809, opened 2024-02-14
build: bump Optimization packages and use only latest versions for all

sathvikbhagavan:
Second attempt for https://github.com/SciML/NeuralPDE.jl/pull/803

All packages have the latest versions except DomainSets because Symbolics.jl needs to bump it - https://github.com/JuliaSymbolics/Symbolics.jl/blob/master/Project.toml#L59 to 0.7 first.

Also fixes: #810


sathvikbhagavan:
@ChrisRackauckas, this PR is ready in terms of all compat bumps. DomainSets cannot be bumped until MTK v9 is released and upgraded here.

But I see a significant increase in CI time of `PDEBPINN`, `NNPDE1` & `NNPDE2` in GHA. `NNPDE1` timed out after 6 hours. Buildkite jobs like `NNODE`, `IntegroDiff` also seem to be significantly slow here. I am not sure why. I will test it locally to see if I can get similar times.

PR #815, opened 2024-02-20
docs: update and cleanup tutorials and examples

sathvikbhagavan:


docs/src/examples/nonlinear_elliptic.md
```diff
@@ -79,18 +79,14 @@ input_ = length(domains)
 n = 15
 chain = [Lux.Chain(Dense(input_, n, Lux.σ), Dense(n, n, Lux.σ), Dense(n, 1)) for _ in 1:6] # 1:number of @variables
 
-strategy = QuadratureTraining()
+strategy = GridTraining(0.01)
```
ChrisRackauckas, line 82: We shouldn't default to grid training.

ChrisRackauckas:
Looks like tests fail?

sathvikbhagavan:
Yeah, doc build is timing out. Currently, checking examples which are slow and trying to speed it up.

ChrisRackauckas:
I'm not sure the regular queue would work because some want to use GPUs?

sathvikbhagavan:
>I'm not sure the regular queue would work because some want to use GPUs?

Yes, I was testing out the build times in the normal queue. I have reverted back to GPU queue

ChrisRackauckas:
Do we know which ones take a long time? @mortonpi is it possible to make the documenter debug output spit out timing information?

PR #828, opened 2024-03-07
build: bump mtk to v9

sathvikbhagavan:


sathvikbhagavan:
MOL requires MTK bump to v9 as it is a test dep.

ChrisRackauckas:
Let's follow up with formatting and all that jazz. Glad to get this out there.

Issue #703, opened 2023-07-17
Allow users to write specialized data-free loss functions and use NeuralPDE training strategies


nicholaskl97:
NeuralPDE develops its loss function in two steps:

1. Develop a data-free loss function $\ell(x, \theta)$ from each equation/boundary condition in the `PDESystem`. This is the role of NeuralPDE's parser.
2. Develop a full loss function $L(\theta)$ from the data-free loss functions and the training strategy. This involves developing a training set from the strategy and merging it with the data free loss functions. 

In some cases, a user may want to use 2 with their own data-free loss functions, specialized to their PDE. For example, as noted in #702, directional derivatives aren't currently optimized, so a current user may wish to write their own specialized data-free loss function, which makes this optimization, without rewriting the training strategies provided by NeuralPDE.

Adding this functionality could possibly look like exporting `merge_strategy_with_loss_function`, however that function relies on having a `PINNRepresentation` that was built up in step 1, which might not be the most user-friendly for someone doing step 1 themselves.

@xtalax

nicholaskl97:
@xtalax, I just came up against this again and am wondering if there are any plans (possibly with your parser re-write) to add a public interface for `merge_strategy_with_loss_function` or similar. 

For more context:

I'm working on [SciML/NeuralLyapunov.jl ](https://github.com/SciML/NeuralLyapunov.jl), and my PDE is always something like $\vec{\nabla} V(\vec{x}) \cdot \vec{f}(\vec{x}) < 0$, where $\vec{f}$ is user-defined and we're searching for $V$. My data-free loss is then always $\ell(\vec{x}, \theta) = \max \left( 0, \vec{\nabla} V_\theta(\vec{x}) \cdot \vec{f}(\vec{x}) \right)^2$, just with a different $\vec{f}$ each time. 

In some functionality that I'm currently adding to NeuralLyapunov, I'm hoping to only enforce that PDE in the region $\\{ \vec{x} : V(\vec{x}) \le \rho \\}$. It would be nice to be able to have an `if ... else ...` statement to make the loss as above when $V(\vec{x}) \le \rho$ and $0$ when $V(\vec{x}) > \rho$, but the NeuralPDE parser doesn't like `if ... else ...`.

[...]

PR #834, opened 2024-03-16
Adding NTK adaptive loss function

ayushinav:
## Checklist


ayushinav:
One of the arguments for NTK Adaptive loss would be the kernel size (eg., 4th cell [here](https://github.com/PredictiveIntelligenceLab/PINNsNTK/blob/master/PINNsNTK_Poisson1D.ipynb)). This is the number of points that should be used to make the NTK. This means we have to sample points again when computing the adaptive loss weights. We can use the already sampled points, but under the current implementation that is not possible because the points are generated inside `merge_strategy_with_loss_function` in `discretize.jl`. Also, the kernel size might be different from the batch size, but yeah, we can reuse points instead of sampling.

As a workaround, I tried generating the points inside `generate_adaptive_loss_function`. The issue now is that we don't have access to data free loss functions because the [current implementation](https://github.com/SciML/NeuralPDE.jl/blob/master/src/discretize.jl#L525) only passes the loss functions after taking the mean from the losses of the sampled points. while they are saved for future calls [later on](https://github.com/SciML/NeuralPDE.jl/blob/master/src/discretize.jl#L691), `pinnrep` does not hold them until then.

So, a simpler solution would be to allocate `pinnrep.loss_function` earlier on, after defining `bc_loss_function` and `pde_loss_function` but since we do not have `full_loss_function` and `additional_loss`, we can have them as null functions, so something like
```
pinnrep.loss_functions = PINNLossFunctions(bc_loss_functions, pde_loss_functions,
                                                ()->(), ()->(), 
                                                datafree_pde_loss_functions,
                                                datafree_bc_loss_functions)
```
[here](https://github.com/SciML/NeuralPDE.jl/blob/master/src/discretize.jl#L524) or probably even before but make all the functions as nulls until we get them. Wanted to confirm if it's fine doing this, @ChrisRackauckas @sathvikbhagavan 

If IIUC, the sampler in `StochasticTraining` best resembles their sampler (3rd cell [here](https://github.com/PredictiveIntelligenceLab/PINNsNTK/blob/master/PINNsNTK_Poisson1D.ipynb)). 

The current implementation I have [here](https://github.com/ayushinav/NeuralPDE.jl/blob/ada_loss/src/adaptive_losses.jl#L316-L320) takes the gradient after taking the mse, that is, the gradient of sum of squared errors, we want to take the sum of the squares of the gradients.

ayushinav:
I'd be happy to work on #703 that will help resolve the issue here as well. As I understand now, I feel like it's mostly about making a struct that contains the strategy and the sampled points, maybe `domains` as well?

PR #836, opened 2024-03-21
refactor: correctly lower quadrature training strategy in NNODE

sathvikbhagavan:
Along with correctly lowering quadrature training strategy in NNODE, things done:

1. `batch` keyword is incorrectly described in `NNODE` docstring
2. default tolerances were switched out for `abstol` and `reltol` in quadrature training.
3. Lux.transform is deprecated for FromFluxAdaptor


ChrisRackauckas (approved):

ChrisRackauckas:
It seems the tolerances just need to be changed?

PR #838, opened 2024-03-26
Adding Training Strategies to dae_solvers.jl

hippyhippohops:
Added in StochasticTraining strategy for generate_loss function

## Checklist


src/dae_solve.jl
```diff
@@ -82,6 +82,43 @@ function generate_loss(strategy::GridTraining, phi, f, autodiff::Bool, tspan, p,
...
+    temp_data = rand(1, trunc(Int, points * item)) .* difference .+ minT .+
+                ((index - 1) * difference)
+    data = append!(data, temp_data)
+end
+
+ts = data
+function loss(θ, _)
+    sum(abs2, inner_loss(phi, f, autodiff, ts, θ, p, differential_vars))
+end
+return loss
```
ChrisRackauckas, line 119: the formatting is off here.

src/dae_solve.jl
```diff
@@ -82,6 +82,43 @@ function generate_loss(strategy::GridTraining, phi, f, autodiff::Bool, tspan, p,
...
+    function loss(θ, _)
+        ts = adapt(parameterless_type(θ),
+            [(tspan[2] - tspan[1]) * rand() + tspan[1] for i in 1:(strategy.points)])
+        sum(abs2, inner_loss(phi, f, autodiff, ts, θ, p, differential_vars))
+    end
+    return loss
+end
+
+
+function generate_loss(strategy::WeightedIntervalTraining, phi, f, autodiff::Bool, tspan, p,differential_vars::AbstractVector)
```
ChrisRackauckas, line 96: Looks like a repeat of the ODE code. Is it actually necessary to repeat or could it be refactored so DAEs and ODEs use the same function?

src/dae_solve.jl
hippyhippohops: Okay, I will look into it.

src/dae_solve.jl
hippyhippohops: Yes, I did draw inspiration from the ODE code. I will look into refactoring it so that the same function will be used.

hippyhippohops (commented):
I have refactoed the generate_loss function for ode_solve. I just merged all the functions into one main function and added the conditional statements to carry out the different tasks based on the abstract type of the strategy.

sathvikbhagavan:
No, that was not the issue. The issue was to refactor common computations in `generate_loss` in ODEs and DAEs into common functions such that we don't repeat ourselves.

hippyhippohops (commented):
I have added the WeightedIntervalTraining Strategy in dae_solve.jl and the relevant test in NNDAE_test.jl, which the strategy passes. I would like to push that to be merged. My code has been formatted using JuliaFormatter too.

sathvikbhagavan:
I think we should add rest of the strategies before merging.

sathvikbhagavan:
Closing this in favour of #876

PR #864, opened 2024-06-15
refactor: remove mutable from `ODEPhi` struct

sathvikbhagavan:


sathvikbhagavan:
It is failing making it immutable because `st` is over written in https://github.com/SciML/NeuralPDE.jl/blob/master/src/ode_solve.jl#L123. `st` does not make much difference for feed forward networks but it might for other kinds of networks.

avik-pal:
ODEPhi is no longer mutable

PR #882, opened 2024-08-20
refactor: trigger build with latest Lux

avik-pal:


avik-pal:
@ChrisRackauckas I tracked down and fixed the upstream issues. The only one remaining is https://github.com/SciML/NeuralPDE.jl/actions/runs/10485955321/job/29043159135?pr=882#step:6:1232 but that is kind of hard to handle -- it is matrix multiply with non-concrete element types.

Might be best to get the `adapt` usage fixed on NeuralPDE end that will fix all the issues

sathvikbhagavan:
> Might be best to get the adapt usage fixed on NeuralPDE end that will fix all the issues

I can look into it. Can you open an issue about it? 

Also, in CI, why is precompilation failing - https://github.com/SciML/NeuralPDE.jl/actions/runs/10485955321/job/29043160310?pr=882#step:6:982

avik-pal:
> Also, in CI, why is precompilation failing - [SciML/NeuralPDE.jl/actions/runs/10485955321/job/29043160310?pr=882#step:6:982](https://github.com/SciML/NeuralPDE.jl/actions/runs/10485955321/job/29043160310?pr=882#step:6:982)

Depwarn errors are on, and currently Enzyme doesn't support the latest GPUCompiler which doesn't have those depwarns (same as https://github.com/SciML/DeepEquilibriumNetworks.jl/pull/159)

ChrisRackauckas:
@wsmoses what should we do about the GPUCompiler Enzyme depwarn issue? How close is that to getting the release?

ChrisRackauckas:
@avik-pal you're unblocked here?

avik-pal:
No I need https://github.com/SciML/NeuralPDE.jl/issues/883 to be fixed, else it tries to propagate `Matrix{Any}`

ChrisRackauckas:
Okay, but at least unblocked from non-NeuralPDE things 😅 . @KirillZubov do you think you can have a look at that?

KirillZubov:
@ChrisRackauckas ok, I will try figure out with it , it's a little related to my current task

src/discretize.jl
```diff
@@ -217,20 +217,20 @@ function generate_training_sets(domains, dx, eqs, bcs, eltypeθ, dict_indvars::D
 
     bcs_train_sets = map(bound_args) do bt
         span = map(b -> get(dict_var_span, b, b), bt)
-        _set = adapt(eltypeθ,
+        _set = adapt(Array{eltypeθ},
```
ChrisRackauckas, line 220: Do not force an `Array` type, since that won't work in most scenarios. Just add a post-adapt `convert.(eltypeθ, x)`

KirillZubov:
I updated this PR also to Lux v1.0 #887

KirillZubov:
I’m not sure what else I can do here. I did what I can. May be you can finish it @sathvikbhagavan

KirillZubov:
@sathvikbhagavan @ChrisRackauckas  Unfortunately, I couldn't cope with this task and I don't have time to do it in the near future. In addition to updating the packages(Lux to v1 ) that I've already done, It also need to review all the tests and update the solvers which is need some time on it. I fixed the problem with Adapt and Convert types(#883) in a couple of places. This template can be used for other solvers. There are also other errors, the cause of which I haven't had time to establish.

avik-pal:
need to close https://github.com/SciML/NeuralPDE.jl/pull/885 https://github.com/SciML/NeuralPDE.jl/pull/893 https://github.com/SciML/NeuralPDE.jl/pull/886

avik-pal:
Lets merge once all the tests pass. I think docs build might fail due to linkchecks but it also takes like 17hrs. I will try to go through and see if we can reduce the time there.

avik-pal:
Logging tests fail due to https://github.com/JuliaLogging/TensorBoardLogger.jl/blob/master/src/protojl/tensorboard/tensorboard/event_pb.jl#L145

avik-pal:
I am going to merge and fix the docs in a followup PR

PR #897, opened 2024-10-11
SDE PINN solver

AstitvaAggarwal:
## Checklist


ChrisRackauckas:
That looks to be on the right track.

ChrisRackauckas:
I would think you'd need sample sizes of at least like 100 to be able to smoothly converge?

AstitvaAggarwal:
Fixed the issue and made it a bit faster !

AstitvaAggarwal:
@ChrisRackauckas All the tests pass. lmk, if you think something can be improved so we can get this bad boy finally merged.

AstitvaAggarwal:
How results look like (sol.solution.interp returns a smooth interpolation. Hence, both 2000 test samples case plot appears smooth, In the 1000 test sample case the solver by default returns sol.mean_fit - mean over n=1000 rough path solutions): 
image
image
image
image

ChrisRackauckas:
This looks correct, though it needs a bit more things tested:

1. Test the strong convergence, i.e. with the same noise process it should give a similar result to the normal SDE solver
2. On the first SDE from https://arxiv.org/abs/1804.04344, what kind of behavior does the solver give?

AstitvaAggarwal:
@ChrisRackauckas sorry for the delay, the solver was failing miserably for the additive noise tests. Turns out I was performing the initial condition transform twice! Finally figured where the code was going wrong ! I'll be posting the results shortly (yes the new results are pretty good as well)

AstitvaAggarwal:
@ChrisRackauckas I have added tests for strong and weak convergence  in case of both GBM and the Additive Noise Test equation in the testing file. Note that we can also train the SDEPINN via strong or weak convergence, I am defaulting to weak convergence ( basis: strong convergence implies weak convergence, but easier to optimize weak solution over the same n iterations). Let me know what you think on this or want more testing on this part.

Ive used MonteCarloMeasurements.Particles for the returned solutions (BPINNs use this as well). Using this the difference in larger vs smaller sub sample sizes is much clear, the confidence intervals are better for more sub samples (in almost all cases they are equal or within the confidence intervals of the truncated/analytical solutions) . The Diffusion term is learnt much well for more sub samples (makes sense as we basically learn the Wiener basis better).

Enforcing weak solution while training.
image

A) Sub samples=1:
image

B) Sub samples=10:
image

AstitvaAggarwal:
And all the related tests pass.

test/NN_SDE_tests.jl
```diff
@@ -0,0 +1,301 @@
...
+                                   for i in eachindex(ts)]
+
+    error_1 = sum(abs2, strong_analytic_solution .- strong_predicted_solution_1)
+    error_2 = sum(abs2, strong_analytic_solution .- strong_predicted_solution_2)
+    @test pmean(error_1) > pmean(error_2)
+
+    @test pmean(sum(abs2.(strong_predicted_solution_1 .- strong_truncated_solution))) >
+          pmean(sum(abs2.(strong_predicted_solution_2 .- strong_truncated_solution)))
+
+    # weal solution tests
```
ChrisRackauckas, line 135:
```suggestion
    # weak solution tests
```

ChrisRackauckas (approved):

ChrisRackauckas:
This looks spot on. Though needs docs?

AstitvaAggarwal:
yeah docs are next up.

ChrisRackauckas:
let's merge and follow up with docs.

PR #956, opened 2025-06-29
SDE-PINN inverse + strong/weak + improved base solver & docstrings.

### LINKED ISSUE #968: Bug in the Bayesian inference for PINNs
**Describe the bug 🐞**

The method `getlogpdf` dispatched on `GridTraining` constructs the grid of points to evaluate the PINN at. 
```
ts = collect(eltype(strategy.dx), tspan[1]:(strategy.dx):tspan[2])
```
To this time vector, it appends `dataset[end - 1]` (in the case of an inverse problem) 
```
  t = isempty(ltd.dataset) ? ts : vcat(ts, ltd.dataset[end - 1])
```
The problem is the `dataset` has the form `[x, y, time]` (in the 2d case) and so it appends tot he time vector `ltd.dataset[end - 1]` which is not time but the independent variable `y`. 

I believe this is a bug. Should it not concatenate the last element of `dataset`? I have dataset defined as 
```
# let's add some noise 
x = u[1, :] + (u[1, :]) .* (0.3 .* randn(length(u[1, :])))
y = u[2, :] + (u[2, :]) .* (0.3 .* randn(length(u[2, :])))
dataset = [x, y, time] # noisy data [x, y, time]
```

### LINKED ISSUE #970: Order of logpdf evaluation in BPINNs is incorrect
**Describe the bug 🐞**

Suppose you have 4 ODE parameters and for each one, there is a prior assigned: 
```
invpriors= [        
        Normal(1, 2), 
        Normal(2, 2), 
        Normal(2, 2),
        Normal(0, 2)]
```
Also suppose you have a vector of parameters `θ = [weights..., biases..., λ1, λ2, λ3, λ4]` where `λ1, λ2, λ3, λ4` are parameter values of the ODE system. The function `priorweights()` on line 247 of `advancedHMC_MCMC.jl` evaluates the log PDF at these parameter values: 
```
invlogpdf = sum(
        logpdf(invpriors[length(θ) - i + 1], θ[i])
    for i in (length(θ) - ltd.extraparams + 1):length(θ))
```
Except the indices are incorrect. Unrolled, this will evaluate `param[4](λ1)`, `param[3](λ2)`, `param[2](λ3)` and `param[1](λ4)`. In many cases this is probably not a big deal, but still could be fixed quickly. 

**Expected behavior**
Should evaluate `param[1](λ1)`, `param[2](λ2)`, `param[3](λ3)` and `param[4](λ4)`. The fix seems easy: Replace 
```
 logpdf(invpriors[length(θ) - i + 1], θ[i])
``` 
with 
```
logpdf(invpriors[i - length(θ) + ltd.extraparams], θ[i])
```
or some other cleaner version. Is this worth fixing?  Happy to submit a quick PR.

AstitvaAggarwal:
## GSoC 2025 FINAL REPORT

Contributor Name: Astitva Aggarwal
Organization: NumFOCUS - SciML
Mentors: Chris Rackauckas, Sathvik Bhagavan.
Project: Multidimensional SDE PINN solvers.

## About the Project

The goal of this project was to extend and document the existing SDE Physics-Informed Neural Network (PINN) solver in the NeuralPDE.jl library, enabling it to efficiently handle multidimensional Stochastic Differential Equations (SDEs) and potentially Stochastic Partial Differential Equations (SPDEs). This was achieved by implementing solver formulations based on the Kosambi-Karhunen-Loève (KKL) expansion for SDEs, which transforms stochastic systems into tractable rough path ODEs, generalizing this approach to SPDEs we can use Polynomial Chaos Expansion (PCE) for spatially distributed noise.
The project also provided comprehensive documentation, inverse problem-solving support, and validation test examples, making NeuralPDE.jl a more powerful tool for modeling and simulating complex stochastic dynamical systems.

## Checklist

- [x] Appropriate tests were added
- [x] Any code changes were done in a way that does not break public API
- [x] All documentation related to code changes were updated
- [x] The new code follows the
- [x] Any new documentation only uses public API
  
## Work and Tasks achieved.

- [x] The Base SDE-PINN solver had the following issues (Fixed them) :
    -  was not efficient in input passing to `sdephi`.
    -  Batching performance was abysmal (not actually performing batching).
    -  non-batching case was also slow.
    -  issues with `inner_sde_loss` loss construction handling.
    -  `du_dt` (like `sdephi` and `inner_sde_loss`) handled only Nested Vector inputs (with nested loops at that).
    - the comments in the code were not clear and very complicated.

- [x] The SDE-PINN solver did not support Inverse solving (`additional_loss` is inadequate). Needed a new loss specification :
  - Formulated and implemented 2 types of losses :
      - Naive Moment Matching of SDE-PINN outputs with the dataset provided. 
      - Moment matching for Distribution got via Euler-Maruyama discretization scheme applied for SDE Parameter Estimation.

- [x] The SDE-PINN solver **ONLY** supported weak-solving SDEs, now it can solve for the strong solution at `n` fixed paths :
    - `add_rand_coeff_2` for strong solving.
    - `add_rand_coeff` for weak solving - This shows pretty good results for the expected solution to the SDE, i still need to figure out the mathematics but in my opinion it has to do with the law of large Numbers, Monte Carlo sampling for the Random coefficients basis in the truncated KKL expansion.

- [x] Added fields to the returned solver object :
    - estimated SDE parameters.
    - Exact input path information used during training (strong/weak solves are different)
    - `sdephi` inputs created from Dataset used for Inverse solves.
    - Random ODE solution object. (contains `sdephi`, trained parameters etc.) via `sol.rode_solution`.
    - solution's `ensemble_fit` and `ensemble_inputs` used for solution validation.
    - `estimated_sol` field contains `MonteCarloMeasurements.Particles` objects created from `ensemble_fit`. It is a SDE-PINN output sized Vector.

- [x] Added and improved existing docstrings for the solver.

A few plots showing Inverse + strong solving the Geometric Brownian Motion SDE with a dataset consisting of `n=15` paths, `dt=0.01` in the timespan `(0, 1)`. The solve/training is done on `1` generated strong path (we solve over the truncation, therefore generate a set of random coefficients `[z_1, z_2, .....z_i]` for each path).
The number of paths can be controlled via `sub_batch` arg to  `NNSDE`.

image
image

(Ran the solve twice)

Estimated Parameters (the solver's loss formulation for the diffusion parameter is sign invariant, therefore might get a mixed up `+/-` for the final estimated parameter). Here the dataset is as described above (to get values closer to `ideal_p` increase `n`), the PINN collocation loss had a discretization `dt=0.02`. parameter initialization: `[0.0, 0.0]`.
```
julia> ideal_p
2-element Vector{Float64}:
 1.5
 0.5

julia> sol_2.estimated_params
2-element Vector{Float64}:
 1.2268535217978696
 -0.5335091621614064

julia> sol_1.estimated_params
2-element Vector{Float64}:
 1.226853523016635
 0.5335091597301435
```

- [x] Additionally this PR closes #970 (fix taken from https://github.com/SciML/NeuralPDE.jl/issues/970#issuecomment-3225797423) and closes #968. credit to @affans for finding these bugs. 
    - for #968 Ive updated the docstrings and kept all the existing and new dataset checks, proofing consistent with the pre-existing tutorials.
    - Further apart from these two issues, we can now use the L2 loss and Data Quadrature loss loglikelihoods even if a dataset is provided during the forward solve (although a dataset is neccesary for Inverse solves, it is optional but helpful for forward solves).

## Future Scope

- Extend to solving multidimensional SDEs via a loss formulation based on the Kolmogorov Forward Equation.
-  Write a SPDE PINN solver that uses the Polynomial Chaos Expansion for loss formulation. This again must solve inverse and forward problems with both the strong and weak solutions (future direction).

## Acknowledgements
This Project has been very interesting and intellectually fulfilling. The process of moving from theory and ideas to actual efficient implementation has been an amazing journey.
 
I would like to thank Chris Rackauckas for his continued support during the project. We had weekly meetings where we would interact and discuss the status of my project, here I would also have the chance to learn about other people's project within SciML. There were many instances where his advice has helped me get out of theoretical or implementation roadblocks.
I have been a long time contributor to SciML and am confident that this will continue going forward into the future.


src/NN_SDE_solve.jl
```diff
@@ -1,31 +1,166 @@
...
+
+Note that the returned weak solution is evaluated at fixed time points according to standard output
+handlers such as `saveat` and `dt`. However, the neural network is a fully continuous
+solution so `sol(t)` is an accurate interpolation (up to the neural network training
+result). In addition, the `OptimizationSolution` is returned as `sol.k` for further
+analysis.
+
+## References
+
+Stochastic Physics-Informed Neural Ordinary Differential Equations : https://arxiv.org/abs/2109.01621
```
ChrisRackauckas, line 126: I'm not sure this reference makes sense since you're not doing a moment closure type of approach.

ChrisRackauckas:
This is looking correct. I think mentioning that reference is a little misleading in some sense because it's advocating an approach of doing an unscented transform to an ODE and then solving that, which isn't what you're doing here. But I assume that you referenced it because you're doing a moment matching fitting, which 🤷  is reasonable.

PR #959, opened 2025-07-15
Fix multioutput `PhysicsInformedNN` construction

nicholaskl97:
## Checklist

- [x] Any code changes were done in a way that does not break public API
- [x] All documentation related to code changes were updated
- [x] The new code follows the
- [x] Any new documentation only uses public API
  
## Additional context

Fixes a bug I introduced in #955, which caused `PhysicsInformedNN` to fail to construct in the multioutput case.

### DIFF
--- a/src/pinn_types.jl
+++ b/src/pinn_types.jl
@@ -124,7 +124,9 @@ function PhysicsInformedNN(
 
     phi = phi === nothing ?
         (multioutput ?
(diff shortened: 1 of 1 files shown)


nicholaskl97:
@ChrisRackauckas, I found that I accidentally introduced a bug in #955 that causes construction of `PhysicsInformedNN` to fail in the multioutput case. This PR fixes it.

PR #961, opened 2025-07-21
Bump NeuralOperators.jl to v0.6

ChrisRackauckas:


ChrisRackauckas:
@avik-pal I threw claude Opus on this:

```
Summary

  I've investigated the NeuralOperators.jl v0.6 update for NeuralPDE.jl.
  Here's what I found:

  Key Breaking Changes in NeuralOperators.jl v0.6:

  1. Removed permuted kwarg (default now corresponds to previous
  permuted=Val(true))
  2. Removed direct dependency on FFT packages
  3. Switched primary backend to Lux + Reactant

  NeuralPDE.jl Status:

  - Good news: NeuralPDE.jl doesn't use the permuted parameter, so that
  breaking change doesn't affect it
  - Good news: The Project.toml already specifies compatibility with
  NeuralOperators v0.6
  - Issue found: There's a bug in NeuralOperators.jl v0.6 where DeepONet
  fails with a dimension mismatch error when performing matrix
  multiplication between 2D and 3D arrays

  The Bug:

  DeepONet in v0.6 attempts to multiply:
  - Branch output (after transpose): 2D matrix of shape (5, 10)
  - Trunk output: 3D array of shape (10, 10, 1)

  This operation is not defined in Julia and causes a MethodError.

  Recommendation:

  The NeuralPDE.jl code itself doesn't need any changes for v0.6
  compatibility. However, there's a bug in NeuralOperators.jl v0.6 that
  prevents DeepONet from working correctly. This issue should be reported
  to the NeuralOperators.jl repository so it can be fixed in a patch
  release (v0.6.1).
```

Are you aware of this? Is it missing a breaking change?

PR #1066, opened 2026-06-06
Re-enable downgrade (strict, allow_reresolve=false); intentionally RED pending upstream StatsFuns 2.0.1 vs MCMCChains — red until MCMCChains upstream allows StatsFuns 2.x (#1023); NOT fixed by ForwardDiff — natural failure, not disabled; auto-greens on upstream release.

ChrisRackauckas-Claude:
## What

Re-enables the strict downgrade job (`allow_reresolve=false`) for NeuralPDE.jl by removing the `if: false` gate in `.github/workflows/Downgrade.yml`, so the job RUNS and fails NATURALLY (visible red) instead of being silently skipped.

## Why it is RED (intentional, not disabled)

The strict resolve currently cannot resolve here: StatsFuns 2.0.1 vs MCMCChains — red until MCMCChains upstream allows StatsFuns 2.x (#1023). This is NOT fixed by ForwardDiff. The job is expected to be RED until that upstream fix releases, at which point it will auto-green. No `allow-reresolve` was added, no compat floors were lowered, and julia stays at "1.10".

The failure is deliberately surfaced (natural red) rather than hidden behind a skip, per the repo owner's policy change.

See SciML/NeuralPDE.jl#1023.

---

Please ignore this PR until reviewed by @ChrisRackauckas.


ChrisRackauckas-Claude:
Pushed a follow-up commit removing the now-invalid `allow-reresolve:` input from `Downgrade.yml`. The reusable `SciML/.github` `downgrade.yml` workflow is being changed to always use `allow_reresolve: false` and to drop the `allow-reresolve` `workflow_call` input entirely (see SciML/.github PR). Once that .github PR merges and `@v1` is retagged, any caller still passing `allow-reresolve:` would error with an invalid-input failure, so this line must go. Behavior is unchanged (it was already strict/false here).

PR #1077, opened 2026-06-23
Fix NNSDE1 noise-grid off-by-one (DiffEqNoiseProcess endpoint change)

ChrisRackauckas-Claude:
**Ignore until reviewed by @ChrisRackauckas.**

## Problem

The `tests / NNSDE1` lane (julia 1.11 + lts) is red on master with:

```
Test-2 GBM SDE: Error During Test at test/NNSDE1/nn_sde__test_2_gbm_sde.jl:4
  DimensionMismatch: tried to assign 52-element array to 51×1 destination
```

at `W_samples[:, i] = Np_sol.u` (line 72).

## Root cause

The NNSDE1 tests size `W_samples` to `length(ts)` (the NNSDE solution's time grid) and fill each column with `Np_sol.u` from a `NoiseProblem` solved over `(0.0, 1.0)` with the same Float32 `dt`.

`DiffEqNoiseProcess#278` ("Fix NoiseProblem solve overshooting tspan[2]", v5.32.0) made the noise solve land *exactly* on `tspan[2]`. With a Float32 `dt` (e.g. `1/50.0f0`), the grid `0.0:dt:1.0` ends just short of `1.0` (at `0.99999998`), so the noise solve now appends one extra corrected point at exactly `1.0`:

* `dt = 1/50.0f0`:  `length(ts) = 51`, `length(Np_sol.u) = 52`
* `dt = 1/100.0f0`: `length(ts) = 101`, `length(Np_sol.u) = 102`

This is a behavior change (not a NeuralPDE regression): the same off-by-one reproduces back to StochasticDiffEq 6.69.1; the NNSDE1 lane was last green (2026-06-03) with DiffEqNoiseProcess **5.31.1**, whose noise solve stopped at `0.99999998` (51 points, matching `ts`). 5.32.x correctly lands on `1.0`, adding the extra point.

## Fix

Sample the noise over the solution's actual grid: solve the `NoiseProblem` on `(0.0, ts[end])` instead of `(0.0, 1.0)`. The noise grid then equals `ts` exactly (verified `Np_sol.t == ts`), which is the test's intent (`W_samples[i, :]` pairs with `ts[i]`). Applied to test_2/test_3/test_4 (all share the pattern).

## Verification

Run locally on Julia 1.10 (lts) with the CI dependency stack (DiffEqNoiseProcess 5.32.2, StochasticDiffEq 7.0.0, Lux 1.31.4):

* baseline (unmodified): `DimensionMismatch` error reproduced
* `test_2 GBM SDE`: **11/11 Pass**
* `test_3 Additive Noise`: **8/8 Pass**
* `test_4 Inverse weak & strong`: **8/8 Pass**

### DIFF
--- a/test/NNSDE1/nn_sde__test_2_gbm_sde.jl
+++ b/test/NNSDE1/nn_sde__test_2_gbm_sde.jl
@@ -67,7 +67,11 @@ using Test
     W_samples = Array{Float64}(undef, num_time_steps, num_samples)
     for i in 1:num_samples
         W = WienerProcess(0.0, 0.0)
-        probtemp = NoiseProblem(W, (0.0, 1.0))
+        # Solve the noise over the actual `ts` grid (which a Float32 `dt` ends just
+        # short of `1.0`) so the path is sampled at exactly `ts`. Using `(0.0, 1.0)`
+        # makes the solver land an extra point on `1.0` (DiffEqNoiseProcess#278), which
+        # no longer matches `length(ts)`.
+        probtemp = NoiseProblem(W, (0.0, ts[end]))
         Np_sol = solve(probtemp; dt = dt)
         W_samples[:, i] = Np_sol.u
     end
--- a/test/NNSDE1/nn_sde__test_3_additive_noise_test_equation.jl
+++ b/test/NNSDE1/nn_sde__test_3_additive_noise_test_equation.jl
@@ -74,7 +74,11 @@ using Test
     W_samples = Array{Float64}(undef, num_time_steps, num_samples)
     for i in 1:num_samples
         W = WienerProcess(0.0, 1.0)
-        probtemp = NoiseProblem(W, (0.0, 1.0))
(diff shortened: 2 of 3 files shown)


ChrisRackauckas-Claude:
Superseded by a root-cause fix in DiffEqNoiseProcess: SciML/DiffEqNoiseProcess.jl#291.

The off-by-one is not a NeuralPDE issue — it's that `solve(::NoiseProblem; dt)` appended a spurious sub-`dt` sliver point past the last grid point when `dt` is a Float32 (`1/50f0`) over a Float64 tspan. DiffEqNoiseProcess#291 scales the end-of-span tolerance by the coarser of the time type's and `dt`'s precision, so the noise solve now yields exactly `length(0.0:dt:1.0)` samples landing on `tspan[2]`. With that release the NNSDE1 tests pass unmodified, so this test-side workaround can be dropped/closed.

**Ignore until reviewed by @ChrisRackauckas.**

PR #1090, opened 2026-07-15
Retarget downstream tests to NeuralLyapunov

ChrisRackauckas-Claude:
Ignore this draft until it has been reviewed by @ChrisRackauckas.

## Summary

- replace the stale `PDESystemLibrary.jl/NeuralPDE` downstream cell with `NeuralLyapunov.jl/Core`
- retain substantive downstream coverage: NeuralLyapunov's four Core test files import NeuralPDE and construct `PhysicsInformedNN` discretizations

## Root cause

PDESystemLibrary no longer declares a `NeuralPDE` test group. Its SciMLTesting migration in `2d30c12a25e6be822d7e1858508c4c07e91e61cc` intentionally dropped the dead NeuralPDE matrix cells and adopted strict folder-group validation. NeuralPDE's downstream workflow continued requesting that removed group, so every run failed before executing a test:

```text
ArgumentError: run_tests (folder mode): GROUP="NeuralPDE" ... is not a group declared in test_groups.toml (declared: ["Core", "QA"])
```

An automated repository bisect identified `2d30c12a25e6be822d7e1858508c4c07e91e61cc` as the first commit where the old workflow request became invalid.

## Verification

The old target was reproduced locally on Julia 1.12.6:

```sh
GROUP=NeuralPDE julia +1.12 --project=pdesystemlibrary -e 'using Pkg; Pkg.develop(PackageSpec(path="neuralpde")); Pkg.update(); Pkg.test(; coverage=false)'
```

It exited 1 with the missing-group error above.

The replacement downstream target was tested with the feature checkout developed into NeuralLyapunov:

```sh
GROUP=Core julia +1.12 --project=neurallyapunov -e 'using Pkg; Pkg.develop(PackageSpec(path="neuralpde")); Pkg.update(); Pkg.test(; coverage=false)'
```

It exited 0 with substantive summaries:

```text
Damped simple harmonic oscillator                     4 / 4 passed
Damped pendulum                                       9 / 9 passed
Damped pendulum - AdditiveLyapunovNet structure       9 / 9 passed
Damped pendulum - MultiplicativeLyapunovNet structure 9 / 9 passed
Testing NeuralLyapunov tests passed
```

Additional checks:

```text
actionlint .github/workflows/Downstream.yml                 passed
julia +1.12 --project=../runic-env -m Runic --check .       passed
git diff --check                                             passed
```

PR #1091, opened 2026-07-15
Stabilize gradient scale adaptive loss test

ChrisRackauckas-Claude:
Ignore until reviewed by @ChrisRackauckas.

## Summary

- restore the 32-neuron hidden width that made this stochastic convergence test reliable before #903
- construct a fresh `GradientScaleAdaptiveLoss` for each logger variant so mutable adaptive-loss state cannot leak between solves

## Root cause

The test became unreliable when #903 simultaneously increased the hidden width from 32 to 40 and reduced its optimization budget. On current dependencies, the original seed and 40-neuron network reproducibly finish at relative error 0.643115, above the 0.4 assertion. An A/B run changing only the width back to 32 finishes at 0.207279. The test also reused one mutable adaptive-loss object across two independent solves, which made the logger comparison order-dependent.

## Validation

- clean `master`, Julia 1.11: `GROUP=AdaptiveLoss julia --project -e 'using Pkg; Pkg.test(; coverage = true)'` failed 2 of 4 assertions in this file (relative errors 0.643115 and 0.520205)
- this branch, Julia 1.11, exact native command: `GROUP=AdaptiveLoss julia --project -e 'using Pkg; Pkg.test(; coverage = true)'` passed all 12 tests across the AdaptiveLoss group
- this branch, Julia 1.10, exact native command: `GROUP=AdaptiveLoss julia --project -e 'using Pkg; Pkg.test(; coverage = true)'` passed the complete AdaptiveLoss group
- direct fresh-process run of the changed test passed 4/4 assertions and produced identical adaptive-loss logs for the logger and no-logger variants
- repository-wide `Runic --check .` passed

### DIFF
--- a/test/AdaptiveLoss/adaptive_loss__2d_poisson_gradientscaleadaptiveloss.jl
+++ b/test/AdaptiveLoss/adaptive_loss__2d_poisson_gradientscaleadaptiveloss.jl
@@ -10,7 +10,7 @@ module AdaptiveLossTestSetup
         logger = haslogger ? TBLogger(logdir) : nothing
 
         Random.seed!(60)
-        hid = 40
+        hid = 32
         chain = Chain(Dense(2, hid, tanh), Dense(hid, hid, tanh), Dense(hid, 1))
         strategy = StochasticTraining(256)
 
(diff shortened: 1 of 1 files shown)


ChrisRackauckas-Claude:
CI classification for IntegrationTest job 87345799986:

The failure is unrelated to this PR. Current `master` still configures PDESystemLibrary with `GROUP=NeuralPDE`; the target now declares only `Core` and `QA`, so strict SciMLTesting validation correctly errors before running tests:

```text
GROUP="NeuralPDE" ... is not a group declared in test_groups.toml (declared: ["Core", "QA"])
```

The stale downstream target is fixed independently in draft PR #1090, which retargets the job to NeuralLyapunov/Core. I am keeping this PR focused rather than stacking that unrelated workflow change here.

PR #1101, opened 2026-08-07
fix(NNODE): fix GPU GridTraining strategy and optimize ODE evaluation

AntiNeutrino27:
- Fixed a typo in `GridTraining` strategy causing GPU execution failures. The `StepRangeLen` type of the input `ts` array failed to transfer to the GPU properly during `phi` invocations.
- Optimized VRAM allocations by moving input array GPU placement into the loss generation phase.
- Added an `ode_batch_eval` configuration parameter to execute the ODE right-hand side (RHS) in a vectorized manner. This mode is automatically enforced when executing on a GPU device.

### DIFF
--- a/src/ode_solve.jl
+++ b/src/ode_solve.jl
@@ -2,7 +2,7 @@ abstract type NeuralPDEAlgorithm <: SciMLBase.AbstractODEAlgorithm end
 
 """
     NNODE(chain, opt, init_params = nothing; strategy = nothing, autodiff = false,
-        batch = true, param_estim = false, additional_loss = nothing,
+        batch = true, ode_batch_eval = true, param_estim = false, additional_loss = nothing,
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
This needs to be reverted because it fails to follow the standard SciML ODEProblem interface. The batching should instead take place via Reactant tracing.

RomanSahakyan03:
@ChrisRackauckas One of the projects I do now is implementing PINN on Enzyme + Rectant AD where the derivative is computing of the NN via Enzyme forward mode AD and the loss backprop computing via Enzyme reverse mode AD. I can change for the NNODE from AutoZygote() to the mentioned part. Also adding the batch part.

ChrisRackauckas:
Okay yeah if you already have this working, go for it.

PR #1120, opened 2026-08-23
fix: make NNODE/PINOODE/NNSDE/SDEPINN subtype the SciMLBase algorithm hierarchy

ChrisRackauckas:
## Failing lanes

At master `62ff9be8` ("Add PDE precompilation workload (#1119)"):

- `tests / NNODE (julia 1.11, ubuntu-latest)` — [job 96902117865](https://github.com/SciML/NeuralPDE.jl/actions/runs/32519247903/job/96902117865)
- `tests / NNODE (julia lts, ubuntu-latest)` — [job 96902117912](https://github.com/SciML/NeuralPDE.jl/actions/runs/32519247903/job/96902117912)

Both abort on `test/NNODE/nnode__nnode_translating_from_flux.jl`, the first file in
the group that constructs an `NNODE` (the two `nndae__*` files that run before it use
`NNDAE`, which was unaffected).

Note these are the only two lanes the NNODE group runs on — `test/test_groups.toml`
declares `[NNODE] versions = ["1.11", "lts"]`. There is no `julia 1` NNODE lane, so
the failure is **not** Julia-version specific.

## Actual error

```
NNODE: Translating from Flux: Error During Test at test/NNODE/nnode__nnode_translating_from_flux.jl:4
  Got exception outside of a @test
  ArgumentError: Passing a `Bool` for `verbose` is no longer supported in OrdinaryDiffEq v7: `verbose` now takes a verbosity object.

      solve(prob, alg; verbose = DEVerbosity(SciMLLogging.None()))  # was verbose = false
      solve(prob, alg; verbose = DEVerbosity())                     # was verbose = true
  Stacktrace:
    [1] _process_verbose_param
      @ DiffEqBase ~/.julia/packages/DiffEqBase/mLfsc/src/verbosity.jl:356 [inlined]
    [2] #_ode_init#99
      @ OrdinaryDiffEqCore ~/.julia/packages/OrdinaryDiffEqCore/ICc50/src/solve.jl:231 [inlined]
    [3] __init(prob::ODEProblem{...}, alg::OrdinaryDiffEqCore.CompositeAlgorithm{...}, ...)
    [4] __solve(::ODEProblem{...}, ::OrdinaryDiffEqCore.CompositeAlgorithm{...}; ...)
    [5] solve_call(_prob::ODEProblem{...}, args::OrdinaryDiffEqCore.CompositeAlgorithm{...}; ...)
[...]

PR #1122, opened 2026-08-26
fix(NNODE): restore pointwise ODE RHS evaluation by default

ChrisRackauckas-Claude:
## Dependency

This PR is stacked on the standalone mechanical formatting draft https://github.com/SciML/NeuralPDE.jl/pull/1125. Until that draft merges, this branch contains its formatting commit followed by the behavior-only commit; after it merges, this PR reduces to the behavior commit alone.

Ignore this PR until it has been reviewed by @ChrisRackauckas.

## What changed and why

The GPU optimization merged in https://github.com/SciML/NeuralPDE.jl/pull/1101 changed `ode_batch_eval` from `false` to `true` and inferred the device from `ODEPhi`. `ODEPhi` has no array payload, so `MLDataDevices.get_device(phi)` is `UnknownDevice` even when the trainable parameters are on the CPU. Ordinary pointwise ODE right-hand sides were therefore called once with a state matrix and time vector.

This restores pointwise CPU evaluation as the default, determines the device from the actual parameter container inside each loss call, and keeps automatic batched RHS evaluation for non-CPU parameters or explicit `ode_batch_eval = true`. The wrapped RHS is local to each call, avoiding mutation of the captured function in stochastic training.

## Failing before

On clean master `dd792519`, the existing complex NNODE test errored for all three batched training strategies:

```text
DimensionMismatch: Batched RHS returned size (4,); expected (4, N).
ODE Complex Numbers | Error 3 Total 3
```

The new default assertion also discriminates directly:

```text
Test Failed
Expression: !((NNODE(chain, Adam(0.01))).ode_batch_eval)
```

Simply restoring the default was not sufficient. The first partial patch still derived the device from `phi`, and the existing scalar CPU case then failed with:

```text
MethodError: no method matching exp(::Vector{Float32})
```

The measured device values explain the failure:

```text
get_device(phi) = MLDataDevices.UnknownDevice()
get_device(phi.smodel) = MLDataDevices.UnknownDevice()
safe_get_device(θ) = CPUDevice{Missing}()
```

## Passing after

Commands were run on Julia 1.12.7 with a local depot and project:

```sh
GROUP=NNODE julia --project -e 'using Pkg; Pkg.test()'
GROUP=QA julia --project -e 'using Pkg; Pkg.test()'
```

Observed results:

- Exact scalar regression: `ODE Example 2 | Pass 4 Total 4 Time 6m53.4s`.
- Exact complex regression: `ODE Complex Numbers | Pass 5 Total 5 Time 52m17.7s`.
- Full NNODE rerun: DAE I 1/1, DAE II 1/1, Flux translation 2/2, complex 5/5 in 37m43.4s, example 2 4/4, example 3 1/1, and ODE I 4/4. The runner then reached an independent master failure: `UndefVarError: BFGS not defined` in `test/NNODE/nnode__ode_parameter_estimation.jl` after OptimizationOptimJL 0.4.19 stopped exporting that owner name.
- Every NNODE file after that independent blocker was run directly in the same locally developed environment: scalar 2/2, other strategies 3/3, weighted interval 1/1, `tstops` 6/6, and vector 12/12. The second parameter-estimation file independently has the same `BFGS` owner-import failure. Those two test-import fixes are intentionally separate from this source fix.
- Exact branch QA reports only the pre-existing `PDETimeSeriesSolution` / PINOODE ambiguity on master (23 pass, 1 fail). With the one-line ambiguity fix from the separate #1120 remediation stack temporarily overlaid, QA is `24/24` in 4m18.6s. The overlay was removed before this commit.
- Runic left every changed range unchanged; `typos --diff` and `git diff --check` exited cleanly.

## Not verified here

- No GPU was available, so the CUDA path was not executed locally. The non-CPU device branch remains explicit and always selects `BatchedRHS`.
- The full docs build is currently blocked on master by the independent OptimizationOptimJL 0.4.19 direct-owner migration across documentation examples. This commit has been handed to that focused docs branch for a temporary full-build overlay; no docs failure is being silenced here.
- The two NNODE `BFGS` test ownership fixes are being handled in a separate direct-owner PR.

AI-Agent: OpenAI Codex

Codex-Session-ID: 01a03a11-47c2-77e0-858b-167004f0b79c

## Links

- Introducing PR: https://github.com/SciML/NeuralPDE.jl/pull/1101
- Introducing commit: https://github.com/SciML/NeuralPDE.jl/commit/dd792519540cde8c3957612a0c5354116002da61
- Separate algorithm-supertype PR: https://github.com/SciML/NeuralPDE.jl/pull/1120

https://claude.ai/code/session_01Rmh6B9eoW3N8oCbuuVPVxJ

### DIFF
--- a/src/ode_solve.jl
+++ b/src/ode_solve.jl
@@ -40,7 +40,7 @@ abstract type NeuralPDEAlgorithm <: SciMLBase.AbstractODEAlgorithm end
 
 """
     NNODE(chain, opt, init_params = nothing; strategy = nothing, autodiff = false,
-        batch = true, ode_batch_eval = true, param_estim = false, additional_loss = nothing,
+        batch = true, ode_batch_eval = false, param_estim = false, additional_loss = nothing,
         dataset = [], estim_collocate = false, kwargs...)
 
 Algorithm for solving ordinary differential equations using a neural network. This is a
@@ -82,6 +82,9 @@ standard `ODEProblem`.
            neural network is done at individual time points one at a time. This is not
            applicable to `QuadratureTraining` where `batch` is passed in the `strategy`
            which is the number of points it can parallelly compute the integrand.
+* `ode_batch_eval`: Whether the ODE right-hand side accepts a matrix of states and a vector
+                    of times. This is always enabled on a GPU and defaults to `false` on a
+                    CPU, where ordinary pointwise ODE functions are supported.
 * `param_estim`: Boolean to indicate whether parameters of the differential equations are
                  learnt along with parameters of the neural network.
 * `strategy`: The training strategy used to choose the points for the evaluations.
@@ -143,7 +146,7 @@ end
 
 function NNODE(
         chain, opt, init_params = nothing; strategy = nothing, autodiff = false,
-        batch = true, ode_batch_eval = true, param_estim = false, additional_loss = nothing,
+        batch = true, ode_batch_eval = false, param_estim = false, additional_loss = nothing,
         dataset = [], estim_collocate = false, kwargs...
     )
     chain isa AbstractLuxLayer || (chain = FromFluxAdaptor()(chain))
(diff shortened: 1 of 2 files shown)


ChrisRackauckas-Claude:
The branch has been rewritten after fetching both upstream and fork and verifying the remote head contained only the two agent commits. The update used an exact `--force-with-lease` against `2f1e83d3d8fd030c08abd053f8c720d074e5222c`.

Current stack:

1. current NeuralPDE master (`1c90d66d`), including https://github.com/SciML/NeuralPDE.jl/pull/1120;
2. the standalone formatting commit from https://github.com/SciML/NeuralPDE.jl/pull/1125;
3. behavior-only commit `ead944970a9ab8a3387293f8b1dcf000f1bebed2`.

Runic, typos, and `git diff --check` pass on the rewritten stack. The full NNODE validation overlay containing this behavior commit plus the independent PINO and optimizer-owner fixes exited 0 after about 1h22m; the final files were `with_tstops.jl` 6/6 and `vector.jl` 12/12, followed by `Testing NeuralPDE tests passed`. Exact current-master QA remains 23/24 only because of the PINO ambiguity already fixed in https://github.com/SciML/NeuralPDE.jl/pull/1123; with that fix overlaid it is 24/24.

Issue #1161, opened 2026-09-11
NeuralPDE Parser Rewrite


ChrisRackauckas:
NeuralPDe.jl's parser needs a pretty major rewrite. We want it to match the general form that has been growing with MethodOfLines.jl, ModelOrderReduction.jl, etc where everything is a PDESystem -> System
  -> solver. So it should do symbolic_discretize to a System, and then generate the numerical system by doing OptimizationProblem on that System. To make it efficient, it needs to represent everything
  using array operations and array equations. https://github.com/SciML/NeuralPDE.jl/pull/877 https://github.com/SciML/NeuralPDE.jl/issues/574 https://github.com/SciML/NeuralPDE.jl/issues/385 https://
  github.com/SciML/NeuralPDE.jl/issues/687 https://github.com/SciML/NeuralPDE.jl/pull/1064, sathvik's PR as probably the better one, but none are really a great starting point. It should be like
  MethodOfLines.jl. In the end, it should be pretty simple. It just finds the operations inside of the symbolic PDESystem expressions and replaces them with a neural network evaluation. that then gives the
  inner function Resid(x) = 0, where x is all spatial dimensions in my simplified prompt here but it would be like (x,y,z,t,...) depending on the PDESystem given (the independent variables). Derivatives
  then transform into derivative expressions, which either transform into autodiff expressions (just make it Enzyme expressions, and note that these should be forward mode because it's small numbers of
  inputs with large numbers of outputs) or finite difference approximations (like D(u) -> u(x+eps)-u(x)/eps, this code can be reused from the current form in the package probably). So basically dependent
  variables like u(x,t) should be then just referring to neural networks u(x,t,theta1), v(x,t) -> v(x,t,theta2), where thetas are array variables of the neural network parameters, and then you just have
  derivative operator functions on top of that, which don't need to be registered so you can se the u(x+eps,t)-u(x,t)/eps in the symbolic array expression (though this would mean that the enzyme autodiff
  call would need to be registered, that should get done in Symbolics.jl as an extension). Note the handling that has to be done on literals, i.e. u(x,1.0) should be supported, and note this needs to lower
  to a Lux neural network, so u(x,1.0,theta) -> NN([x,1.0),theta] for some lux neural network. Use ModelingToolkitNeuralNetworks.jl for the representation here so that there's uniformity with other
[...]

Issue #469, opened 2021-12-27
Periodic boundary setting


vavrines:
Say solving the advection equation with periodic boundary condition
```latex
u_t+u_x=0, u(0,x)=sin(2πx), u(t,0)=u(t,1)
```
The following code works well
```julia
using NeuralPDE, ModelingToolkit, Plots, Flux, DiffEqFlux, GalacticOptim, Optim
using ModelingToolkit: Interval

@parameters t, x
@variables u(..)
∂t, ∂x = Differential(t), Differential(x)
eq = ∂t(u(t, x)) + ∂x(u(t, x)) ~ 0
domains = [t ∈ Interval(0.0, 1.0), x ∈ Interval(0.0, 1.0)]
bcs = [u(t, 0) ~ sin(-2π * t), u(t, 1) ~ u(t, 0), u(0, x) ~ sin(2π * x)]
dx = 0.05
chain = FastChain(FastDense(2, 16, Flux.σ), FastDense(16, 16, Flux.σ), FastDense(16, 1))
initθ = Float64.(DiffEqFlux.initial_params(chain))
discretization = PhysicsInformedNN(chain, GridTraining(dx); init_params = initθ)
@named pde_system = PDESystem(eq, bcs, domains, [t, x], [u(t, x)])
prob = discretize(pde_system, discretization)
cb = function (p, l)
    println("loss: $l")
    return false
end
opt = Optim.BFGS()
res = GalacticOptim.solve(prob, opt; cb = cb, maxiters = 1200)
phi = discretization.phi
sol0 = [phi([0, x], res.u)[1] for x in xs]
sol1 = [phi([0.5, x], res.u)[1] for x in xs]
plot(xs, sol0)
plot!(xs, sol1)
```
![right](https://user-images.githubusercontent.com/21040679/147505679-5e786872-45d2-4637-be3f-cc71274a0119.png)

If I change the sentence `bcs = [u(t, 0) ~ sin(-2π * t), u(t, 1) ~ u(t, 0), u(0, x) ~ sin(2π * x)]` to a real periodic setting, e.g., `bcs = [u(t, 0) ~ u(t, 1), u(t, 1) ~ u(t, 0), u(0, x) ~ sin(2π * x)]`, the result becomes
![wrong](https://user-images.githubusercontent.com/21040679/147505769-391b22ca-9b69-42d3-a73b-c29cd34dac23.png)

I'm using NeuralPDE v4.2.0.

[...]

PR #1165, opened 2026-09-12
Rewrite the PDE parser: PDESystem → System → OptimizationProblem → PDESolution

ChrisRackauckas-Claude:
> **Note:** please ignore this PR until it has been reviewed by @ChrisRackauckas.

Implements the parser rewrite of https://github.com/SciML/NeuralPDE.jl/issues/1161 following the revised milestone plan in that thread (Milestones 0 and 1, plus the parts of 2 and 3 that were reachable). Breaking: NeuralPDE 6.3.2 → 7.0.0.

## What changed and why

`PhysicsInformedNN` is now a `PDEBase.AbstractOptimizationSystemDiscretization`, and the PDE pipeline is `PDESystem → symbolic_discretize → ModelingToolkit.System → discretize → OptimizationProblem → solve → PDENoTimeSolution`, the same shape as MethodOfLines:

- **Representation.** Each dependent variable is a `ModelingToolkitNeuralNets.SymbolicNeuralNetwork` callable; its flat parameter vector is an array *unknown* of the `System` (as decided in the issue). Every equation and boundary condition becomes a `ResidualBlock`: its free independent variables are sampled into a non-tunable `d × n` collocation matrix parameter, `u(x, t)` lowers to a batched network evaluation `nn_eval(NN_u, P * xs .+ c, θ_u)` (literal arguments enter through `c`, so `u(x, 1.0)`, periodic conditions `u(t, 0) ~ u(t, 1)` and point conditions `Dx(u(1.0)) ~ 1` fall out of the same rule), derivatives lower to central finite-difference stencils on the network inputs (orders 1–4 explicit, higher and mixed orders recursive; `FiniteDifferenceDerivative` is the first `AbstractDerivativeLowering`), and every scalar operation is broadcast. The residual is a lazy `1 × n` symbolic array whose size does not depend on `n`; the generated code is one batched network call per stencil point with no per-point scalarization. Costs are `sum(abs2, r) / n` (or quadrature-weighted sums); boundary conditions are penalty costs by default or equality constraints with `boundary_policy = :constraints`; `param_estim = true` promotes the `PDESystem` parameters to unknowns; `additional_loss(phi, θ, p)` becomes a callable parameter cost.
- **Driver.** The lowering runs through the new PDEBase optimization-system driver (https://github.com/SciML/PDEBase.jl/pull/120) as hook implementations (`construct_disc_state`, `construct_discrete_space`, `construct_differential_discretizer`, `discretize_equation!`, `generate_metadata`, `generate_system`), so `VariableMap`, `make_pdesys_compatible` and `handle_complex` are shared with MethodOfLines.
- **Problem generation.** `discretize` is `mtkcompile` + sampling + `OptimizationProblem(sys, op; adtype, weights, ...)`; all keywords are forwarded, so ModelingToolkit's `weights` (#5109) and `adtype` (#5110) do the scalarization and AD selection. `remake(prob; u0 = θ)` is transfer learning by warm start, `remake(prob; p = [block.xs => X])` replaces collocation points, `resample!(p, md)` redraws them from a callback for `StochasticTraining`/`QuasiRandomTraining`.
- **Solutions.** `solve` returns a `PDENoTimeSolution` (via `problem_type` metadata and `wrap_sol`, needing https://github.com/SciML/SciMLBase.jl/pull/1600, https://github.com/SciML/Optimization.jl/pull/1354 and https://github.com/SciML/ModelingToolkit.jl/pull/5115): `sol[u(x, y)]` on the evaluation grid, `sol(x, y; dv = u(x, y))` at points or grids, `sol.original_sol` for the `OptimizationSolution`.
- **Strategies.** `GridTraining`, `StochasticTraining`, `QuasiRandomTraining` keep their meaning (mean squared residual over their points; interior grids exclude boundary-pinned values); `QuadratureTraining` for PDEs is a fixed tensor-product Gauss–Legendre rule (`quadrature_alg = GaussLegendre(n = ...)`) since adaptive cubature cannot be a static problem parameter. Custom strategies implement `collocation_count`/`sample_points`.
- **Removed (breaking, per the issue):** the `Expr`-level parser (`symbolic_utilities.jl`), `PINNRepresentation`/`PINNLossFunctions`, `Phi`/`numeric_derivative`, adaptive losses (now `weights` parameters updated from callbacks), `LogOptions`/logger extension, `neural_adapter`, the PDE Bayesian PINN (`BayesianPINN`, `ahmc_bayesian_pinn_pde`), integro-differential support (`Integral` now errors with a pointer to the follow-up), and the PDE CUDA tests. The ODE solvers (`NNODE`, `NNDAE`, `BNNODE`, `PINOODE`, `NNSDE`) are untouched apart from `get_loss_function` losing its `PINNRepresentation` branch; `SDEPINN` is ported to the new API; `DeepGalerkin` keeps working (its skip connection is now a container layer so `stateless_apply` works).
- **Docs.** First tutorial is Reactant-compiled (objective + Enzyme gradient under `@compile`); all PDE tutorials and examples ported; developer pages describe the lowering hooks and how to inspect residuals through SymbolicIndexingInterface.

## Review round 1 (2026-09-13)

- **`sol[u(x, y)]` first.** The first PDE tutorial and `examples/3rd.md` now evaluate the trained network with `sol[u(x, y)]` on the grid `sol[x]`, `sol[y]` (same indexing as MethodOfLines.jl, implemented in NeuralPDE like MethodOfLines does it), then show `sol(x, y; dv = u(x, y))` for arbitrary points.
- **No `mtkcompile`.** It was only there because ModelingToolkitBase's `OptimizationProblem` constructor rejected array unknowns (`ArgumentError: The system has array unknowns. Call mtkcompile`). https://github.com/SciML/ModelingToolkit.jl/pull/5131 lifts that, so `discretize` now calls `complete` and the network parameters stay array unknowns; compat is `ModelingToolkitBase = "1.72"`, which that PR releases. On the 2D Poisson PINN the two paths are bit-identical; at 4,417 parameters `mtkcompile(sys)` did not finish in 1h39m while `complete` builds the problem in 16 s.
- **Rebased** on master (four NNSDE commits). One of them makes the `SDEPINN` OU test pass a Lux `NamedTuple` as initial parameters, which the rewrite's `PhysicsInformedNN` did not accept (`MethodError: _param_eltype(::NamedTuple)`); commit 5b982d32 accepts flat vectors, Lux `NamedTuple`s and `ComponentArray`s, with a test that fails before and passes after (below).
- Compat bounds now point at the released SciMLBase 3.54.0, SymbolicUtils 4.46.6, PDEBase 0.1.36 and Optimization 5.9.1 (OptimizationBase 5.6.0).

## Review round 2 (2026-09-13)

- https://github.com/SciML/ModelingToolkit.jl/pull/5131 was widened to every problem type (no trait, no kwarg, initialization enabled); this PR's compat `ModelingToolkitBase = "1.72"` is that release. All 13 PDE test files and all 17 PDE docs pages were re-run against the rebased widened branch: all pass, with one exception fixed in commit b7bc6c8c below.
- **Lorenz parameter-estimation test** (`test/NNPDE2/inverse_and_data.jl`): it failed against the widened branch with σ ≈ 2.6 instead of 10. Bisected: objective, indices, symbolic reads and `remake` are bit-identical between the first and widened branch versions; the Zygote gradient differs by 8e-15 relative (summation order), and 2000 BFGS iterations with backtracking amplify that into a different basin. The deeper cause is the test's starting weights: `symbolic_discretize` was called before `discretize` and each call draws from `disc.rng`, so training started from the second draw, from which 13 of 16 schedule × branch-version combinations end at objectives in the hundreds or thousands (the original schedule passed on the first version by luck). From the first draw all six schedules on both versions converge to objective ≈ 0.0025. The test now discretizes first and asserts `objective < 1.0`; it passes 4/4 on both branch versions. The GBM weak-solve flake noted below is the same class (BFGS sensitivity) and is left as is.

## Verification

**Final commit (`complete` pipeline, needs MTKBase 1.72):** every PDE test file run on Julia 1.11.9 in an environment with the released packages plus ModelingToolkitBase dev'd from https://github.com/SciML/ModelingToolkit.jl/pull/5131 (`julia +1.11 --project=env_complete -e 'using Test; include("test/<group>/<file>.jl")'`): `NNPDE1` (poisson_2d 42, ode_and_derivatives, systems), `NNPDE2` (boundaries, inverse_and_data), `Interface` (lowering incl. the new init-params cases, abstract contracts 22, precompile workload 2), `DGM` (2 + 1 + 1), `NNSDE2` (GBM 1, OU 1) all pass, zero errors. All 17 PDE docs pages (`tutorials/pdesystem`, `low_level`, `systems`, `constraints`, `param_estim`, `derivative_neural_network`, `dgm`, `examples/3rd`, `wave`, `ks`, `heterogeneous`, `complex`, `nonlinear_elliptic`, `nonlinear_hyperbolic`, `linear_parabolic`, `developer/debugging`, `interfaces`) executed block by block in the docs environment with that MTKBase: every block ok.

**Commit cff7c592 (rebased, `mtkcompile` pipeline, released packages only):** `GROUP=<g> julia +1.11 --project -e 'using Pkg; Pkg.test()'` from the registry: QA, Interface, NNPDE1, NNPDE2, DGM, NNSDE1, NNODE pass. NNSDE2 failed twice: the OU `MethodError` above (fixed) and, once out of three otherwise identical runs, the GBM tolerance (`mean(abs2, diff) = 0.0757 < 0.05` failed; the two re-runs on the same manifest passed, and four standalone runs of the same file, with 1 and 8 BLAS threads, all give exactly 0.0284). I could not reproduce the 0.0757 and did not change the test.

**Init-params fix, before/after** (same test, `test/Interface/lowering.jl` "initial parameters and eltype"):
```
BEFORE (commit cff7c592):
NamedTuple/ComponentArray init_params: Error During Test
  MethodError: no method matching _param_eltype(::@NamedTuple{layer_1::@NamedTuple{weight::Matrix{Float32}, bias::Vector{Float32}}, ...
Test Summary:                         | Error  Total  Time
AFTER (commit 5b982d32):
Test Summary:                         | Pass  Total   Time
NamedTuple/ComponentArray init_params |    3      3  21.0s
```

The table below is the original verification of commit cff7c592 with the upstream branches dev'd (`julia --project=env test/<group>/<file>.jl`), kept for the per-file counts.

| Test file | Result |
|---|---|
| `NNPDE1/poisson_2d.jl` (2D Poisson under Grid/Stochastic/QuasiRandom×2/Quadrature, max error < 0.01 vs analytic, solution interface) | 42 pass |
| `NNPDE1/ode_and_derivatives.jl` (3rd-order ODE with `Dx(u(1.0))` point condition, wave `Dtt`, mixed `Dx(Dy(u))`) | 6 pass |
| `NNPDE1/systems.jl` (one chain per variable, shared multi-output chain, heterogeneous arguments) | 12 pass |
| `NNPDE2/inverse_and_data.jl` (Lorenz `param_estim` recovers σ, ρ, β within 10%; fixed parameters through `p =`/`remake`; data-fit `additional_loss`) | 7 pass |
| `NNPDE2/boundaries.jl` (periodic BC of #469, `boundary_policy = :constraints`, error messages) | 9 pass |
| `Interface/lowering.jl` (lowered residual vs manual FD to `≈`, Zygote vs ForwardDiff gradients, Zygote/Enzyme/ForwardDiff solves, `resample!`, `remake`, `weights`, quadrature weights, Float32) | 28 pass |
| `Interface/interface__abstract_contracts.jl`, `Interface/precompile_workload.jl` | 22 + 2 pass |
| `DGM/dgm__poisson_s_equation.jl`, `DGM/dgm__burger_s_equation.jl`, `DGM/dgm__black_scholes_pde_european_call_option.jl` | 2 + 1 + 1 pass |
| `NNSDE2/nn_sde_weaksolve__ou_process.jl`, `NNSDE2/nn_sde_weaksolve__gbm_sde.jl` (`SDEPINN` on the new pipeline) | 1 + 1 pass |
| ODE-side regression sample, untouched code: `NNODE/nnode__ode_i.jl`, `NNODE/nnode__training_strategy_others.jl`, `NNODE/nndae__dae_case_i.jl`, `NNODE/nndae__dae_case_ii.jl`, `NNSDE1/gbm_reference_moments.jl` | 4 + 3 + 1 + 1 + 48 pass |
| `qa/qa.jl` | 19/20: the only error is Aqua's persistent-tasks check, which builds a fresh environment and cannot resolve the unreleased PDEBase 0.1.36 |

**Docs.** A full `docs/make.jl` run with the same dev'd branches expanded `api.md`, `developer/debugging.md`, `developer/interfaces.md`, `examples/3rd.md`, `complex.md`, `heterogeneous.md`, `ks.md`, `linear_parabolic.md` and `nonlinear_elliptic.md` without a single `failed to run`, then sat for over an hour on `examples/nonlinear_hyperbolic.md` *before* logging the evaluation of its first block (a second attempt did the same on `linear_parabolic.md`; an index-only `makedocs` also stalled for >8 min before expanding anything, with the profile pointing into `Document`/`interpret_repo_and_remotes` under `invokelatest`, i.e. compilation, not the page code). Since the pages themselves run in a few minutes (`nonlinear_hyperbolic.md`: 78 s discretize, 0.3 s per 10 BFGS iterations), every PDE page was instead executed block by block exactly as Documenter does (`Documenter.parseblock` + `Core.eval` in a fresh sandbox module under `IOCapture.capture`), in the docs environment: `tutorials/pdesystem.md` (10 blocks, including the Reactant `@compile` of the objective and its Enzyme gradient), `systems.md`, `constraints.md`, `low_level.md`, `param_estim.md`, `derivative_neural_network.md`, `dgm.md`, `examples/wave.md` (the damped-wave block alone takes 30 min) all pass; `nonlinear_elliptic.md`, `nonlinear_hyperbolic.md`, `linear_parabolic.md` and `developer/debugging.md` passed in the full run and/or standalone. Two docs bugs found this way are fixed (a top-level loop rebinding `θ` in the Reactant block, and the systems tutorial's callback now evaluating residuals through a `ProblemState`). `gpu.md` is intentionally not executed (see gaps). The docs CI job on a clean runner is the arbiter for the Documenter stall; if it reproduces there, it is a Documenter/compile-latency issue with the loaded package set rather than page code. Three ODE-side files (`nnode__nnode_translating_from_flux.jl`, `pino_ode__example_chain_du_cos_p_t.jl`, `bpinn__bpinn_ode_iii_inverse_solve_improvement.jl`) could not run in the ad-hoc environment only because Flux, NeuralOperators and Functors are not installed there; they are untouched by this PR.

Accuracy of the rewrite on the reference problem (`proto/p4.jl` in the session): 2D Poisson, `GridTraining(0.1)`, `Chain(Dense(2,12,σ), Dense(12,12,σ), Dense(12,1))`, Adam 2000 + BFGS 1000 → objective 5e-7, max error 5e-4 vs the analytic solution; the lowered residual matches a hand-written finite-difference residual of the same Lux chain to 6e-8.

Representation size and codegen (`proto/p2.jl`): the generated objective is one batched network call per stencil point (no per-point scalarization; the expression string is independent of the number of collocation points), `remake` with new points works, and Zygote, ForwardDiff and Enzyme all differentiate it. Build time is dominated by ModelingToolkit scalarizing the array unknown:

| parameters | points | `mtkcompile` | `OptimizationProblem` | objective eval |
|---|---|---|---|---|
| 337 | 500 | 0.15 s | 0.35 s | 1.0 ms |
| 1185 | 2000 | 0.22 s | 1.4 s | 9.2 ms |
| 4417 | 5000 | 1.6 s | 14 s | 86 ms |

Reactant (`proto/p6_reactant.jl`): the generated objective compiles with `@compile`, and `Enzyme.gradient` under Reactant matches the ForwardDiff gradient to 6e-9 (16 µs per gradient vs 430 µs for CPU Enzyme on the small problem).

`typos` over `src`, `test`, `docs/src`: clean. Runic applied to every changed file.

## Upstream dependencies

All released except the last row, so CI cannot resolve until ModelingToolkitBase 1.72 is tagged.

| Package | Change | PR |
|---|---|---|
| SymbolicUtils 4.46.6 | `expand`/ordering on array terms | https://github.com/JuliaSymbolics/SymbolicUtils.jl/pull/1072 |
| SciMLBase 3.54 | `problem_type` slot on `OptimizationProblem` | https://github.com/SciML/SciMLBase.jl/pull/1600 |
| OptimizationBase 5.6 | `solve` wraps through `problem_type` | https://github.com/SciML/Optimization.jl/pull/1354 |
| ModelingToolkitBase 1.71.1 | `weights`, `adtype`, `ProblemTypeCtx` forwarding, `tovar` public (all merged) | https://github.com/SciML/ModelingToolkit.jl/pull/5115, https://github.com/SciML/ModelingToolkit.jl/pull/5120 |
| PDEBase 0.1.36 | optimization-system driver | https://github.com/SciML/PDEBase.jl/pull/120 |
| **ModelingToolkitBase 1.72 (open)** | array unknowns on a `complete`d system for every problem type | https://github.com/SciML/ModelingToolkit.jl/pull/5131 |

## Not verified / known gaps

- **Enzyme on the CPU path is not the default.** Static activity analysis rejects the generated objective ("constant memory is stored to a differentiable variable" through the `MTKParameters` buffers), and with `set_runtime_activity(Reverse)` the reverse pass through an `additional_loss` closure overwrote the data array the closure captured (`proto/p17.jl`: `xs unchanged=false`), silently changing later objectives. Zygote is the default; the Reactant-compiled Enzyme gradient is correct. Tracked in the follow-up issue.
- **Integro-differential equations** are not supported yet (`Integral` raises an `ArgumentError`), nor **GPU training** (the MTKNN parameter wrapper copies to the host), nor **general arguments** such as `u(2x)`. All listed with contracts in the follow-up issue.
- **`boundary_policy = :constraints`** emits array-valued equality constraints into the `System`; lowering them through `OptimizationProblem` needs array-aware `cons`/`lcons`/`ucons` in ModelingToolkitBase (follow-up), so no constrained-solver test exists yet.
- **Problem generation cost** no longer grows with the parameter count (`complete` keeps the parameters as array unknowns); the DGM tests and tutorial still use small networks and could be re-enlarged in a follow-up.
- **Not run locally:** CUDA group (removed for PDEs, see above), Julia 1.10, downstream packages, a full `docs/make.jl` (see the Documenter stall above), and `Pkg.test` on the final commit (needs MTKBase 1.72; the previous commit's groups ran from the registry as described).
- **GBM weak-solve tolerance** failed once in three identical `Pkg.test` runs and never standalone; see Verification. Worth watching in CI.
- **Judgment calls to push back on:** `QuadratureTraining` for PDEs now means a fixed Gauss–Legendre rule; the evaluation grid of the solution interface is `eval_points` (default 100) per independent variable; the Bayesian PDE solver and `neural_adapter` are removed rather than ported.

🤖 Generated with [Claude Code](https://claude.com/claude-code) (model: claude-fable-5-1)

https://claude.ai/code/session_01WYPUXLxE9jVu2jXLFRd8ax


docs/src/examples/3rd.md
```diff
@@ -60,10 +56,9 @@ using Plots
 
 analytic_sol_func(x) = (π * x * (-x + (π^2) * (2 * x - 3) + 1) - sin(π * x)) / (π^3)
 
-dx = 0.05
-xs = [leftendpoint(d.domain):(dx / 10):rightendpoint(d.domain) for d in domains][1]
+xs = 0:0.005:1
 u_real = [analytic_sol_func(x) for x in xs]
-u_predict = [first(phi(x, res.u)) for x in xs]
+u_predict = sol(xs; dv = u(x))
```
ChrisRackauckas: `solu = sol[u(x)]` should be how it is just like MethodOfLines.jl? That should be given via PDEBase interface?

docs/src/manual/pinns.md
```diff
@@ -25,22 +25,59 @@ the parameter of the equation (like `[x,y]`), and `var` represents variables (li
...
+   stored as a parameter of the system; each residual contributes one cost (or, for
+   boundary conditions with `boundary_policy = :constraints`, one equality constraint).
+2. `discretize(pdesys, discretization)` compiles the system with `mtkcompile`, samples the
+   collocation points with the training strategy and generates an `OptimizationProblem`
+   through ModelingToolkit's `OptimizationProblem(sys, op)` constructor. Keyword arguments
+   such as `adtype` (automatic differentiation backend) and `weights` (weighted sum of the
+   costs) are forwarded to it.
+3. `solve(prob, optimizer)` returns a `PDENoTimeSolution` which evaluates the trained
+   networks: `sol[u(x, t)]` on the evaluation grid, `sol(x, t; dv = u(x, t))` at arbitrary
+   points, and `sol.original_sol` for the underlying `OptimizationSolution`.
```
ChrisRackauckas, line 48: I see, the first tutorial should use the simple form first though.

docs/src/tutorials/low_level.md
```diff
@@ -39,50 +37,62 @@ domains = [t ∈ Interval(0.0, 1.0),
...
+```@example low_level
+md = pinn_metadata(sys)
+md.blocks[end].residual
+```
 
-f_ = OptimizationFunction(loss_function, AutoZygote())
-prob = OptimizationProblem(f_, sym_prob.flat_init_params)
+Because the result is a `System`, the standard ModelingToolkit compiler builds the
+`OptimizationProblem`. `discretize` does exactly this: it calls `mtkcompile`, samples the
+collocation points and calls `OptimizationProblem(sys, op)`, forwarding keyword arguments
```
ChrisRackauckas: No `mtkcompile` for now, it won't help much on OptimizationProblem?

docs/src/manual/pinns.md
```diff
@@ -25,22 +25,59 @@ the parameter of the equation (like `[x,y]`), and `var` represents variables (li
...
+as the other SciML discretizers (for example MethodOfLines.jl):
+
+1. `symbolic_discretize(pdesys, discretization)` returns a `ModelingToolkit.System`. Each
+   dependent variable is represented by a `ModelingToolkitNeuralNets.SymbolicNeuralNetwork`
+   whose parameters are array unknowns of the system; the residual of every equation and
+   boundary condition is a symbolic array expression over a matrix of collocation points
+   stored as a parameter of the system; each residual contributes one cost (or, for
+   boundary conditions with `boundary_policy = :constraints`, one equality constraint).
+2. `discretize(pdesys, discretization)` compiles the system with `mtkcompile`, samples the
+   collocation points with the training strategy and generates an `OptimizationProblem`
```
ChrisRackauckas: What's the reason for `mtkcompile`?

docs/src/examples/3rd.md
ChrisRackauckas-Claude: Done: the first tutorial and this example now lead with `xs = sol[x]`, `u_predict = sol[u(x)]` and show `sol(x; dv = u(x))` afterwards for arbitrary points. The indexing is implemented in NeuralPDE the same way MethodOfLines does it (a `getindex` on `PDENoTimeSolution` dispatching on the discretizer's metadata type), not in PDEBase.

docs/src/manual/pinns.md
ChrisRackauckas-Claude: Done in commit ae17b902: `tutorials/pdesystem.md` uses `sol[u(x, y)]` on `sol[x]`, `sol[y]` first, then shows `sol(x, y; dv = u(x, y))` for arbitrary points.

docs/src/tutorials/low_level.md
ChrisRackauckas-Claude: Removed. `discretize` now calls `complete` only. It needed https://github.com/SciML/ModelingToolkit.jl/pull/5131, because ModelingToolkitBase's `OptimizationProblem` constructor threw `ArgumentError: The system has array unknowns. Call mtkcompile` for the parameter arrays; with that PR the parameters stay array unknowns and the two paths are bit-identical on the Poisson PINN. At 4,417 parameters `mtkcompile` did not finish in 1h39m; `complete` builds the problem in 16 s. Compat is now `ModelingToolkitBase = "1.72"`.

docs/src/manual/pinns.md
ChrisRackauckas-Claude: Only that ModelingToolkitBase's `OptimizationProblem` constructor refused array unknowns without it (it was the scalarization pass, not structural simplification). Gone now, see the comment on `low_level.md` and https://github.com/SciML/ModelingToolkit.jl/pull/5131.

PR #1167, opened 2026-09-13
Reactant nnode batching

RomanSahakyan03:
## Checklist

* [x] Appropriate tests were added
* [x] Any code changes were done in a way that does not break public API
* [x] All documentation related to code changes were updated
* [x] The new code follows the
  [[contributor guidelines](https://github.com/SciML/.github/blob/master/CONTRIBUTING.md)](https://github.com/SciML/.github/blob/master/CONTRIBUTING.md), in particular the [[SciML Style Guide](https://github.com/SciML/SciMLStyle)](https://github.com/SciML/SciMLStyle) and
  [[COLPRAC](https://github.com/SciML/COLPRAC)](https://github.com/SciML/COLPRAC).
* [x] Any new documentation only uses public API

## Additional context

This PR adds an opt-in Reactant/Enzyme training path for `NNODE` through the new
`reactant = true` keyword. The existing NNODE path remains the default and is
unchanged when `reactant = false`.

The main goal is to compile the collocation loss with Reactant while keeping the
user-facing ODE RHS in the standard pointwise form

```julia
f(u, p, t)
```

rather than requiring a separately batched RHS.

For the compiled path, the collocation loop is expressed using
`Reactant.@trace`. Reactant raises this loop to loop-free StableHLO tensor
operations in both the primal loss and its reverse gradient. The emitted
operation structure does not grow with the number of collocation points,
although the compiled tensor shapes remain specialized to the grid size.

Time derivatives can be computed using an Enzyme forward-mode JVP. The
single-pass all-ones tangent assumes that the neural network does not couple
different collocation points across the batch dimension, which is satisfied by
the standard Dense NNODE architectures.

The compiled objective:

* supports both real and complex network parameters;
* keeps the scalar loss real-valued for complex networks;
* reuses a persistent Reactant parameter buffer between evaluations;
* lazily compiles the objective and gradient thunks on first use;
* raises on unsupported `reactant = true` configurations rather than silently
  falling back to the legacy path.

The current Reactant path is intentionally limited to batched `GridTraining`
forward problems. Parameter estimation, datasets, additional losses, `tstops`,
and other training strategies remain on the existing NNODE implementation.

A dedicated `[Reactant]` test group was added. It covers:

* Reactant tracing and compiled residual evaluation;
* Enzyme forward JVPs against per-point `ForwardDiff` references;
* outer reverse gradients against central finite differences for real
  parameters and complex parameters along both real and imaginary directions;
* persistent parameter-buffer reuse;
* end-to-end `NNODE -> OptimizationFunction -> Adam -> solve` training.

The legacy batched NNODE loss was also changed to evaluate the user RHS
pointwise instead of explicitly constructing a batched RHS evaluation. This
avoids relying on a batched `f` interface and provides the same pointwise RHS
semantics used by the Reactant path.


RomanSahakyan03:
Hi @ChrisRackauckas,

I opened the PR for the Reactant/Enzyme NNODE changes. It is now rebased on the latest master and includes the dedicated Reactant regression tests

ChrisRackauckas:
Rebase this

RomanSahakyan03:
> Rebase this

@ChrisRackauckas Done

RomanSahakyan03:
Hi @ChrisRackauckas,

I’ve fixed the issues that came up in the CI checks in the latest commit.

I also noticed a bug related to callbacks and "verbose", which I addressed in a separate PR #1209

Issue #1170, opened 2026-09-13
Make Enzyme the CPU default for the generated PINN objective


ChrisRackauckas-Claude:
NeuralPDE 7.0 (https://github.com/SciML/NeuralPDE.jl/pull/1165) uses `AutoZygote()` as `default_adtype()`. `OptimizationEnzymeExt` on the generated objective needs runtime activity (static activity analysis reports "constant memory is stored to a differentiable variable" through the `MTKParameters` constant buffers), and with `set_runtime_activity(Reverse)` the reverse pass through an `additional_loss` closure overwrote the array the closure captured, which silently changed the objective for the rest of the session. The Reactant-compiled Enzyme gradient of the same objective is correct (matches ForwardDiff to 6e-9), so this is CPU-only.

**Contract.** (1) A standalone Enzyme reproducer, importing no SciML package, of the captured-array write under runtime activity, filed on Enzyme.jl. (2) A generated objective that passes static activity analysis so `AutoEnzyme()` works without runtime activity, then `default_adtype() = AutoEnzyme()` with the AD-agreement tests in `test/Interface/lowering.jl` extended to it.

---
*Opened by an AI agent (Claude Code 2.1.269, model claude-fable-5-1) on behalf of Chris Rackauckas; session https://claude.ai/code/session_01WYPUXLxE9jVu2jXLFRd8ax*

ChrisRackauckas-Claude:
Taking this — opencode 1.18.31, model opencode-go/kimi-k3 (local job 1170-opencode on amdci2.julia.csail.mit.edu).

ChrisRackauckas-Claude:
First-pass reduction result: the captured-array overwrite does **not** reproduce on current releases (Julia 1.11.9, Enzyme 0.13.204, OptimizationBase 5.7.0).

- `solve(prob, Adam; maxiters=2)` with `adtype = AutoEnzyme(; mode = Enzyme.set_runtime_activity(Enzyme.Reverse), function_annotation = Enzyme.Const)` and an `additional_loss` closure capturing `xs`/`ys`: `xs unchanged=true`, `ys unchanged=true`.
- `Enzyme.autodiff(Enzyme.set_runtime_activity(Enzyme.Reverse), Const(firstapply), Active, Const(prob.f.f), Duplicated(prob.u0, g), Const(prob.p))` directly: captures unchanged, gradient computed cleanly.

[...]

Issue #1166, opened 2026-09-12
Parser rewrite follow-ups: integrals, Enzyme/GPU, exact constraints, array coordinates, Bayesian PDEs


ChrisRackauckas-Claude:
Tracking issue for the work that the parser rewrite (https://github.com/SciML/NeuralPDE.jl/pull/1165) deliberately left out of NeuralPDE 7.0, with the concrete contract each item has to meet. Items marked **[external]** live in other repositories.

## NeuralPDE.jl

1. **Integro-differential equations.** `Integral` terms currently raise an `ArgumentError` in the lowering. Contract: a registered quadrature term over the batch of collocation points (`quadrature(NN, X, θ, lb, ub, weights)` with the inner integration variable appended to the network input), fixed-node rules only so that the result is Reactant compatible, plus the infinite-bound transformations from `transform_inf_integral.jl` of NeuralPDE 6. Depends on the registration home decision in Symbolics/Integrals (item 15).
2. **Enzyme as the CPU default.** `OptimizationEnzymeExt` on the generated objective needs runtime activity (static analysis reports "constant memory is stored to a differentiable variable" through the `MTKParameters` constant buffers), and with `set_runtime_activity(Reverse)` the reverse pass through an `additional_loss` closure overwrote an array the closure captured (`xs unchanged=false` in `proto/p17.jl` of the rewrite session), which silently changed the objective for the rest of the session. Zygote is therefore the default in 7.0. Contract: a standalone Enzyme reproducer (no SciML packages) of the captured-array write under runtime activity, and a generated objective that passes static activity analysis so that `AutoEnzyme()` works without runtime activity. The Reactant-compiled Enzyme gradient of the same objective is correct (matches ForwardDiff to 6e-9).
3. **Enzyme forward-mode derivative backend.** `AbstractDerivativeLowering` has one implementation (`FiniteDifferenceDerivative`). Contract: `lower_derivative(::EnzymeForwardDerivative, ...)` emitting a registered batched JVP term whose outer reverse-mode gradient (`∇θ L(∂²ₓ NN)`) is verified against the finite-difference backend and against ForwardDiff. Depends on the SymbolicsEnzymeExt (item 14).
[...]

PR #1191, opened 2026-09-18
Fix DGM tutorial AD backend so the docs build passes

ChrisRackauckas-Claude:
Please ignore until reviewed by @ChrisRackauckas.

## What changed and why

`docs/src/tutorials/dgm.md` now passes `adtype = AutoZygote()` to `discretize`.

The default backend became `AutoEnzyme()` in https://github.com/SciML/NeuralPDE.jl/pull/1184, but the DGM network threads the collocation points through Lux `SkipConnection`s next to the active hidden state, and Enzyme's static activity analysis rejects the generated objective with `EnzymeRuntimeActivityError: Detected potential need for runtime activity` ("Constant memory is stored (or returned) to a differentiable variable"). That made the `@example dgm` block fail and Documenter terminated the build before rendering. `AutoZygote()` is the fallback the package documents for objectives the static-Enzyme default cannot handle, and it is what this example used before #1184.

The real fix — a generated DGM objective that passes static activity analysis (or an Enzyme-side fix) — belongs upstream and is tracked in https://github.com/SciML/NeuralPDE.jl/issues/1170 and https://github.com/SciML/NeuralPDE.jl/issues/1166. This PR only makes the documentation build again; it does not change `default_adtype()`.

## Verification (local, Julia 1.12.4)

Docs environment, with the same resolved versions as the failing CI run (Enzyme v0.13.204, Lux v1.31.4, LuxCore v1.5.3):

```
julia --project=docs -e 'using Pkg; Pkg.develop(PackageSpec(path=pwd())); Pkg.instantiate()'
```

**Before the fix** — the `@example dgm` block extracted verbatim from `docs/src/tutorials/dgm.md` (lines 60-130):

```
julia --project=docs dgm_before.jl
```

output tail:

```
ERROR: LoadError: EnzymeRuntimeActivityError: Detected potential need for runtime activity.

Constant memory is stored (or returned) to a differentiable variable and correctness cannot be guaranteed with static activity analysis.
...
in expression starting at .../dgm_before.jl:58
```

(line 58 is `res = solve(prob, Adam(0.1); maxiters = 100)`).

**After the fix** — the `@example dgm` block extracted from the changed file:

```
julia --project=docs dgm_after.jl
```

output (full log):

```
DGM EXAMPLE COMPLETED OK
```

The script runs both solves (`maxiters = 100` and `maxiters = 500`), `sol(ts, xs; dv = u(t, x))`, and the three `Plots` calls to completion.

**Full docs build** — `julia --project=docs docs/make.jl` from the repo root evaluates `tutorials/dgm.md` without any `:example_block` error. The build still terminates with `makedocs encountered an error [:example_block]`, but every remaining error is on an unrelated page (`complex.md`, `dae.md`, `data_collocation_Inverse.md`, `Lotka_Volterra_BPINNs.md`, `ode.md`, `ode_parameter_estimation.md`) and is the pre-existing `MethodError: no method matching __solve(::ODEProblem/...; callback)` on master, which the `docs/add-documentation-ci` branch fixes separately. On that branch's failing run (https://github.com/SciML/NeuralPDE.jl/actions/runs/35242789475) `dgm.md` was the *only* failing `@example` block, so this change is the remaining docs error there.

## Not verified

- CI on Julia 1.13.0: the failing run used Julia 1.13.0 and the local reproduction used Julia 1.12.4. The failure and the fix were reproduced with the identical resolved dependency versions (Enzyme v0.13.204, Lux v1.31.4), but the exact 1.13.0 lane was not run locally.
- `master` has no `.github/workflows/Documentation.yml` (docs run through Buildkite), so opening this PR will not by itself trigger the Documentation workflow.
- GPU / Reactant / downstream CI lanes, which are out of scope for this docs-only change.

🤖 Generated with OpenCode (model: muse-spark-1.3-contributor-free); no public session URL (local session at /home/crackauc/sandbox/tmp_20260917_040834_91074).


ChrisRackauckas-Claude:
Superseded: this fix is now a commit on #1190, which is the PR that introduces the NeuralPDE docs workflow. A separate PR off master cannot run the docs CI (the workflow file only exists on #1190's branch), so the fix belongs there where the documentation build is actually exercised.

